PR #8, opened 2020-01-31
Add yaml input to submit-nonmem-model

seth127:
Added v1 of yaml input capability which the following behavior:
* yaml must contain a key `model_path` that points to the model file
* any keys that are the same are valid `.args` are parsed through to `.args` and overwrite anything passed to the function (is this right or should it be the other way?)
* any other keys are parsed to a JSON (that currently goes nowhere)


seth127:
This was a stale old PR that has been replaced by https://github.com/metrumresearchgroup/rbabylon/pull/10

PR #19, opened 2020-02-26
Post processing v1 and prep for initial release

seth127:
## Summary
There's a ton of stuff in here. Aside from some heavy refactoring, these are the primary user-facing features that have been added:
* S3 object interface to most NONMEM functionality, based around three main classes:
  * `bbi_nonmem_spec`
  * `bbi_nonmem_result`
  * `bbi_nonmem_summary`
* `submit-nonmem-model.R` deleted in favor of more general `submit-model.R` which determines model type from S3 class and accompanying YAML
* `workflow.R` generalized (similarly to `submit-model.R`) to determine model type from S3 class and accompanying YAML, and interact with that object and YAML for model iteration.
* `workflow.R` also enables users to add tags, decisions, based_on, and description in R.
* `run_log()` function improved and `config_log()` function added.
* `summary.R` includes functionality to create a `bbi_nonmem_summary` object by running `bbi nonmem summary --json` and then interact with that object to view summary info.
* Functionality to monitor NONMEM jobs and output in `R/read-output.R`

## Outstanding work
This still needs a few things before we merge (or at least before we release):
* clean up `inst` folder and figure out any vignettes or examples we want to include
* finish `tests/testthat/test-workflow-composable-bbi.R` which will be the only test to actually run `bbi` commands and will only run on Metworx.

## Out of Scope
The following things were determined to be out-of-scope for this release (by @seth127). We should discuss these with @dpastoor to make sure we're all in agreement.
* Hash checking on the YAML file to be sure it is in sync with the S3 object in memory
* Reconstructing `result` or `summary` objects from a run log df. This seems like the most reasonable way to interact with results for analysis later in the project. 
* Filtering run log df's on tags, etc. Currently this requires a lot of dplyr gymnastics because they are in list columns. We should make helper functions for this and/or reimplement some of it with `vctrs`
* Submit multiple models in parallel, ideally in a single (or as few as possible) `bbi` calls. Currently this is possible only with `map(list_of_models, submit_model)` which makes a separate `bbi` call for each model.)
* Making summary calls on more than one model and extracting relevant data from the list of summary objects. Probably similar implementation to the "submit models in parallel..." bullet. Possibly involving `vctrs`


dpastoor:
ok I wired up CI so now devtools::test() and devtools::check() will be part of CI so we'll have both dev machine + docker container as the tests to confirm behavior

R/workflow.R
```diff
@@ -303,6 +310,48 @@ copy_nonmem_model_from <- function(
   return(.new_spec)
 }
 
+
+#####################
+# Result object
+#####################
+
+#' Create a `bbi_nonmem_result` object from a file path (should work with either path to a model file, yaml file, and output folder)
```
dpastoor: does this really make sense? like the result object is really the result of _submitting_ the model and contains things like the stdout. Wouldn't it just make sense that the anything that does postprocessing of results can work off of the information in a spec object or path. Basically, any post processing functions that take a result object should be able to act on a spec object or path to yaml (and/or results folder)

R/workflow.R
```diff
@@ -307,6 +321,11 @@ copy_nonmem_model_from <- function(
   new_yaml_path <- yaml_ext(.new_model)
   save_mod_yaml(.new_spec, .out_path = new_yaml_path)
 
+  # optionally open the control stream for editing
+  if (isTRUE(.open_mod_file)) {
```
dpastoor: these are the types of innocent helpers that can complicate an api. For example, "oh why use that, I'd prefer to open it via the rstudio api when xyz". These types of patterns are often best handled as a separate function like edit_mod_file(). Think of this similar to how dplyr functions don't have like a .preview or something arg that will automatically open the df in the viewer. Instead, if you want that you learn to `df %>% ... %>% View()`

R/workflow.R
```diff
@@ -276,13 +277,26 @@ copy_nonmem_model_from <- function(
...
     mod_str <- .parent_spec %>% get_model_path() %>% read_file()
 
     # replace the $PROBLEM line(s)
     mod_str <- str_replace(mod_str,
                 "\\$PROB(.|\n)*?\\$",
-                as.character(glue("$PROBLEM {get_mod_id(.new_model)} {.description}\n\n$")))
+                as.character(glue("$PROBLEM {new_mod_id} {.description}\n\n$")))
+
+    # boilerplate code from Sam to replace other occurances of model id #### !!!!!!
+    # need to test and refactor to use stringr
```
dpastoor: look at str_replace_all(). That said, I don't think this is the right place to have this. It is too metrum specific and will lock this function into a specific api. These replacement functions should be separate and modular pieces `replace_mod_references(spec, ...)` or the like. For example, what happens in bayesian land when you also have some other types of files, or use PSN conventions and want sdtab, patab, etc. 

These are good helpers to have, but just not here

dpastoor:
merging as this really is develop, and will spin off fixes/refactoring

PR #24, opened 2020-03-03
model object refactor

seth127:
# Summary
Refactoring both `bbi_nonmem_spec` and `bbi_nonmem_result` to a common `bbi_nonmem_model` object. Also creating `babylon_process` object for interacting with running process.

User will no longer need to think about the distinction between a `spec` and a `result`, but will just have a `bbi_nonmem_model` object that is able to do everything described in https://github.com/metrumresearchgroup/rbabylon/issues/25 . This will simplify the interface significantly. It should also function as the base class for all (or at least most) user interactions with a model for future development.

The `babylon_process` object can be used to access the actual `processx$new$process` object, as well as some metadata about what was passed in to initiate the call to bbi. The `babylon_process` object can also be passed through to `model_summary()` as described in https://github.com/metrumresearchgroup/rbabylon/issues/22 .


NAMESPACE
```diff
@@ -32,32 +38,29 @@ export(check_ext)
...
 export(check_grd)
 export(check_nonmem_args)
-export(check_nonmem_progress)
 export(check_nonmem_table_output)
 export(check_output_dir)
 export(check_status_code)
 export(combine_list_objects)
 export(config_log)
 export(copy_model_from)
 export(copy_nonmem_model_from)
```
callistosp: I don't think `copy_nonmem_model_from` should be exported

NAMESPACE
```diff
@@ -32,32 +38,29 @@ export(check_ext)
...
-export(create_model)
-export(create_model_from_yaml)
 export(format_cmd_args)
 export(get_mod_id)
```
callistosp: all other functions use `model` instead of `mod`

NAMESPACE
```diff
@@ -32,32 +38,29 @@ export(check_ext)
...
 export(get_output_dir)
 export(get_path_from_object)
 export(get_yaml_path)
-export(import_result)
 export(model_summary)
+export(new_model)
 export(nonmem_summary)
 export(param_estimates)
 export(parse_args_list)
-export(parse_mod_yaml)
```
callistosp: switch `mod` to `model` for this and `save_mod_yaml`

dpastoor:
The main reason for these extra exports is to explore the degree of abstraction
already one thing that was frustrating me was the lack of ide support for submit model, like the ... makes it hard to remember what arg names are and needing to `?submit_model` so the solution was to export the "raw" functions that aren't full generics that actually have all the explicit args, so we can see what interface people prefer the more explicit submit_nonmem_model() which would give you the code completion goodness and is specific to nonmem or just by knowing the args you can still use submit_model

For now we can leave this in and say this is good to merge. It's still worth considering exactly what we expose to the user

NAMESPACE
callistosp: based on @dpastoor's comment below, this should actually be exported

PR #34, opened 2020-03-10
Sets and uses options(rbabylon.model_directory)

seth127:
# Summary
This branch adds the `set_model_directory()` function that sets `options(rbabylon.model_directory)`. Additionally it makes `read_model()` and all model-related S3 dispatches for `character` (i.e. file paths as opposed to model objects) look for that option by default.

It also adds `numeric` dispatches for all the model-related functions. These will only work when `options(rbabylon.model_directory)` is set to the correct directory. This limitation is captured in the docs.

# Tests
- tests/testthat/test-workflow.R
  - compare read_model() and new_model() objects with numeric input
  - copy_from_model numeric
- tests/testthat/test-submit-model.R
  - submit_model(.dry_run=T) with numeric input parses correctly
- tests/testthat/test-summary.R
  - model_summary.numeric produces expected output
- tests/testthat/test-workflow-composable-dryrun.R
  - basic workflow is correct using rbabylon.model_directory = 'model-examples'
  - summary call is the same using rbabylon.model_directory = 'model-examples'
- tests/testthat/test-workflow-composable-bbi.R
  - step by step create_model to submit_model to model_summary works
  - copying model works and new model runs correctly
  - run_log() captures runs correctly


seth127:
One note off the top: right now all the calling code is expecting `options("rbabylon.model_directory")` to be an absolute path. 

If you all `set_model_directory(...)` it does set it to an absolute path (and makes sure it exists) so that's cool.

_However_ if you try to do `options("rbabylon.model_directory" = "some/relative/path")` then all the calling code will likely blow up. Maybe that's just unsustainable and I need to fix that in the calling code. Because this seems dangerous. Plus, I remember us wanting to use that syntax in the `.Rprofile` so that we didn't need rbabylon installed.

tests/testthat/test-workflow-composable-bbi.R
```diff
@@ -42,59 +36,92 @@ cleanup <- function(.recreate_dir = FALSE) {
...
+    expect_identical(class(sum1a), SUM_CLASS_REF)
+    expect_identical(names(sum1a), SUM_NAMES_REF)
 
-  # submit model
-  proc1 <- submit_model(mod1, .mode = "local", .wait = TRUE)
-  expect_identical(class(proc1), PROCESS_CLASS_REF)
+    # get summary from process object
+    sum1b <- proc1 %>% as_model %>% model_summary()
+    expect_identical(class(sum1b), SUM_CLASS_REF)
+    expect_identical(names(sum1b), SUM_NAMES_REF)
```
callistosp: potential additional tests

   ```
 # get summary from numeric
    sum1c <- 1 %>% model_summary()
    expect_identical(class(sum1c), SUM_CLASS_REF)
    expect_identical(names(sum1c), SUM_NAMES_REF)

    # get summary from character
    sum1d <- "1" %>% model_summary()
    expect_identical(class(sum1d), SUM_CLASS_REF)
    expect_identical(names(sum1d), SUM_NAMES_REF)
```

dpastoor:
IMO given we want a broader strategy around options handling - we can keep this one as is for now and deprecate it quickly once we have a formal strategy. I don't want to hold up the release, and do want to get something in play there.

callistosp:
slap me if I'm thinking too object-orientedly, but would it make sense to also add a `get_model_directory()` function?

dpastoor:
you're on the right track - see the pmplots link - we need a structure that provides both getters and setters

PR #39, opened 2020-03-18
copy_model_from checks for existing model file and exposes .overwrite_model_file arg

seth127:
Fixing problem described in https://github.com/metrumresearchgroup/rbabylon/issues/37

User now has an argument that can overwrite existing control stream. If `.overwrite_model_file=FALSE` (default) the existing control stream will be used for the new model.


R/workflow.R
```diff
@@ -395,22 +405,18 @@ copy_nonmem_model_from <- function(
...
-                as.character(glue("$PROBLEM {new_mod_id} {.description}\n\n$")))
-
-    # read parent control stream
-    write_file(mod_str, get_model_path(.new_mod))
+  .new_model_path <- .new_mod %>% get_model_path()
+  .parent_model_path <- .parent_mod %>% get_model_path()
+
+  if (fs::file_exists(.new_model_path)) {
+    if (isTRUE(.overwrite_model_file)) {
+      warning(glue("Overwriting file at {.new_model_path} because `.overwrite_model_file=TRUE`"))
```
dpastoor: given overwrite= TRUE is a user forced setting, I would not message/warn by default here since the only way to get through this path is if the user explicitly sets it

R/workflow.R
```diff
@@ -395,22 +405,18 @@ copy_nonmem_model_from <- function(
...
+      copy_control_stream(.parent_model_path, .new_model_path, .update_model_file, .description)
+    } else {
+      warning(glue("File already exists at {.new_model_path} -- using existing file because `.overwrite_model_file=FALSE` (default)"))
+    }
   } else {
-    fs::file_copy(get_model_path(.parent_mod), get_model_path(.new_mod))
+    copy_control_stream(.parent_model_path, .new_model_path, .update_model_file, .description)
```
dpastoor: instead of nesting this in an else block - a better pattern around control flow is to exit early given an if doesn't work out. Basically the goal is to minimize the need for else statements so people can minimize the number of branches they have to reconcile with. Remind me to discuss this in our next pair programming session

R/workflow.R
```diff
@@ -425,6 +431,40 @@ copy_nonmem_model_from <- function(
...
 
 
+#' Copy a NONMEM control stream file
+#'
+#' Helper function to copy a NONMEM control stream file and optionally update the description, etc. in the new file.
+#' Note that any file existing at `.new_model_path` will be overwritten.
+#' @param .parent_model_path Path to the control stream to copy
+#' @param .new_model_path Path to copy the new control stream to
+#' @param .update_model_file Boolean for whether to update the `$PROBLEM` line in the new control stream. By default it is TRUE, but if FALSE is passed `{.new_model}.[mod|ctl]` will be an exact copy of its parent control stream.
+#' @param .description Description of new model run. This will be passed into the `$PROBLEM` of the new control stream (if `.update_model_file=TRUE`).
```
dpastoor: don't forget to export

seth127:
> This seems sufficient to me. One thought is that in the future it would be nice to build out `copy_control_stream` to grep all the potential instances of run number in the new control stream.

Agree. We have that captured here: https://github.com/metrumresearchgroup/rbabylon/issues/32 (though it won't be on this release)

But yes, I think this function is probably the right place to implement that, which was part of why I broke it out like I did.

R/workflow.R
seth127: I was actually thinking to keep this one private. I feel like we should encourage people to just use `copy_model_from()` which will call this internally. Can you think of a reason people would want this as a standalone?

R/workflow.R
seth127: makes sense, I'll change it

seth127: Actually... maybe I'm being dense but I don't see how to do this smoothly without an else block. Since we're not messaging on `.overwrite=TRUE` I changed it to this, which is at least only else instead of two:
```
  if (fs::file_exists(.new_model_path) && !isTRUE(.overwrite)) {
    # if .overwrite != TRUE, warn that file already exists
    warning(glue("File already exists at {.new_model_path} -- using existing file because `.overwrite=FALSE` (default)"))
  } else {
    copy_control_stream(.parent_model_path, .new_model_path, .update_model_file, .description)
  }
```

seth127:
@dpastoor made requested changes, although small question about the `else` block comment.

seth127:
This closes out all functionality for the first release.

PR #43, opened 2020-03-24
Cleaning up .drone.yml

seth127:
A few little cleanup things based on @shairozan 's comments


shairozan:
For clarity and communication, the reason that these builds were failing is actually an issue with the release from Babylon. Goreleaser borked the linux_amd64 binary, while the MacOS binary worked fine. I purged the assets and re-ran releaser which generated the release without issue.

PR #47, opened 2020-03-31
.bbi_args and .mode reordering

seth127:
Changing the order of .bbi_args and .mode arguments in submit_model so that .bbi_args can be passed positionally.

Note: this will be tagged as version 0.2.0 because it changes the API (potentially breaking changes).


seth127:
Tagged with `0.1.2.8000`
export is here `s3://mpn.metworx.dev/releases/rbabylon/0.1.2.8000`

Note again, this is intended to be included in the `0.2.0` release.

PR #52, opened 2020-04-03
Submit multiple models in batch

seth127:
Adding feature to submit multiple models in batch, as per  https://github.com/metrumresearchgroup/rbabylon/issues/27

NOTE: _summarizing_ models in batch is _not_ included in this PR.


seth127:
@dpastoor @callistosp this is ready for review. Do we want to add any examples or vignettes for this functionality?

R/submit-model.R
```diff
@@ -200,3 +203,272 @@ submit_nonmem_model <- function(.mod,
...
+) {
+
+  if (!is.null(.directory)) {
+    warning(paste(glue("Passed `.directory = {.directory}` to submit_models.bbi_nonmem_model().") ,
+                  "This argument is only valid when passing a path to `submit_models()`.",
+                  "`bbi_nonmem_model` objects were passed, so `.directory` inferred from `.mod${WORKING_DIR}`"))
+  }
+
+  # check that all elements are a model, and that all models are the same type
+  all_models_classes <- map(.mods, function(.x) { intersect(VALID_MOD_CLASSES, class(.x)) })
```
dpastoor, line 260: to check for a particular class you use `inherits`

`inherits(.x, VALID_MOD_CLASSES)`

R/submit-model.R
```diff
@@ -200,3 +203,272 @@ submit_nonmem_model <- function(.mod,
...
+        .description = as.character(glue("{.mod} passed directly to submit_model()")),
+        .model_type = c("nonmem")
+      )
+    })
+  } else {
+    stop(glue("Unsupported file type passed to submit_model(): `{.mod}`. Valid options are `.yaml`, `.mod`, and `.ctl`"))
+  }
+
+  # pass to submit_models.list
+  .model_type <- .mods[[1]][[YAML_MOD_TYPE]]
```
dpastoor, line 366: i'm not sure inference from the first model is the appropriate way to go - especially from a 'future' proofing

R/utils.R
```diff
@@ -110,6 +110,49 @@ format_cmd_args <- function(.args, .collapse = FALSE) {
...
 
+#' Builds list of unique parameter sets for multiple model bbi calls
+#' @importFrom purrr map_lgl
+#' @param .mods a list containing only `bbi_{.model_type}_model` objects
+#' @param .bbi_args A named list specifying arguments to pass to babylon. This will over-ride any shared arguments from the model objects.
+build_bbi_param_list <- function(.mods, .bbi_args = NULL) {
+
+  # check that everything in list is a model object
+  all_models_bool <- map_lgl(.mods, function(.x) {
+      valid_model_types <- intersect(VALID_MOD_CLASSES, class(.x))
```
dpastoor, line 121: inherits

R/submit-model.R
```diff
@@ -200,3 +203,272 @@ submit_nonmem_model <- function(.mod,
...
+    return(map(
+      cmd_args_list,
+      function(.run) { bbi_dry_run(.run$cmd_args, .run$model_dir) }
+      ))
+  }
+
+  # launch models
+  res_list <- map(
+    cmd_args_list,
+    function(.run, ...) { bbi_exec(.run$cmd_args, .wait = .wait, .dir = .run$model_dir, ...) }
```
dpastoor, line 470: no need to declare this function with the ..., those dots actually are going to attempt to inherit anyway from the ... in the map call.

R/submit-model.R
```diff
@@ -200,3 +203,272 @@ submit_nonmem_model <- function(.mod,
...
+  })
+
+  if (.dry_run) {
+    # construct fake res object
+    return(map(
+      cmd_args_list,
+      function(.run) { bbi_dry_run(.run$cmd_args, .run$model_dir) }
+      ))
+  }
+
```
dpastoor, line 466: an extra message here might be nice to actually specify in how many unique configurations models are getting submitted in. like "3 unique configurations identified for separate submission groups"

R/submit-model.R
seth127: I changed this a bit, but I still do want to do the intersect here because I need to check that they're all _the same type_ of model, but only that they're all valid model objects.

seth127: actually it doesn't need either any more because we're using `YAML_MOD_TYPE` instead

callistosp:
I will throw a bootstrap at this sometime today or tomorrow and let you know if I run into any issues.

dpastoor:
@callistosp seth is going to give a bootstrap a whirl - @seth127 have sam review your implementation when you have something working 

sam goal was to use the acop model + PKPDmisc::resample_df --> try 10/100/1000/10000 samples to see if everything takes ok

callistosp:
Even better, I would be happy to review it once it's working (or troubleshooting implementation as well)

PR #54, opened 2020-04-08
Release/0.3.0

seth127:
Will fill this in more later.

The main addition here is being able to submit multiple models in batch.


dpastoor:
I tagged it and released the new version

PR #56, opened 2020-04-13
Setting .wait=FALSE on submit_models throws an error

seth127:
Fixing the issue described here:
https://github.com/metrumresearchgroup/rbabylon/issues/55

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: rbabylon
 Type: Package
 Title: r package for babylon
-Version: 0.2.0.8005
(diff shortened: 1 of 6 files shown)


DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: rbabylon
 Type: Package
 Title: r package for babylon
-Version: 0.2.0.8100
+Version: 0.2.0.7101
```
dpastoor, line 4: since this is a fix associated with the 0.3.0 release, this would increment the build number, not downgrade the version --> 8100 --> 8101.

PR #57, opened 2020-04-13
Feature/bootstrapping dev

seth127:
This is a dev branch to try out ideas for bootstrapping implementation.


seth127:
Decided to just do this in adhoc function code instead of a dev branch, at least until we get it more ironed out.

PR #63, opened 2020-04-24
Feature/tidynm param table

seth127:
WIP: this is the replacement for the tidynm functions that are most used.
* getting parameter table with labels
* ~~possibly adding some stuff to the run_log (obj. function value, number of params, etc.)~~ _haven't done this at all, but I'd kinda like to. It should probably be a separate PR though._

Also added three test models (with output) that were test cases from `tidynm`. They're in `tests/testthat/data/tidynm_extdata/`


seth127:
@dpastoor just want to double check that you're cool with adding this to `param_estimates()` 
https://github.com/metrumresearchgroup/rbabylon/commit/e77885b965d6cfa9a88386a9a5798652aceb7596

(and that I did it right)

dpastoor:
> @dpastoor just want to double check that you're cool with adding this to `param_estimates()`
> [e77885b](https://github.com/metrumresearchgroup/rbabylon/commit/e77885b965d6cfa9a88386a9a5798652aceb7596)
> 
> (and that I did it right)

I think it looks reasonable, however it really matters what the scientists want. Not so bad to include it and let people trim it down. This will be much more targetted discussion once kyle has some bandwidth to dig in a bit

R/summary.R
```diff
@@ -221,13 +221,40 @@ param_estimates.bbi_nonmem_summary <- function(.summary) {
...
-  tibble::tibble(
+
+  param_df <- tibble::tibble(
     names = unlist(param_names),
     estimate = unlist(param_estimates),
     stderr = unlist(.summary$parameters_data[[num_methods]]$std_err) %||% NA_real_,
-    fixed = unlist(.summary$parameters_data[[num_methods]]$fixed),
+    random_effect_sd = c(
+      rep(NA, length(param_names$theta)),
+      unlist(.summary$parameters_data[[num_methods]]$random_effect_sd)),
```
dpastoor, line 231: These pieces likely need some `%||% NA_real_` portions as this `.summary$parameters_data[[num_methods]]$random_effect_sd` could be NULL

R/summary.R
```diff
@@ -221,13 +221,40 @@ param_estimates.bbi_nonmem_summary <- function(.summary) {
...
+#' Private helper to unpack an matrix index string like '(3,3)' is for a diagonal (i.e. if the numbers are the same)
+#' @param .name A character scaler containing an index string
+#' @importFrom stringr str_replace_all str_split
+is_diag <- function(.name) {
+  .ind <- .name %>%
+    str_replace_all(glue("^.*\\(|\\)"), "") %>%
+    str_split(",") %>% unlist()
+
+  if (length(.ind) == 1) {
+    invisible()
```
dpastoor, line 255: do you mean you want to `return(invisible())`

PR #64, opened 2020-04-27
WIP: refactoring into smaller files

seth127:
Refactoring the bloated workflow file (and a few others) into smaller more manageable files.


R/modify-model-field.R
```diff
@@ -0,0 +1,233 @@
...
+  # Either append new value or overwrite with new value
+  if (isTRUE(.append)) {
+    .mod[[.field]] <- c(.mod[[.field]], .value)
+  } else {
+    .mod[[.field]] <- .value
+  }
+
+  # de-duplicate values
+  if (isTRUE(.unique)) {
+    .mod[[.field]] <- .mod[[.field]] %>% unique()
```
dpastoor, line 30: I'll go into this more in a tech session, but its best to avoid pipes unless needed in programmatic contexts

```suggestion
    .mod[[.field]] <- unique(.mod[[.field]])
```

that is because the pipe is an extra function call, which means it adds an extra layer to stack traces as well as also makes profiling harder to digest

seth127: That makes sense. I have this pattern in... a lot of places.
```
$ grep -r \%\>\% rbabylon/R | wc -l
      64
```
We can talk at tech session and I'll make a cleanup PR if you think it's worthwhile. If not, I'll just keep this in mind moving forward.

PR #74, opened 2020-05-19
Feature/absolute paths

seth127:
Changes to `run_log()`, `config_log()` and some others to maintain absolute paths that will work correctly with nested modeling directories


seth127:
@callistosp @adamlafontaine The new code is in place and you should be able to accomplish https://github.com/metrumresearchgroup/rbabylon/issues/30 with something like the following:
```
# add config log to run log
log_df <- run_log() %>% add_config()

# get absolute paths to data and model files
norm_data_paths <- fs::path_norm(file.path(log_df$absolute_model_path, log_df$data_path))
norm_model_paths <- get_model_path(log_df)

# check config md5's against ctl md5's
log_df <- log_df %>% mutate(
  current_data_md5  = tools::md5sum(norm_data_paths),
  current_model_md5 = tools::md5sum(norm_model_paths),
  data_md5_match    = data_md5 == current_data_md5,
  model_md5_match   = model_md5 == current_model_md5
)
```

I still need to add/cleanup some unit tests before this is ready to merge, but feel free to look through the code and make comments.

seth127:
@dpastoor once this merges, should be call https://github.com/metrumresearchgroup/rbabylon/issues/30 done/closed? There is a little extra detail in there, but I think this is good enough.

**Related Questions:**
1. Should I go ahead and make a simple helper function for `check_data_hash` and `check_model_hash` that wraps the code in the previous comment? Or should we just leave that up to users?
2. Especially if we're _not_ gonna add a helper function, should I add this to a vignette? Maybe the bottom of the Getting Started vignette?

PR #75, opened 2020-05-21
Summary log

seth127:
This branch is up for discussion. It adds functionality to parse some fields from the `model_summary()` output and append them to the `run_log()`. This has been requested in various forms by scientists, but it may not be fully baked yet.


seth127:
Decided we're going to put this on ice until we have more clarity on what users would want in a summary_log

Issue #77, opened 2020-05-22
support both yml and yaml extensions


dpastoor:
Either are valid, and yaml is actually preferred through yaml spec, though R community seems to use .yml for most things.

backstory: https://stackoverflow.com/questions/21059124/is-it-yaml-or-yml

dpastoor:
https://github.com/metrumresearchgroup/rbabylon/blob/master/R/get-path-from-object.R#L145

start with a switch to `%in% c("yaml", "yml")`

We should consolidate all our tooling around `yaml` going forward.

* [ ] confirm examples to use `.yaml`
* [ ] confirm any generation/copying to do `.yaml`
* [ ] ?

seth127:
(Devin's original comment)

Either are valid, and yaml is actually preferred through yaml spec, though R community seems to use .yml for most things.

backstory: https://stackoverflow.com/questions/21059124/is-it-yaml-or-yml

PR #82, opened 2020-05-28
Fix/based on paths

seth127:
After making changes in https://github.com/metrumresearchgroup/rbabylon/issues/71 the `based_on` field is longer working because it only store run id's which are a) ambiguous and b) no longer included in the `run_log()` output (because of a).

This PR addresses the fix defined in https://github.com/metrumresearchgroup/rbabylon/issues/81


seth127:
@dpastoor I'm working on a vignette for this now that I'll probably finish in the morning. The code should be done though. 

Also, I've kinda touched the #77 issue here, but I think I just wanna make that a separate PR to keep things cleaner. They should probably be part of the same release though.

seth127:
Vignette is done. This is ready for full review and merge to develop if there are no changes.

R/modify-model-field.R
```diff
@@ -184,6 +184,50 @@ replace_bbi_args <- function(.mod, .bbi_args) {
...
+###################
+# helper functions
+###################
+
+#' Checks for yaml files relative to a starting location
+#' @importFrom fs file_exists is_absolute_path path_rel
+#' @importFrom purrr map_lgl
+#' @param .start The directory of the model (i.e. the YAML file) that the `based_on` will be added to.
+#' @param .based_on Character vector or scaler of paths (with or without extension) to the models that will be added to `based_on`. Paths should be relative to `.start` argument.
+check_based_on <- function(.start, .based_on) {
```
seth127, line 196: change to `safe_based_on`

R/get-based-on.R
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  return(sort(.results))
+}
+
+
+#' @rdname get_based_on
+#' @param .bbi_object Character scaler of a path to a model that can be loaded with `read_model(.bbi_object)`
+#' @export
+get_model_ancestry.character <- function(.bbi_object) {
```
seth127, line 143: think about memoise package

R/get-based-on.R
```diff
@@ -0,0 +1,169 @@
...
+}
+
+
+#' The default method attempts to extract the path from any object passed to it
+#' @importFrom fs path_norm
+#' @param .bbi_object The object to attempt to query. Could be a partially built bbi_{.model_type}_object or some other custom object containing model data.
+#' @rdname get_based_on
+#' @export
+get_based_on.default <- function(.bbi_object, .check_exists = FALSE) {
+  tryCatch(
```
seth127, line 24: @dpastoor I worked on refactoring this to just use get_path_from_object under the hood and rediscovered the reasons I didn't do it in the first place. Long story short: I'm gonna leave it like it is, but here are some of the reasons. Number 3 is the main reason.

1. I wanted to die if the key requested is absent, but for `based_on` it is perfectly reasonable for it to be absent, and the function should just return NULL instead of dying.
2. `get_path_from_object` expects the thing it's fetching to be a scaler, not a vector. I can get rid of this, but...
3. the `bbi_run_log_df` dispatch calls `map_chr` so that it can return a character vector of the paths fetched, one per row of the df. _But_ if what's fetched is _not_ a scaler (as in based_on) this fails, which means I need to use `map` instead, but makes the return from that dispatch either a) always a list or b) type-unstable. I don't like either one of those options.

Happy to discuss this further if you want to, but I'd be surprised if you really care. Just wanted to put my thoughts down here so you can follow the thought process. And so we can revisit later if you'd like to.

dpastoor: >  I'd be surprised if you really care. 

ouchies ;-)

Yes you are correct and your logic above makes complete sense, esp around map_chr's constrains around scalar returns.

R/get-based-on.R
seth127: Looked at `memoise` but I don't actually want to do this because the same input would have different results if the underlying yaml's have changed. Keeping track of that would get complicated quickly.

Did a refactor that sped things up dramatically. With the code below, it runs in half a second on 300 models, each with 3 different models in their `based_on` field.

```

MODS_TO_TEST <- 300
mod_names <- as.character(seq_len(MODS_TO_TEST))

mods <- map(mod_names, function(.m) {
  suppressSpecificWarning(new_model(yaml_ext(.m), .description = "fake"), .regexpr = "No model file found")
})

BASED_ON_COUNT <- 3
mods <- map(mods, ~ replace_based_on(.x, sample(mod_names, BASED_ON_COUNT)))


log_df <- run_log()
# View(log_df)

bo_bm <- rbind(
  benchmark(based_on_list <- log_df %>% get_based_on(), replications = 1),
  benchmark(based_on_list <- log_df %>% get_based_on(), replications = 10)
)
bo_bm
#                                        test replications elapsed relative user.self sys.self user.child sys.child
# 1 based_on_list <- log_df %>% get_based_on()            1   0.449        1     0.417    0.032          0         0
# 2 based_on_list <- log_df %>% get_based_on()           10   5.293        1     4.949    0.310          0         0

anc_bm <- rbind(
  benchmark(ancestry_list <- log_df %>% get_model_ancestry(), replications = 1),
  benchmark(ancestry_list <- log_df %>% get_model_ancestry(), replications = 10)
)
anc_bm
#                                               test replications elapsed relative user.self sys.self user.child sys.child
# 1 ancestry_list <- log_df %>% get_model_ancestry()            1   1.636        1     1.613    0.024          0         0
# 2 ancestry_list <- log_df %>% get_model_ancestry()           10  17.382        1    17.045    0.339          0         0
```

PR #84, opened 2020-05-29
Make both yaml and yml valid model files

seth127:
This PR is to address #77 

It was a original a branch of off https://github.com/metrumresearchgroup/rbabylon/tree/fix/based_on_paths but should merge to develop after that branch is merged (and may need to pull changes)


seth127:
There is a subtle change in behavior here, in the character dispatch of `submit_model()` (roughly lines 90-100 of `R/submit-model.R`.

When a path to a control stream was passed it to `submit_model()`...

Previously it would:
* error if a corresponding YAML was found
* call `new_model()` to create a corresponding YAML

Now it will:
* error if a corresponding YAML is _not_ found
* call `read_model()` on the corresponding YAML

This behavior is better because we want users to explicitly create and define their model objects,  not have them magically created by passing around paths to control streams. Passing a path to a control stream should be used only in the scenario when a model has already been created (i.e. the YAML exists on disk) but the control stream path is what the user happens to have handy (i.e. it was extracted from a run_log or something like that).

R/submit-model.R
```diff
@@ -323,42 +303,35 @@ submit_models.character <- function(
   all_ext <- tools::file_ext(.mods) %>% unique()
 
   if (length(all_ext) != 1) {
-    stop(paste(
-          "If passing character vector to submit_models(), must pass all the same type of file.",
-          glue("Got files with {length(all_ext)} different extensions: {paste(all_ext, collapse=', ')}"),
-          sep = " "
-         ))
+    # .yaml and .yml OR .mod and .ctl can coexist, but otherwise must all be the same.
```
dpastoor, line 306: per screenshare review - this extra carefulness is likely not needed as the map/read_model block will protect the user against invalid inputs

dpastoor: it would be good to add a test around demonstrating that we can pass these multiple types successfully

PR #94, opened 2020-06-16
Adding new file flags from bbi 2.2

seth127:
Adding `.bbi_args` flag(s) associated with https://github.com/metrumresearchgroup/rbabylon/issues/91


seth127:
This has pulled all the drone fixes from 0.6.1 and is ready for review, but it is branched off of https://github.com/metrumresearchgroup/rbabylon/tree/feature/constrain-bbi-version so ideally we should approve and merge https://github.com/metrumresearchgroup/rbabylon/pull/95 to `develop` first.

.drone.yml
```diff
@@ -5,7 +5,8 @@ name: rbabylon-40
 #Gobal variable definitions and anchors
 global-variables:
     environment: &default_environment
-        BABYLON_URL: https://www.github.com/metrumresearchgroup/babylon/releases/latest/download/bbi_linux_amd64.tar.gz
+        #BABYLON_URL: https://www.github.com/metrumresearchgroup/babylon/releases/latest/download/bbi_linux_amd64.tar.gz
```
copernican, line 8: Do we want to leave in commented-out code?

.drone.yml
```diff
@@ -53,7 +54,8 @@ name: rbabylon-36
...
```
copernican, line 57: Do we want to leave in commented-out code?

.drone.yml
```diff
@@ -101,7 +103,8 @@ name: rbabylon-36-old
...
```
copernican, line 106: Commented-out code

.drone.yml
```diff
@@ -149,7 +152,8 @@ name: rbabylon-35-old
...
```
copernican, line 155: Commented-out code

.drone.yml
```diff
@@ -195,7 +199,8 @@ name: rbabylon-release
...
```
copernican, line 202: Commented-out code

tests/testthat/test-summary.R
```diff
@@ -16,7 +16,7 @@ withr::with_options(list(rbabylon.bbi_exe_path = '/data/apps/bbi',
   MOD2_PATH <- file.path(MODEL_DIR, "2")
 
   # references
-  SUMMARY_REF_FILE <- "data/acop_summary_obj_ref_200520.rds"
+  SUMMARY_REF_FILE <- "data/acop_summary_obj_ref_200616.rds"
```
copernican, line 19: How is this file created?

seth127: `sum1 <- MOD1 %>% model_summary()`

Whenever one of these tests breaks I run this and compare `sum1` to `readRDS(SUMMARY_REF_FILE)` manually. Usually it's some little structural thing that's changed and is causing it to break. If they look good, I resave `sum1` out to a new reference.

Do you think we should document that somewhere?

.drone.yml
seth127: Good catch. This is all related to the issue mentioned here: https://github.com/metrumresearchgroup/rbabylon/pull/95#discussion_r442441634

When the babylon release has been made (before we merge) we will switch these back.

seth127:
> so how will this work with min version requirements. We now have a feature that is gated by 2.2 - the tricky extra piece is that this really is a supplemental feature that many/most will not need. IMO i would gate this behind a feature flag of sorts - if they try to use that flag we confirm the version was >= 2.2.0 and error if not

So, my feeling is that, given how few users we currently have, we should just gate this right here: rbabylon >= 0.7.0 requires babylon >= 2.2.0

The way it works right now it will check the version the first time the user tries to use `bbi_exec()` 
https://github.com/metrumresearchgroup/rbabylon/blob/feature/constrain-bbi-version/R/rbabylon.R#L101

At least at this stage, I can't think of any reason _not_ to tell everyone they just need to upgrade their bbi to 2.2.0, right?

Related: if we have more of these, where certain features/functions/args require a certain version of bbi, do we really want to get into this selective constraint thing, where certain functions throw an error based on version constraints and others don't? I don't love that.

copernican:
> Related: if we have more of these, where certain features/functions/args require a certain version of bbi, do we really want to get into this selective constraint thing, where certain functions throw an error based on version constraints and others don't? I don't love that.

Implementing a selective versioning strategy is definitely a pain, so if we can constrain the version at the package level, I think that's much easier all around. Selective versioning is most useful if you have a lot of users who for whatever reason cannot upgrade the dependency, and you also want to make new functionality (unrelated to the version constraint) available to them. Is that the case here?

tests/testthat/test-summary.R
copernican: Yes, I think that we should document the creation of any artifacts used in testing, but we can take that as a separate issue after we merge this.

PR #97, opened 2020-06-17
testing 06-08

seth127:
Getting all the tests to pass on the `2020-06-08` MPN snapshot. Details here: https://github.com/metrumresearchgroup/rbabylon/issues/98


copernican:
The CI failures were largely related to breaking changes in two dependencies.

1. `dplyr` 1.0.0 uses `vctrs` internally, and as a consequence requires that any class extending a data frame have its class name added first rather than last. The RDS objects used to test `param_labels()` had the class `nibble_param` in the wrong order. The class order was manually changed for each object, and the new objects committed to version control.

2. `tibble` 3.0.0 allows for elements of an individual column to be named. This caused tests for `param_estimates()` to fail because the constructed tibble had names and the reference tibble did not. This is resolved by passing `use.names = FALSE` to `unlist()`.

There were also some failures because some of the reference objects for `param_labels()` did not have the same set of classes as a tibble.

PR #99, opened 2020-06-18
cleanup docs

seth127:
Doing documentation cleanup per https://github.com/metrumresearchgroup/rbabylon/issues/92


seth127:
This is ready for review. You should be able to see it rendered with 
https://htmlpreview.github.io/?https://raw.githubusercontent.com/metrumresearchgroup/rbabylon/fix/docs_cleanup/docs/reference/index.html

seth127:
This has pulled all the drone fixes from 0.6.1 and is ready for review, but it is branched off of https://github.com/metrumresearchgroup/rbabylon/tree/feature/new-file-flags so ideally we should approve and merge #94 to develop first.

R/copy-model-from.R
```diff
@@ -2,7 +2,10 @@
 # Iterating on models
 #######################
 
-#' Generic S3 method from iterating on models.
+#' Create new model by copying existing model
+#'
+#' Create new model by copying existing model. Useful for iterating during model development.
+#' Also fills `based_on` field by default, for constructing model ancestry. See \href{../articles/using-based-on.html}{"Using based_on field" vignette} for details.
```
copernican, line 8: I think we need to [enable markdown support](https://roxygen2.r-lib.org/articles/markdown.html#turning-on-markdown-support) in `roxygen` comments for `based_on` and similar to render correctly.

R/copy-model-from.R
```diff
@@ -5,11 +5,11 @@
...
 #' @param .parent_mod Model to copy from
 #' @param .new_model Path to write new model files to WITHOUT FILE EXTENSION. Function will create both `{.new_model}.yaml` and a new model file based on this path.
-#' @param .description Description of new model run. This will be stored in the yaml (to be used later in `create_run_log()`).
-#' @param .based_on_additional Character scaler or vector of paths to other models that this model was "based on." These are used to reconstuct model developement and ancestry. \strong{Paths must be relative to `.new_model` path.} Note that the `.parent_model` will automatically be added to the `based_on` field, so no need to include that here.
+#' @param .description Description of new model run. This will be stored in the yaml (to be used later in `run_log()`).
+#' @param .based_on_additional Character scaler or vector of paths to other models that this model was "based on." These are used to reconstuct model developement and ancestry. **Paths must be relative to `.new_model` path.** Note that the `.parent_model` will automatically be added to the `based_on` field, so no need to include that here.
```
copernican, line 12: Typo: "scaler." In this case, though, it might be simpler to say

```
.based_on_additional Character vector of paths to other models...
```

because a vector can be of length 1, hence not necessary to comment that it can be scalar.

R/copy-model-from.R
```diff
@@ -5,11 +5,11 @@
...
 #' @param .add_tags A character scaler or vector with any new tags to be added to `{.new_model}.yaml`
```
copernican, line 13: Typo: "scaler." Again, we can probably just say "Character vector with any new..."

R/copy-model-from.R
```diff
@@ -5,11 +5,11 @@
...
 #' @param .inherit_tags Boolean for whether to inherit any tags from `.parent_mod`
```
copernican, line 14: The Boolean type is more commonly referred to by its R type name, `logical`. That said, it may be more helpful to phrase the description for logical arguments in terms of what will happen in the relevant cases, e.g., 

```
@param When `TRUE` (the default), the new model will inherit tags from `.parent_mod`. Set to `FALSE` to avoid this.
```

R/copy-model-from.R
```diff
@@ -5,11 +5,11 @@
...
 #' @param .update_model_file Boolean for whether to update the newly created model file. By default it is TRUE, but if FALSE is passed new model file will be an exact copy of its parent.
```
copernican, line 15: Similar comment about description for logical arguments.

R/param-labels.R
```diff
@@ -5,8 +5,8 @@
 #' However it will be extended to parse labels from the model YAML file as well.
 #' The syntax for the labeling is described in "Details" below.
 #'
-#' This function will *not* return indices for the parameters, though they can be added with \code{param_labels() \%>\% apply_indices()}.
-#' See the \code{\link{apply_indices}} documentation for more details.
+#' This function will *not* return indices for the parameters, though they can be added with `param_labels() \%>\% apply_indices()`.
```
copernican, line 8: This renders oddly (the escape character `\` is shown in the result). Maybe consider

```
...though they can be adding by calling `apply_indices()` on the output.
```

R/copy-model-from.R
seth127: changed all these

R/copy-model-from.R
seth127: Changed all these. It felt weird at first to not declare the type, but it does read nicer.

R/param-labels.R
seth127: this was leftover from changing to markdown parsing. Fixed it.

R/modify-model-field.R
```diff
@@ -8,9 +8,9 @@
 #' Also checks the object against the corresponding YAML before modifying (and errors if they are out of sync) and then writes the modified object back to the YAML.
 #' @param .mod The `bbi_{.model_type}_model` object to modify
 #' @param .field Character scaler of the name of the component to modify
```
copernican, line 10: Typo: "scaler." Consider "string" rather than "character scalar."

R/modify-model-field.R
```diff
@@ -41,7 +41,7 @@ modify_model_field <- function(.mod, .field, .value, .append = TRUE, .unique = T
 
 #' Add tags to a model object and corresponding YAML
 #' @param .mod The `bbi_{.model_type}_model` object to modify
-#' @param .tags Character scaler or vector of tags to add
+#' @param .tags Character vector of tags to add
```
copernican, line 44: This description will not appear in the documentation because we also have `@rdname modify_model_field`, so the description for `.tags` in the function of the same name will be used. If this description needs to be different, there are a few strategies:

1. Document in separate files (so, change or remove the `@rdname` tag).

2. Generalize the description for `.tags` in `modify_model_field()` to incorporate behavior relevant to any function with that argument that is documented in the same file.

3. Change the argument name, which will result in creating a new argument with the accompanying description.

Note also that if one function takes an argument whose description is the same as found in another function, we can use [`@inheritParams`](http://r-pkgs.had.co.nz/man.html#dry2) to use the documentation from that function, e.g., `@inheritParams modify_model_field`. If the functions share an argument of the same name, but its description differs, then it can be handled as above, or a new `@param` description added, which will override anything inherited from `@inheritParams`.

R/modify-model-field.R
```diff
@@ -54,7 +54,7 @@ add_tags <- function(.mod, .tags) {
 
 #' Replaces tags on a model object and corresponding YAML with new tags
 #' @param .mod The `bbi_{.model_type}_model` object to modify
-#' @param .tags Character scaler or vector of tags use as replacement
+#' @param .tags Character vector of tags use as replacement
```
copernican, line 57: Same comment about `.tags` description.

R/modify-model-field.R
```diff
@@ -80,7 +80,7 @@ add_decisions <- function(.mod, .decisions) {
 
 #' Replaces `decisions` field in a model object and corresponding YAML with new values
 #' @param .mod The `bbi_{.model_type}_model` object to modify
-#' @param .decisions Character scaler or vector to use as replacement
+#' @param .decisions Character vector to use as replacement
```
copernican, line 83: Similar comment as for `.tags` description.

R/modify-model-field.R
```diff
@@ -107,7 +107,7 @@ add_based_on <- function(.mod, .based_on) {
 
 #' Replaces `based_on` field in a model object and corresponding YAML with new values
 #' @param .mod The `bbi_{.model_type}_model` object to modify
-#' @param .based_on Character scaler or vector to use as replacement
+#' @param .based_on Character vector to use as replacement
```
copernican, line 110: Similar comment as for `.tags` description.

R/param-estimates.R
```diff
@@ -3,7 +3,7 @@
 # format NONMEM output to parameter tables
 ############################################
 
-#' S3 generic for parsing parameter estimate table
+#' Parses parameter estimates table from model_summary() output
```
copernican, line 6: Consider formatting (and linking) `model_summary()`.

R/param-labels.R
```diff
@@ -5,8 +5,8 @@
 #' However it will be extended to parse labels from the model YAML file as well.
 #' The syntax for the labeling is described in "Details" below.
 #'
-#' This function will *not* return indices for the parameters, though they can be added with \code{param_labels() \%>\% apply_indices()}.
-#' See the \code{\link{apply_indices}} documentation for more details.
+#' This function will *not* return indices for the parameters, though they can be added with `param_labels() %>% apply_indices()`.
+#' See the `apply_indices()` documentation for more details.
```
copernican, line 9: In a case like this, we can use the `@seealso` tag to include a link to related functions.

R/read-output.R
```diff
@@ -80,8 +81,8 @@ tail_output <- function(.mod, ...) {
 #' @param .mod Path to OUTPUT file
 #' @param .head Integer for number of lines to read from the top of the file
 #' @param .tail Integer for number of lines to read from the bottom of the file
-#' @param .print Boolean for whether to print resulting head and tail to console. Defaults to TRUE.
-#' @param .return Boolean for whether to return resulting head and tail as a character vector. Defaults to FALSE.
+#' @param .print If `TRUE`, the default, print resulting head and tail to console.
```
copernican, line 84: Opportunities for `@inheritParams` throughout this file.

R/copy-model-from.R
```diff
@@ -147,18 +150,19 @@ copy_model_from.numeric <- function(
 
 #' Copy model from an existing NONMEM model
 #'
+#' Private implementation function called by `copy_model_from()` dispatches.
```
copernican, line 153: General comment: no need to specify that a function is "private." `@keywords internal` will take care of that, and if in the future you decide to export the function, now you have to make two changes rather than one.

R/use-bbi.R
```diff
@@ -101,8 +102,10 @@ bbi_current_release <- function(os = "linux"){
 #' @param .this_os Character scaler of OS
 #' @param .dir directory to install bbi to
 #' @param .version version of babylon to install. Must pass a character scaler corresponding to a tag found in `https://github.com/metrumresearchgroup/babylon/releases`
-#' @param .force Boolean for whether to force the installation even if current version and local version are the same
-#' @param .quiet Boolean for suppressing output printed to the console. False by default.
+#' @param .force If `FALSE`, the default, skips installation if requested version and local version are the same. If `TRUE` forces installation if it will be the same version.
```
copernican, line 105: Opportunities for `@inheritParams`.

R/modify-model-field.R
seth127: Yes... this is a good call. I see several places where I used `rdname` haphazardly and created some sloppy docs. I'll correct that.

The 0.7.0 release is kinda being thought of as the "first public release" (because it's the first one most scientists will see) so I think it is worth spending the time to get some of this stuff right. Thanks for the fine-grained review.

seth127:
Ok, this is finally ready for final review. Worth the effort, but not the most fun I've had in recent memory.

copernican:
Running `pkgdown::build_site()` results in changes, for example `inst/nonmem/1.yaml` is deleted. Are you seeing changes as well? I am using `pkgdown` 1.4.1.

seth127:
> Running `pkgdown::build_site()` results in changes, for example `inst/nonmem/1.yaml` is deleted. Are you seeing changes as well? I am using `pkgdown` 1.4.1.

I didn't run `pkgdown::build_site()` on this last one, but you're right we should. It'll get run again when we do the release, but that's the right move. I'll do that and you can take a final look.

DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: rbabylon
 Type: Package
 Title: R package for babylon
-Version: 0.6.1.8001
+Version: 0.6.1.7002
```
copernican, line 4: Do we want to decrement the version?

R/classes.R
```diff
@@ -1,11 +1,13 @@
-# Constructor functions for objects of fundamental rbabylon classes
-
+#' Create bbi object
+#'
+#' Constructor functions for objects of fundamental rbabylon classes.
+#' @name create_bbi_object
+NULL
 
-#' Creates S3 object of class `bbi_{.model_type}_model` from list with `MODEL_REQ_INPUT_KEYS`
+#' @describeIn create_bbi_object Creates list object of class `bbi_{.model_type}_model` from named list with `MODEL_REQ_INPUT_KEYS`
```
copernican, line 7: Everything to be documented in `create_bbi_object` also has `@keywords internal`, which means that `?create_model_object` (for example) shows the documentation, but this documentation page is not included in the index. Just confirming that is the desired behavior.

R/get-based-on.R
```diff
@@ -49,9 +62,7 @@ get_based_on.default <- function(.bbi_object, .check_exists = FALSE) {
   return(as.character(fs::path_norm(file.path(.bbi_object[[WORKING_DIR]], .bbi_object[[YAML_BASED_ON]]))))
 }
 
-
-#' @rdname get_based_on
-#' @param .bbi_object Character scaler of a path to a model that can be loaded with `read_model(.bbi_object)`
+#' @describeIn get_based_on Takes a character scaler of a path to a model that can be loaded with `read_model(.bbi_object)`.
```
copernican, line 65: Typo: "scaler."

R/get-path-from-object.R
```diff
@@ -167,41 +176,40 @@ yml_ext <- function(.x) {
...
 ###################################
 
+#' Get model identifier
+#'
 #' Helper to strip path and extension from model file to get only model identifier
-#' @param .mod generic model
+#' (i.e. the model file name without extension)
+#' @param .mod Model to use, either a `bbi_{.model_type}_model` or file path to a model.
 #' @importFrom tools file_path_sans_ext
 #' @returns Character scaler with only model identifier
```
copernican, line 185: Typo: "scaler."

R/read-output.R
```diff
@@ -3,21 +3,24 @@
...
                         .mod[[YAML_OUT_DIR]],
                         paste0(get_model_id(.mod[[YAML_MOD_PATH]]), ".", .extension))
   return(out_file_path)
 }
 
+#' Check an output file
+#'
 #' Reads the head and tail of specified file and prints it the console and/or returns it as a character vector.
+#' This is called internally in several S3 methods described below.
 #' @param .file Character scaler of path to file to read
```
copernican, line 18: Typo: "scaler."

R/read-output.R
```diff
@@ -180,13 +152,14 @@ check_output_dir.bbi_nonmem_model <- function(.mod, ...) {
 }
 
 
-#' Checks nonmem output file that's a whitespace-delimited file (for instance .grd or .ext)
-#' @param .path Character scalar path to the gradient file
+#' Check NONMEM output files
+#'
+#' Checks a NONMEM output file that's a whitespace-delimited file (for instance .grd or .ext)
+#' @param .path Character scalar path to the file
```
copernican, line 158: Typo: "scaler."

DESCRIPTION
seth127: This is kind of a grey area in our versioning scheme. Technically yes. We saw to increment this dev version number when you _open_ the branch, and then pull it in when you merge to `develop` so that the version number in `develop` always reflects the last branch that was pulled. But yes, sometimes this does mean we decrement it. 

It doesn't bother me, but if you have strong feelings I'm open.

R/classes.R
seth127: yes, these are all private methods so no need to include them in the index. They're for developers only.

R/get-based-on.R
seth127: You would think I would've just grepped for this by now...

DESCRIPTION
copernican: I'm fine with leaving it as is.

PR #103, opened 2020-06-26
bbi patches for 0.7.0

seth127:
Began a refactor here: https://github.com/metrumresearchgroup/rbabylon/tree/feature/constrain-bbi-refactor but decided to save most of it for another release. These are the few changes we're pulling in to 0.7.0:

* Parse version constraint from DESCRIPTION file
* `bbi_exec()` becomes private


seth127:
This didn't work and there isn't time to debug it. Going to close this and deal with it in https://github.com/metrumresearchgroup/rbabylon/tree/feature/constrain-bbi-refactor

PR #109, opened 2020-07-08
Feature/use jsonnet

copernican:
This PR introduces a Jsonnet configuration for Drone. This will allow us to more easily define matrix builds (rather than needing to specify each build separately in `.drone.yml`).

A key consideration for `rbabylon` is the need to have a babylon executable (`bbi`) available for tests and building vignettes. The existing tests use `withr::with_options()` to specify the path to `bbi` as `/data/apps/bbi`. This means that whether the test suite passes (and the vignettes can build) depends on having `bbi` at that path on the Drone worker.

While this requirement can be satisfied by installing `bbi` at that path in Drone, the goal of moving to Jsonnet is to minimize the number of steps that must be customized for a particular repository. To remove the coupling between a particular `bbi` path and the `with_options()` pattern, this PR introduces a helper function `read_bbi_path()`, which attempts to read the `bbi` path from an environment variable (`BABYLON_EXE_PATH`), and defaults to `/data/apps/bbi` if unset. We also take the opportunity to pass the `bbi` path to `bbi_exec()` explicitly.

This approach results in new patterns.

- Local development

  - If `bbi` is not at `/data/apps/bbi`, then `BABYLON_EXE_PATH` must be set to the `bbi` path, e.g., with `Sys.setenv()`. `devtools::check()` will inherit this variable, including through the RStudio Build pane.

   - If `bbi` is at `/data/apps/bbi`, then local development can proceed unchanged.

- Continuous integration: `BABYLON_EXE_PATH` must be set in `.drone.jsonnet`.


R/rbabylon.R
```diff
@@ -47,7 +60,7 @@ bbi_exec <- function(.cmd_args, .dir = ".", .verbose = FALSE, .wait = FALSE, ...
   res <- list()
   res[[PROC_PROCESS]] <- p
   res[[PROC_STDOUT]] <- output
-  res[[PROC_BBI]] <- bbi_exe_path
+  res[[PROC_BBI]] <- .bbi_exe_path
```
seth127, line 63: oops, were we not testing this?

vignettes/getting-started.Rmd
```diff
@@ -50,22 +50,23 @@ The _first time_ you use `rbabylon` on a system or disk, you must first install
 
 To reiterate, there are two relevant paths for this section.
 
-* `BBI_EXE_PATH` -- the path the `babylon` executable file on your system.
+* `BBI_EXE_PATH` -- the path to the `babylon` executable file on your system.
 * `MODEL_DIR` -- the root modeling directory where you will be running models with `rbabylon`
 
 ```{r define dirs}
-BBI_EXE_PATH <- "/data/apps/bbi" # this should be an absolute path
+BBI_EXE_PATH <- read_bbi_path()
```
seth127, line 57: I think we should actually have two different blocks here:
1. this new one, but set to `include = FALSE`
2. the old one, but set to `eval = FALSE`

My reasoning is that we're showing the user how to set this and we _don't_ actually want them to set it with `read_bbi_path()`, we want them to set it explicitly (ideally in their .Rprofile, as we describe later).

vignettes/getting-started.Rmd
```diff
@@ -50,22 +50,23 @@ The _first time_ you use `rbabylon` on a system or disk, you must first install
...
 
 You can install the `babylon` binary with:
-```
+
+```{r, eval = FALSE}
 rbabylon::use_bbi()
 ```
 
-This will install to the `/data/apps/` folder by default, which matches the `BBI_EXE_PATH` you just defined (and will use soon). If you have changed that, you can pass your custom path to the _directory_ where you want to put bbi like so `use_bbi("/some/other/dir")`. In that case, you would want to set `BBI_EXE_PATH = "/some/other/dir/bbi"`.
+This will install to the `/data/apps/` folder by default, which will usually match the `BBI_EXE_PATH` you just defined (it will not match only if the `BABYLON_EXE_PATH` environment variable is deliberately set to a different path). If you have changed that, you can pass your custom path to the _directory_ where you want to put bbi like so `use_bbi("/some/other/dir")`. In that case, you would want to set `BBI_EXE_PATH = "/some/other/dir/bbi"`.
```
seth127, line 69: yea, unless you strongly disagree, I think we should revert this back to the original. No one but developers and _maybe_ very advanced users should be using this environment variable. This is a "Getting Started" vignette. Let's try to keep it as simple as possible. Setting this path up already seems to confuse some users.

R/rbabylon.R
copernican: This PR adds `.bbi_exe_path` as an argument to `bbi_exec()`. The existing implementation calls `getOption()` within `bbi_exec()` and assigns the result to `bbi_exe_path`. By making the path an argument, we no longer need the internal `getOption()` call. The package convention prefixes argument names with `.`, so this assignment should be for the argument `.bbi_exe_path` rather than the internal variable `bbi_exe_path`.

vignettes/getting-started.Rmd
copernican: Good point, I've made this change.

R/rbabylon.R
seth127: ah ok. I follow.

seth127:
Looks good

PR #111, opened 2020-07-21
[do not merge] Feat/new ci images

copernican:
This PR will point Drone at the new CI images. `v1.3` is used for initial testing.

- With the new images, we will abstract away the specific R patch versions. The major.minor version will be specified, and Drone will use environment variables in the image to resolve the patch version.

- This PR also includes a lint step. Bringing the package into compliance with the default linters will take some effort, so we can either do the investment now, temporarily disable this step, or accept an overall failing status on the pipeline until we do.

**This PR should not be merged until the new set of CI images is built.**

FYI @evol262 @bkyoung @dpastoor

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -8,14 +8,6 @@ local name = "rbabylon";
 # where to mount a temporary volume; useful for persisting files across steps
 local temp_volume_dir = "/ephemeral";
 
-# R versions to test against; first element will be used to build release
-local r_versions = [
-  "4.0.1",
-  "3.6.3",
-  "3.6.2",
-  "3.5.3",
-];
-
 # environment variables available to every R call, including devtools::check();
 # R_LIBS_USER will be set separately; use the empty object {} if there are no
 # variables to pass
@@ -50,6 +42,14 @@ local exclude_events = [
 ########################################
 ##        START SETTINGS BLOCK        ##
 ########################################
+
+# R major.minor versions to test against
+# - first element will be used to lint package and build release
+# - must be updated when versions in images are updated
+local r_versions = [
+  "4.0",
+  "3.6",
+];
 
 local bbi_url_base = "https://github.com/metrumresearchgroup/babylon/releases/download";
 local bbi_artifact_name = "bbi_linux_amd64.tar.gz";
@@ -98,10 +98,10 @@ local create_build_tag(name, r_major_minor, mpn_snapshot) =
 # tag             tag name
 local create_mpn_image(tag) = std.join(":", [mpn_image_base, tag]);
 
-# Get the major.minor version of R
-#
-# r_version       R version to use, as major.minor.patch, e.g., "4.0.1"
-local get_major_minor(r_version) = r_version[0:3];
+# Get the name of the environment variable holding the path to the R executable
+# r_major_minor   R version to use, as major.minor, e.g., "4.0"
+local get_r_exe_var(r_major_minor) =
+  std.join("_", ["R", "EXE"] + std.split(r_major_minor, "."));
 
 ########################################
 ##        END UTILITIES BLOCK         ##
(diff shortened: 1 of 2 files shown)


.drone.jsonnet
```diff
@@ -24,12 +16,13 @@ local r_env_vars = {
...
 # https://console.aws.amazon.com/ecr/repositories/mpn-dev/?region=us-east-1
-local mpn_snapshots = [
-  "cran-latest",
-  "2020-05-27",
-  "2020-03-24",
+# generally, only the second element should be modified
+local ci_images = [
+  "mpn-dev:latest",      # latest MPN snapshot
+  "mpn-dev:2020-03-24",  # oldest compatible snapshot
+  "cran-latest:latest",  # latest MPN snapshot for MRG packages + current CRAN
```
seth127, line 25: Maybe these comments are just confusing but can you explain a little bit more about what the first and third snapshots are, and how often they will be refreshed?

copernican: The idea is that the package will be tested against three images, described in the table below. The maintainer should only have to modify the second element, because only that one is expected to vary by package.

image | purpose | can vary by package? | changes when?
------|---------|------------------------|--------------
latest MPN snapshot | all MRG packages should work with the latest snapshot | no | when new snapshot is released
oldest compatible MPN snapshot | define the range of snapshot compatibility (through the current snapshot) | yes | images will be rebuilt periodically, but R package versions will not change
"CRAN latest" | MRG packages from the latest snapshot, plus non-MRG packages from the current source, e.g., current CRAN release of `dplyr`; this is the "canary in the coal mine" that should alert us to breaking changes in non-MRG packages so that we can update the MRG package being tested | no | triggered by change in upstream source, e.g., new version released on CRAN, or new MPN snapshot

copernican:
> > We cannot test the Jsonnet configuration until we change the setting in Drone, which probably should not happen until we merge to `master`...
> 
> I see your point here, but I strongly think we should turn it on and see this pass before we merge to develop. Then, if you want, we can turn it back off. But I also think we should turn it back on when we open the release branch, to make sure everything is passing there with the new config, before we merge.
> 
> I'm also not opposed to just leaving this setting on and everyone doing development needs to be sure to pull the jsonnet from `develop` or their stuff will fail. Maybe that's a bad idea, but honestly I kinda like it better than having to remember to do it later when we cut the release branch. Also, it will make the drone build time a lot faster (because the images are smaller), right? Your call on that one.

I'm happy to switch to Jsonnet now, which would imply

1. No CI for `master` until `develop` is merged, but that's fine, as we shouldn't have anything merged into `master` that doesn't go through `develop`.

2. Rebasing existing branches on `develop`. I think this a reasonable cost, provided those branches are not too far behind `develop` in terms of non-CI commits.

copernican:
> > Bringing the package into compliance with the default linters will take some effort...
> 
> My vote here is to disable the pipeline step for now and make an issue in the repo for getting the linter to pass. Even if you want to take that on right now, it feels like it should be a separate PR. I do _not_ support turning it on and having a failing build in the repo until we fix it.

The linting step is (temporarily) disabled. Agreed it should be a separate PR.

.drone.jsonnet
seth127: For all of these, are we doing the "dev" subset that is only the packages we know are used dependencies for MRG packages, or are we doing the full snapshot? My impression was that we were switching to the former so that build would be faster. Is that right? If not, are we still planning to do that at some point?

> "CRAN latest" ... triggered by change in upstream source

I'm not sure I follow. Are we actually gonna be monitoring CRAN and rebuilding this whenever one of our dependencies updates? If so... cool. If not, I just want to make sure this isn't manual (like someone checking CRAN and manually triggering an update). Like, it would be cool for this happen at least once a week or so. Unless it's automatic in which case I love it.

copernican: > For all of these, are we doing the "dev" subset that is only the packages we know are used dependencies for MRG packages, or are we doing the full snapshot? My impression was that we were switching to the former so that build would be faster. Is that right? If not, are we still planning to do that at some point?

Yes, these images include only a subset of external (CRAN, Bioconductor) packages, including the dependencies of MRG packages.

> > "CRAN latest" ... triggered by change in upstream source
> 
> I'm not sure I follow. Are we actually gonna be monitoring CRAN and rebuilding this whenever one of our dependencies updates? If so... cool. If not, I just want to make sure this isn't manual (like someone checking CRAN and manually triggering an update). Like, it would be cool for this happen at least once a week or so. Unless it's automatic in which case I love it.

Yes, the plan is to have a job that watches upstream sources and triggers a rebuild when a dependency changes. We're aiming to automate as much of this as possible, and in particular, the plan is definitely _not_ to have this check be manual.

copernican:
I switched the repo to Jsonnet, and the build [is passing](https://github-drone.metrumrg.com/metrumresearchgroup/rbabylon/912).

.drone.jsonnet
seth127: Sounds good. Thanks for clarifying.

PR #114, opened 2020-07-22
feature/batch parameter estimates

seth127:
Implementing https://github.com/metrumresearchgroup/rbabylon/issues/113


seth127:
I'm probably going to need to do something similar to what happened to https://github.com/metrumresearchgroup/rbabylon/pull/108 with this before we try to do a real review and merge. This dev work got a bit tangled while iterating with scientists on what the functionality should be.

seth127:
I moved these changes to a new branch because of some tangled old commits. They are now here https://github.com/metrumresearchgroup/rbabylon/pull/120

Issue #113, opened 2020-07-22
Batch parameter estimates


seth127:
# Summary

As a user I would like to be able to pass multiple model summary objects, or a summary log tibble, to the `param_estimates()` function and get back a table containing the parameter estimates for each model. Specifically: a long table with a columns for:

* model identifier
* parameter name
* parameter estimate

# Tests
- tests/testthat/test-param-estimates.R
  - param_estimates.bbi_summary_list gets expected table
  - param_estimates.bbi_summary_log_df gets expected table

copernican:
> I would like to be able to pass multiple models (or model summary objects)

How much flexibility do we need? `param_estimates()` currently takes only a `bbi_{.model_type}_summary` object. Do we want to build in the flexibility to take models as well, or can we provide a workflow that looks like

```r
mods %>%
  model_summaries() %>%
  param_estimates()
```

which simplifies what this function needs to do. If we do need to support models, do we need to support the four types of arguments allowed for the `.mods` argument to `model_summaries()`? Is the user allowed to pass a list that includes different types of model specifications, e.g., one element is a model summary, another is a character?

> to the `param_estimates()` function

Does this mean that we will change the existing function signature, so that `.summary` can have multiple elements? Or are we contemplating writing a new function that essentially wraps `map()`? Or some other option?

> get back a table with each parameter as a column and each model as a row

This tibble could be quite wide. And, is it possible that not all parameters will be available for all models (leading to `NA`s)? Would it be better to return a tibble whose rows are something like

```
model_id | parameter | value
```

and then let the user pivot that table as desired?

seth127:
Good catch, that was written wrong: we will _not_ be accepting models directly. I changed the original. It now reads:
`pass multiple model summary objects, or summary log tibble`

To the other question, we will need to get some clarification from the scientists. Questions that need answers:

* Will you ever want to run this on models with different numbers of params (or different param definitions)? 
* If so, what should happen? Would you want wide with NA's or would a long table be better? 
* If not, should the function error if there are different numbers of params? That feels aggressive, but if it really should _never_ happen then that's probably safest.

seth127:
@copernican in response to

> This tibble could be quite wide... let the user pivot that table as desired?

[...]

PR #108, opened 2020-07-01
Feature/summary log dev

seth127:
I have been working on this with the Matt R and 10am group and @aspirinpkpd (Ana R) specifically.

There are some ongoing tweaks that are being made, but I would like to merge this to `develop` before it becomes _too_ monstrous (maybe too late?) and then make those tweaks on additional branches. 

The functionality outlined in https://github.com/metrumresearchgroup/rbabylon/issues/53 and https://github.com/metrumresearchgroup/rbabylon/issues/67 is covered in this PR.

PR #120, opened 2020-08-13
feature/batch parameter estimates real

seth127:
Implementing #113 

I did something similar to what happened to https://github.com/metrumresearchgroup/rbabylon/pull/108#issuecomment-663709424 with this (copying from #114) because the dev work got a bit tangled while iterating with scientists on what the functionality should be (and I was branched off that #108 branch originally in #114).


R/param-estimates.R
```diff
@@ -53,11 +77,44 @@ param_estimates.bbi_nonmem_summary <- function(.summary) {
     ) %>%
     tibble::as_tibble()
 
-  param_df[["diag"]] <- map_lgl(param_df[["names"]], is_diag)
+  param_df[["fixed"]] <- map_lgl(param_df[["fixed"]], ~ .x == 1)
```
copernican, line 80: Because `==` is vectorized, this could be more simply expressed as `param_df[["fixed"]] == 1`.

R/param-estimates.R
```diff
@@ -53,11 +77,44 @@ param_estimates.bbi_nonmem_summary <- function(.summary) {
...
+
+#' @describeIn param_estimates Takes a `bbi_summary_list` object, the output of [model_summaries()].
+#' @importFrom purrr map_df
+#' @importFrom dplyr mutate select
+param_estimates.bbi_summary_list <- function(.summary) {
+
+  param_df <- map_df(.summary, function(.s) {
+    if(!is.na(.s[[SL_ERROR]])){
+      warning(paste("Missing summary for", .s[[ABS_MOD_PATH]], "\n", .s[[SL_ERROR]]))
+      return(invisible())
```
copernican, line 95: Why `return(invisible())` rather than `return(NULL)`?

R/model-summaries.R
```diff
@@ -190,3 +190,39 @@ model_summaries.bbi_run_log_df <- function(
...
+###################
+# helper functions
+###################
+
+#' Convert object to `bbi_summary_list`
+#'
+#' This is used to convert an object containing `bbi_{.model_type}_summary` objects into a `bbi_summary_list`.
+#' Currently it is only used for converting a `bbi_summary_log_df` into a `bbi_summary_list`
+#' (primarily so that it can more easily be mapped over), but theoretically it could be used for other purposes in the future.
+#' Note this is primarily intended as a developer functions, though it was exposed because users may have a use for it as well.
```
copernican, line 203: typo: `this is...intended as a...functions`

R/model-summaries.R
```diff
@@ -190,3 +190,39 @@ model_summaries.bbi_run_log_df <- function(
...
+  # create list of lists
+  .sum_list <- group_split(.sums, rn = row_number())
+  .sum_list <- map(.sum_list, function(.row) {
+    .row <- select(.row, -.data$rn)
+    .row <- as.list(.row)
+    .row[[SL_SUMMARY]] <- .row[[SL_SUMMARY]][[1]] # this gets buried one level deep by the as.list call
+    return(.row)
+  })
+
+  .sum_list <- create_summary_list(.sum_list)
```
copernican, line 225: This implementation is a little confusing because the same name is used for assignment multiple times (both `.sum_list` and `.row`). Would it be clearer with different names, or using `%>%`?

R/classes.R
```diff
@@ -150,47 +150,72 @@ create_process_object <- function(res) {
...
+#' @describeIn create_bbi_object Create tibble object of `bbi_config_log_df` class, first checking that all the required columns are present.
+#' @keywords internal
+create_config_log_object <- function(log_df) {
+  log_df <- create_log_df_impl(log_df, CONF_LOG_CLASS, CONFIG_LOG_REQ_COLS, ABS_MOD_PATH)
+  return(log_df)
+}
+
+#' @describeIn create_bbi_object Create tibble object of `bbi_summary_log_df` class, first checking that all the required columns are present.
+#' @keywords internal
+create_summary_log_object <- function(log_df) {
```
copernican, line 173: Are all of the `create_*_log_object()` functions necessary to implement the new functionality for `param_estimates()` (as described in #113)? I can see the case for `create_summary_log_object()`, and the desire to have a unified approach to creating such objects, but I'm not sure the other functions are strictly necessary. If this is the case, can they be moved into a separate PR that focuses just on unifying `*_log_object` creation?

tests/testthat/test-param-estimates.R
```diff
@@ -1,31 +1,59 @@
-context("Test bbi summary functions")
+context("Test param_estimates functions")
 
 if (Sys.getenv("METWORX_VERSION") == "" && Sys.getenv("DRONE") != "true") {
   skip("test-param-estimates only runs on Metworx or Drone")
 }
 
+setup({
```
copernican, line 7: `testthat` 3e [will deprecate](https://testthat.r-lib.org/articles/third-edition.html#changes-1) `setup()` and `teardown()`, now may be the time to start thinking about another pattern, i.e., test fixtures.

R/classes.R
seth127: They are not strictly related, they're a refactor that happened naturally as part of development. I was trying to dispatch on some classes and then realized those classes didn't exist and then realized there was no standard way to create them consistently. So I created one.

R/model-summaries.R
seth127: yea, I see that. This is kind of a preference thing though. I've heard the argument that using pipes in these situations is (infinitesimally) slower and also makes the stack traces a bit confusing when debugging. This argument advocates doing it this way. I googled for a link but couldn't find one so maybe I hallucinated this argument. It makes sense to me though. (And, if you're not gonna use pipes, I think it's much less confusing to do it this way than renaming the variable each time.)

R/param-estimates.R
```diff
@@ -5,24 +5,48 @@
...
+#' Details about the tibble tibble that is returned are in the return value section below.
+#'
+#' @return
+#' **Single model:** For `bbi_{.model_type}_summary` dispatch returns a tibble with the following columns:
+#'
+#' * parameter_names -- Parameter name ("THETA1", "THETA2", etc.)
+#' * estimate -- Parameter estimate
+#' * stderr -- Standard Error
+#' * random_effect_sd
+#' * random_effect_sdse
```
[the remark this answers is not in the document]
seth127: Oh yea I missed those. That's a good point. It's buried in the NONMEM docs, but it's not easy just to link to (because those docs aren't in any central public place that I know of). I'll add the description though.

tests/testthat/test-param-estimates.R
seth127: oof. This is used in a few places. I shall "start thinking" about it and we'll cross that bridge when we decide to switch. Functionally, this won't be hard to replicate with fixtures.

R/model-summaries.R
copernican: Both performance and stack traces should improve with [`magrittr` 2.0](https://www.tidyverse.org/blog/2020/08/magrittr-2-0/).

Issue #123, opened 2020-08-24
Update bbi flags


seth127:
There are flags that can be passed to bbi through `.bbi_args` and these are checked against [a constant defined in aaa.R](https://github.com/metrumresearchgroup/rbabylon/blob/master/R/aaa.R#L4) to make sure they exist. These have gotten out-of-date so they need to be updated to reflect the current flags available in `bbi nonmem --help`.

seth127:
Verified these are the flags as of the `v2.3.0-rc1` commit:
```
✔ ~/go/src/github.com/metrumresearchgroup/babylon [develop|✔] 
15:21 $ git rev-parse HEAD
d15ef96c4926a84169632a6f37ece32cbaf883a1
✔ ~/go/src/github.com/metrumresearchgroup/babylon [develop|✔] 
15:31 $ go build -o /usr/local/bin/bbi cmd/bbi/main.go 
✔ ~/go/src/github.com/metrumresearchgroup/babylon [develop|✔] 
15:32 $ /usr/local/bin/bbi nonmem --help

run nonmem model(s), for example: 
bbi nonmem run <local|sge> run001.mod
bbi nonmem run  --clean_lvl=1 <local|sge> run001.mod run002.mod
bbi nonmem run <local|sge> run[001:006].mod // expand to run001.mod run002.mod ... run006.mod local
bbi nonmem run <local|sge> .// run all models in directory
 

summarize model(s), for example: 
bbi nonmem summary run001/run001
bbi nonmem summary run001/run001.lst
bbi nonmem summary run001/run001.res
 

load .cov and .cor output from model(s), for example: 
bbi nonmem covcor run001/run001
bbi nonmem covcor run001/run001.cov

Usage:
  bbi nonmem [flags]
  bbi nonmem [command]

Available Commands:
  clean       clean files and folders
  covcor      load .cov and .cor output from a model run
  probs       summarize information about project
  reclean     clean files in an estimation directory by clean level
  run         run a (set of) models locally or on the grid
  scaffold    scaffold directory structures
  summary     summarize the output of model(s)

Flags:
      --background             RAW NMFE OPTION - Tells nonmem not to scan StdIn for control characters
  -h, --help                   help for nonmem
      --licfile string         RAW NMFE OPTION - Specify a license file to use with NMFE (Nonmem)
      --maxlim int             RAW NMFE OPTION - Set the maximum values set for the buffers used by Nonmem (default 100)
[...]

PR #126, opened 2020-08-27
Fix/update bbi flags

seth127:
Implementing #123


seth127:
I put a comment here about which flags to update: https://github.com/metrumresearchgroup/rbabylon/issues/123#issuecomment-682173435

Also, it's a little unnecessary and unrelated, but I changed `print_nonmem_args`, `check_nonmem_args` and the constant `NONMEM_ARGS` to their `bbi_` variants for two reasons:

1. It more accurately reflects what they actually do/are. I wrote this very early on and calling them `nonmem_` was just a naive misnomer.
2. It matches the `.bbi_args` argument to various functions, which is what they are related to. Now, if you want to know what you can pass to `.bbi_args` you will call `print_bbi_args()` instead of `print_nonmem_args()`.

seth127:
This is currently failing CI because the changed flags reflect changes in the new release of bbi that is necessary for https://github.com/metrumresearchgroup/rbabylon/issues/122 (which has not yet been implemented).

I will circle back to this PR once I implement https://github.com/metrumresearchgroup/rbabylon/issues/122 and everything should pass then.

copernican:
Why does `check_nonmem_args()` exist? If one of the arguments is invalid, could we instead bubble up the error from the `bbi` call? That would avoid having to keep the list of valid arguments in sync between the two projects.

seth127:
I would consider this. One of the main reasons we did this is that bbi used to have a mix of snake_case, dash-case, and camelCase, additionally having the CLI-style `--` or `-` prefix (like `--no-grd-file`) and we wanted to standardize it all to snake_case in R so this was kind of a mapping table.

At this point though, that's probably not needed and we can just change anything that's inconsistent on the bbi side. 

To move forward I would want:

* a defined strategy on argument mapping (dash vs snake)
* an issue in bbi to standardize args accordingly
* an issue in bbi to bubble back more user friendly (and less duplicative/messy) errors

R/aaa.R
```diff
@@ -3,36 +3,38 @@ BBI_DEFAULT_PATH <- "/data/apps/bbi"
...
 CACHE_ENV$bbi_exe_paths <- list()
 
-NONMEM_ARGS = list(
-  cache_dir = list(type = "character", flag="--cache_dir", description = "directory path for cache of nonmem executables for NM7.4+"),
-  cache_exe = list(type = "character", flag="--cache_exe", description = "name of executable stored in cache"),
+BBI_ARGS = list(
+  additional_post_work_envs = list(type = "character", flag="--additional_post_work_envs", description = "Any additional values (as ENV KEY=VALUE) to provide for the post execution environment"),
+  background = list(type = "logical", flag="--background", description = "RAW NMFE OPTION - Tells nonmem not to scan StdIn for control characters"),
   clean_lvl = list(type = "numeric", flag="--clean_lvl", description = "clean level used for file output from a given (set of) runs (default 1)"),
   config = list(type = "character", flag="--config", description = "config file (default is $HOME/babylon.yaml)"),
```
copernican, line 10: This does not match the description from 2.3.0:

```sh
➜  ~ /data/apps2/bbi nonmem run --help | grep \\-\\-config
      --config string                       Path (relative or absolute) to another babylon.yaml to load
```

R/modify-model-field.R
```diff
@@ -128,7 +128,7 @@ replace_description <- function(.mod, .description) {
 
 #' @describeIn modify_model_field Modifies model object and corresponding YAML by adding new `bbi_args`,
 #' overwriting any args that are already present with the new values.
-#' Use `print_nonmem_args()` to see a list of valid babylon arguments.
+#' Use `print_bbi_args()` to see a list of valid babylon arguments.
```
copernican, line 131: Suggest creating a link.

R/modify-model-field.R
```diff
@@ -150,7 +150,7 @@ add_bbi_args <- function(.mod, .bbi_args) {
 }
 
 #' @describeIn modify_model_field Modifies model object and corresponding YAML by replacing `bbi_args` with new list passed to `.bbi_args`.
-#' Use `print_nonmem_args()` to see a list of valid babylon arguments.
+#' Use `print_bbi_args()` to see a list of valid babylon arguments.
```
copernican, line 153: Suggest creating a link.

copernican (changes requested):
Looks good generally, just a few questions/suggestions.

copernican (approved):
LGTM. Should we bump the version? I'm okay if we just wait for the 0.8.0 release.

PR #129, opened 2020-08-28
Remove covcor from summary

seth127:
Implementing https://github.com/metrumresearchgroup/rbabylon/issues/128 to remove the covariance and correlation sections from the summary object.


seth127:
Turns out this is exceedingly simple because no downstream code touched these elements. We only had to modify the tests that check to see that they are created when we expect them (which we no longer do).

copernican (changes requested):
These changes are fine, but we missed something in #154. Specifically, the test that `check_bbi_version_constraint()` works with semantic versioning is failing. I would support removing the test in this PR (it's an internal function), and in a future PR rethinking the way we make `check_bbi_version_constraint()` aware of the minimum version.

copernican (approved):
LGTM. Again, we are accepting the risk of merging into `develop` despite the build failing. As discussed in #154, we expect the build to succeed after merging this, and then #137.

PR #138, opened 2020-09-04
export s3 methods for param_estimates and as_summary_list

seth127:
Oops we forgot to export the s3 methods that were created in https://github.com/metrumresearchgroup/rbabylon/pull/120


copernican (approved):
LGTM

PR #148, opened 2020-09-09
feature: config_log() shows model and data status 

### LINKED ISSUE #30: User can check which models are up-to-date from config_log
# Summary
The tibble output from `config_log()` contains the following:
* path to model file (control stream)
* md5 hash of model file
* path to data file
* md5 hash of data file

The user should be able to easily check if any of those paths refer to a file that does _not_ match the md5 hash stored in the log tibble. This would indicate that the model is "out-of-date" and should be refreshed in the appropriate way (i.e. probably re-run).

copernican:
This PR closes #30. `model_has_changed` and `data_has_changed` are added to the tibble returned by `config_log()`, indicating whether the model file and data file, respectively, have changed relative to the MD5 sums in `bbi_config.json`.

The build is failing because the golden file `bbi_config.json` contains the wrong model MD5 sum. In particular, the `model_md5` field reflects `model-examples/1/1.ctl`. This file does not have a trailing newline, whereas the actual model file, `model-examples/1.ctl`, does. Thus, the MD5 sums do not match.

The root cause of the mismatch is likely metrumresearchgroup/babylon#168, which has since been resolved. The test model needs to be re-run for other open issues, which will produce a `bbi_config.json` with the correct MD5 sum. Once complete, we can pull in the updated golden files, bump the version on this branch, and the associated build should pass.


copernican:
@seth127 the build now succeeds, and this PR is ready for your review.

R/config-log.R
```diff
@@ -86,29 +100,104 @@ config_log_impl <- function(.mods) {
...
+#'
+#'   * whether the model file has changed
+#'
+#'   * whether the data file has changed
+#'
+#'   The return value is `NULL` if any element of `fields` is not found in
+#'   `path`.
+#'
+#' @keywords internal
+parse_bbi_config <- function(path,
```
seth127, line 129: Consider renaming `config_log_entry()` just for consistency with this https://github.com/metrumresearchgroup/rbabylon/blob/develop/R/run-log.R#L136 which is pretty much the same pattern.

R/config-log.R
```diff
@@ -86,29 +100,104 @@ config_log_impl <- function(.mods) {
...
+        "`{paste(fields[!(fields %in% names(config))], collapse = ', ')}`",
+        "and will be skipped.",
+        .sep = " "
+      ),
+      glue::glue(
+        "This is likely because it was run with an old version of babylon.",
+        "Model was run on version {config[['bbi_version']]}",
+        .sep = " "
+      ),
+      glue::glue(
```
seth127, line 151: does this need to be glue? I don't see any variables in the string. I think could just be `paste`

tests/testthat/setup-workflow-ref.R
```diff
@@ -170,3 +170,15 @@ cleanup <- function() {
   suppressSpecificWarning(rm(log_df, pos = parent.frame()),.regexpr = "object.+not found")
 }
 
+#' Temporarily perturb a file
+#'
+#' Appends a line to the end of a text file, and flexibly revert the change.
```
seth127, line 175: this says `flexibly revert`. I'm not sure what that means, but I read it as `optionally revert` which doesn't seem to be true. If I'm misunderstanding just ignore.

tests/testthat/setup-workflow-ref.R
```diff
@@ -170,3 +170,15 @@ cleanup <- function() {
...
+#'
+#' @param path string giving the file path
+#' @inheritParams withr::defer
+perturb_file <- function(path, envir = parent.frame()) {
```
seth127, line 179: this function is sexy

tests/testthat/test-config-log.R
```diff
@@ -1,6 +1,22 @@
...
 
-check_config_ref <- function(log_df, run_nums, col_count) {
+# to minimize changes to the existing tests, we define the model and data status
+# for each of the models any particular test might need
+run_status <- dplyr::tribble(
+  ~rel_path, ~model_has_changed, ~data_has_changed,
+        "1",              FALSE,             FALSE,
+        "2",               TRUE,             FALSE,
+        "3",              FALSE,             FALSE,
+ "level2/1",               TRUE,                NA
```
seth127, line 10: interesting... I hadn't considered this case when the data path (or model path) can't be found. It makes sense to return `NA` in that case, but I'm not actually sure that's what we want. Mainly for usability: if a user does `config_log %>% filter(!model_has_changed | !data_has_changed)` will that return the `NA` rows too? I think no, right? Maybe I'm wrong about that.

My gut is that we would want a query like this to return all the rows that had issues and missing data file is definitely an issue. Interested in your thoughts, but I think I would prefer this to return `TRUE`, or maybe more to the point, that the `file_matches` function returns `FALSE` when the file doesn't exist.

tests/testthat/test-config-log.R
```diff
@@ -1,6 +1,22 @@
 context("Constructing config log from bbi_config.json")
 
-check_config_ref <- function(log_df, run_nums, col_count) {
+# to minimize changes to the existing tests, we define the model and data status
+# for each of the models any particular test might need
+run_status <- dplyr::tribble(
```
seth127, line 5: I'm not clear on why this is a variable that is passed into the function each time. It feels to me like a constant that should be in `setup-workflow-ref.R`. Am I missing something? Do you disagree?

seth127 (changes requested):
Looks good overall. I few small comments, the most important is probably about whether to return `NA` or `FALSE` from the `file_matches()` helper if the file in question doesn't exist.

R/config-log.R
copernican: I agree it could be `paste()`. I used `glue()` for consistency (the previous two "chunks" do use it).

tests/testthat/setup-workflow-ref.R
copernican: By "flexibly," I mean that the caller can control _when_ the change is reverted through the `envir` argument. I can spell that out if you prefer, e.g., "and revert the change in an environment specified by the caller."

tests/testthat/setup-workflow-ref.R
copernican: Thanks, my first attempt at implementing a ["local" helper](https://testthat.r-lib.org/articles/test-fixtures.html#-local-helpers).

tests/testthat/test-config-log.R
copernican: It's passed as an argument for referential transparency. We only call this function when we run the tests, so this should be fairly lightweight.

`run_status` is defined here rather than in `setup-workflow-ref.R` because 

- It's not used by any other test file.

- The further a value is defined from where it will be used, the harder it is to look it up. The values in this object are a direct consequence of the `setup()` call (copy this model here, copy that model there), so I wanted to keep the "expectation" close to its definition.

If you feel strongly, I can adopt the pattern you suggest.

tests/testthat/test-config-log.R
copernican: The result of `x | y` is `NA` in the case that at least one of `x` and `y` is `NA`, and the other value is not `TRUE`. As you suggest, `filter()` does not return rows where the condition is `NA`, so a call like you describe would omit rows where `model_has_changed` or `data_has_changed` indicate an issue. In the example below, we probably want to retain rows 3, 7, and 9 from `df_or`.

```r
library(dplyr, warn.conflicts = FALSE)
df <- dplyr::tibble(
  x = rep(c(TRUE, FALSE, NA), each = 3L),
  y = rep(c(TRUE, FALSE, NA), times = 3L)
)
df_or <- dplyr::mutate(
  df, 
  x_or_y = x | y,
  not_x_or_not_y = !x | !y
)
print(df_or)
#> # A tibble: 9 x 4
#>   x     y     x_or_y not_x_or_not_y
#>   <lgl> <lgl> <lgl>  <lgl>         
#> 1 TRUE  TRUE  TRUE   FALSE         
#> 2 TRUE  FALSE TRUE   TRUE          
#> 3 TRUE  NA    TRUE   NA            
#> 4 FALSE TRUE  TRUE   TRUE          
#> 5 FALSE FALSE FALSE  TRUE          
#> 6 FALSE NA    NA     TRUE          
#> 7 NA    TRUE  TRUE   NA            
#> 8 NA    FALSE NA     TRUE          
#> 9 NA    NA    NA     NA
dplyr::filter(df_or, not_x_or_not_y)
#> # A tibble: 5 x 4
#>   x     y     x_or_y not_x_or_not_y
#>   <lgl> <lgl> <lgl>  <lgl>         
#> 1 TRUE  FALSE TRUE   TRUE          
#> 2 FALSE TRUE  TRUE   TRUE          
#> 3 FALSE FALSE FALSE  TRUE          
#> 4 FALSE NA    NA     TRUE          
#> 5 NA    FALSE NA     TRUE
```

Depending on how "serious" of a problem we think having missing files is, we could

1. have `file_matches()` return `FALSE` when the file doesn't exist

2. have `file_matches()` throw an error or raise a warning

Do you prefer one of these, or a third option?

tests/testthat/setup-workflow-ref.R
seth127: Yea that's more clear. Thanks.

tests/testthat/test-config-log.R
seth127: I think FALSE is the right move. It makes sense as is "Q: does this match? A: no there's nothing there." 

And as far as the column goes, "Q: has the data changed? A: yes, the file is gone." 

I don't think and error or warning is necessary, but like I said, I think people will try to filter on this column to see if everything is as expected and the NA rows would likely slip through the cracks.

tests/testthat/test-config-log.R
seth127: No I'm fine with that reasoning.

seth127 (approved):
LGTM

PR #152, opened 2020-09-10
Fix: revert batch param estimates

seth127:
Implementing https://github.com/metrumresearchgroup/rbabylon/issues/149 to get rid of `bbi_summary_log_df` and `bbi_summary_list` dispatches for the `param_estimates()` function.


copernican (approved):
LGTM

seth127:
removed from milestone because not needed for validation

PR #154, opened 2020-09-10
Fix/rerun tests cleanup

seth127:
Cleanup up version of https://github.com/metrumresearchgroup/rbabylon/pull/145 before things got tangled. Started from https://github.com/metrumresearchgroup/rbabylon/commit/f14bb175ae6a8f8e2f074bfb775641632d129754


tests/testthat/model-examples/1/1.cpu
```diff
@@ -1 +1 @@
-       2.543
+       2.696
```
copernican, line 1: Do we still need this file? I don't see any references to it.

DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Type: Package
 Package: rbabylon
 Title: R package for babylon
-Version: 0.7.0.8006
+Version: 0.7.0.8007
```
copernican, line 4: We may also need to bump the version constraint for babylon, which appears in

- `.drone.jsonnet`

- `DESCRIPTION`

- `test-rbabylon.R`

- `zzz.R`

Given that this PR updates the model output using babylon 2.3.0, it feels appropriate that it also change the babylon version throughout.

copernican (changes requested):
Just a couple of questions.

tests/testthat/model-examples/1/1.cpu
seth127: Yes, bbi summary parses it. If it's missing it will throw an error.

DESCRIPTION
seth127: Yea I thought I did it in the DESCRIPTION but I guess that was on the other branch that we closed. 

Do we really have to manually bump this in 4 places? That's definitely not ideal. If there's an easy way around that we should do it. If not, we should think about doing it in a future PR somewhere. 

I'll bump them and commit though.

seth127: Ok I bumped it in drone, DESRIPTION, and zzz. The test-rbabylon is actually a test of whether we can parse semantic versioning so that does _not_ need to be changed.

Sidenote: we should streamline this at some point. And semi-related, can we delete the `.drone.yml`? We don't use it do we? And it's out of date.

copernican: 1. Yes, we can remove `.drone.yml`. Happy for you to do that in this PR, if you like.

2. The version constraint should definitely be in `DESCRIPTION`. We can [parse that file](https://github.com/metrumresearchgroup/rbabylon/pull/95/files#r443783425) in `zzz.R`, assuming we want to keep that pattern, or directly in `check_bbi_version_constraint()`.

3. Good point about the version reference in `test-rbabylon.R`. Incidentally, because `check_bbi_version_constraint()` is an internal function, I'd be okay removing that test altogether (in keeping with the philosophy of only testing the "API").

4. The version reference in `.drone.jsonnet` exists so that Drone knows which version to install. We could call `use_bbi()` in Drone, but we'd still want to specify the version number to get predictable builds. In other words, we want to be intentional about changing the version of babylon used in CI rather than always grabbing the latest release. So, I think it makes sense to have a version _constraint_ in `DESCRIPTION`, and a _specific_ version in `.drone.jsonnet`.

seth127: 1. doing it now...
2. I like this idea but let's punt on it for now. I tried this once and there was some subtlety that I can't remember.
3. probably fine to remove that test, but I don't feel like thinking about it now so let's punt.
4. that makes sense. My guy says we should just take `latest` but if you disagree I'll defer on that.

copernican (approved):
LGTM. We are going to accept the risk of merging this to `develop` despite the build failing. This is in part due to breaking changes in the latest release of babylon. We expect these to be resolved by #129 and #137, at which point `develop` should build successfully.

Issue #119, opened 2020-08-10
Request: can shrinkage be included in the tibble output by param_estimates()


KatherineKayMRG:
Hi @seth127, 

We discussed this earlier - it would be very helpful if the `param_estimates()` function could also return the shrinkage estimates. Currently, for every parameter table I make, I then have to grab the shrinkage values from the full summary table and merge it into param_estimates tibble. It's not difficult but it is repetitive and it would be nice to just have it returned automatically. 

Thanks,
Katherine

seth127:
@KatherineKayMRG would you mind pasting a sample of code that you use to do this? Just to make sure that we're pulling the same numbers, etc.

Thanks

KatherineKayMRG:
Hey @seth127 - I use in the example project I use

```
shrinkage = sum$shrinkage_details[[1]][[1]]$eta_sd
shrinkDF <- tribble(
  ~name, ~shrink, 
  "OMEGA11", shrinkage[1], 
  "OMEGA22", shrinkage[2], 
  "OMEGA33", shrinkage[3], 
)
```
but this would have to adapt to pull the shrinkage from the last estimation method if there were multiple. 

Also, we wouldn't do this for a BAYES run

seth127:
@KatherineKayMRG excuse the basic questions but I want to make sure we have the spec all buttoned up for this. A few questions:

## 1 - which field?

I see the following fields under `shrinkage_details` in the summary object
```
> s <- MOD1 %>% model_summary()
> str(s$shrinkage_details)
List of 1
 $ :List of 1
  ..$ :List of 11
  .. ..$ sub_pop     : int 1
  .. ..$ eta_bar     : num [1:2] 0.0018 -0.00739
  .. ..$ ebv_bar_se  : num [1:2] 0.04 0.0599
  .. ..$ pval        : num [1:2] 0.964 0.902
  .. ..$ eta_sd      : num [1:2] 17.51 2.06
  .. ..$ eps_sd      : num 4.09
  .. ..$ ebv_sd      : num [1:2] 18.52 2.54
  .. ..$ num_subjects: int [1:2] 40 40
  .. ..$ eta_vr      : num [1:2] 31.96 4.08
  .. ..$ eps_vr      : num 8.01
  .. ..$ ebv_vr      : num [1:2] 33.62 5.02
```
above you indicate that, for this column that we are creating, we are only concerned with `eta_sd`. Is that correct?

## 2 - which params are relevant?

[...]

PR #157, opened 2020-09-14
param_estimates() should error on Bayesian methods

seth127:
In working on #119 it was decided that the `param_estimates()` function does not return usable results for Bayesian estimation methods. Because the returned results could actually be wrong and misleading, it was decided (in [this thread](https://github.com/metrumresearchgroup/rbabylon/issues/119#issuecomment-690290391)) that the function should error with some "not implemented" message for the time being.

There is a plan to work on this in the next few months so that `param_estimates()` gives a correct result for Bayesian methods too. @timwaterhouse may be involved in that effort.

Implementation note: all the relevant methods will have `"Bayesian"` in the string contained in `bbi_nonmem_summary$run_details$estimation_method` so we can use that to determine if we should error.

### DIFF
--- a/R/param-estimates.R
+++ b/R/param-estimates.R
@@ -8,6 +8,12 @@
 #' Returns a tibble containing parameter estimates from a `bbi_{.model_type}_summary` object.
 #' Details about the tibble that is returned are in the return value section below.
 #'
+#' @details
+#' Note that **Bayesian methods are not yet supported** by this function. Creating a parameter table
+#' like this for Bayesian estimation methods requires a different approach, which has not yet been
+#' implemented. If the final estimation method of the input model is Bayesian, a `not implemented`
+#' error will be thrown.
+#'
 #' @return
 #' Returns a tibble with the following columns:
 #'
@@ -31,10 +37,17 @@ param_estimates <- function(.summary) {
 #' @importFrom tibble tibble as_tibble
 #' @importFrom purrr map_at map_depth map_lgl
 #' @importFrom rlang list2
+#' @importFrom stringr str_detect
 #' @export
 param_estimates.bbi_nonmem_summary <- function(.summary) {
   num_methods <- length(.summary[[SUMMARY_PARAM_DATA]])
   param_names <- .summary[[SUMMARY_PARAM_NAMES]]
+
+  # if Bayesian method (includes NUTS) do not return df because it is incorrect and misleading
+  est_method <- .summary[[SUMMARY_DETAILS]][[SUMMARY_EST_METHOD]][[num_methods]]
+  if (str_detect(est_method, "Bayesian")) {
+    stop(glue("param_estimates() is not currently implemented for Bayesian methods. Passed has final estimation method: {est_method}"), call. = FALSE)
+  }
 
   summary_vars <- with(
     .summary[[SUMMARY_PARAM_DATA]][[num_methods]],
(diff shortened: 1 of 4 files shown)


R/param-estimates.R
```diff
@@ -8,6 +8,12 @@
 #' Returns a tibble containing parameter estimates from a `bbi_{.model_type}_summary` object.
 #' Details about the tibble that is returned are in the return value section below.
 #'
+#' @details
+#' Note that **Bayesian methods are not yet supported** by this function. Creating a parameter table
+#' like this for Bayesian estimation methods requires a different approach, which has not yet been
+#' implemented. If the final estimation method of the input model is Bayesian, a `not implemented`
```
copernican, line 14: This is a place to consider [custom conditions](https://adv-r.hadley.nz/conditions.html#custom-conditions).

tests/testthat/test-param-estimates.R
```diff
@@ -6,10 +6,6 @@ if (Sys.getenv("METWORX_VERSION") == "" && Sys.getenv("DRONE") != "true") {
 
 setup({
   cleanup()
-  invisible(copy_model_from(yaml_ext(MOD1_PATH), NEW_MOD2, "model from test-param-estimates.R", .directory = "."))
```
copernican, line 9: Is removing these lines related to the requirement here? It looks like general cleanup.

R/param-estimates.R
```diff
@@ -31,11 +37,18 @@ param_estimates <- function(.summary) {
...
+#' @importFrom stringr str_detect
 #' @export
 param_estimates.bbi_nonmem_summary <- function(.summary) {
   num_methods <- length(.summary[[SUMMARY_PARAM_DATA]])
   param_names <- .summary[[SUMMARY_PARAM_NAMES]]
 
+  # if Bayesian method (includes NUTS) do not return df because it is incorrect and misleading
+  est_method <- .summary[[SUMMARY_DETAILS]][[SUMMARY_EST_METHOD]][[num_methods]]
+  if (str_detect(est_method, "Bayesian")) {
+    stop(glue("param_estimates() is not currently implemented for Bayesian methods. Passed has final estimation method: {est_method}"), call. = FALSE)
```
copernican, line 49: > Passed has final estimation method

What is meant by `Passed`? The summary object? Perhaps we'd consider

> \`.summary\` has final estimation method : {est_method}

copernican (approved):
LGTM, just a few comments to consider.

tests/testthat/test-param-estimates.R
seth127: No I just noticed they were hanging around and should've been deleted in https://github.com/metrumresearchgroup/rbabylon/pull/152 because they're no longer used.

R/param-estimates.R
seth127: Yes, this is a good point and something I've considered in the past. Devin pointed me towards reading up on `rlang::abort()` awhile ago with this in mind.

I would love to address this at some point, but given our current time constraints, I think I'm gonna punt. Good to keep reminding ourselves of this though. I love errors with classes. (I miss thee Python...)

R/param-estimates.R
seth127: Oh I think I forgot a noun in there. Good catch. I think I'll just address this in https://github.com/metrumresearchgroup/rbabylon/pull/136 to avoid making a commit with only this change.

PR #161, opened 2020-09-16
fix: summary hang on 510

dpastoor:
```
R -e "options('rbabylon.bbi_exe_path' = 'bbi'); devtools::check()"
```

```

── R CMD check results ──────────────────────────────── rbabylon 0.7.0.8009 ────
Duration: 51.2s

❯ checking installed package size ... NOTE
    installed size is  7.5Mb
    sub-directories of 1Mb or more:
      param_examples   6.7Mb


0 errors ✔ | 0 warnings ✔ | 1 notes ✖
>
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -44,7 +44,8 @@ Suggests:
     knitr,
     rmarkdown,
     testthat (>= 2.1.0),
-    withr
+    withr,
+    tidyselect
 VignetteBuilder: 
     knitr
 Encoding: UTF-8
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -44,8 +44,7 @@ Suggests:
     knitr,
     rmarkdown,
     testthat (>= 2.1.0),
-    withr,
-    tidyselect
+    withr
 VignetteBuilder: 
     knitr
 Encoding: UTF-8
--- a/R/rbabylon.R
+++ b/R/rbabylon.R
@@ -44,10 +44,33 @@ bbi_exec <- function(.cmd_args,
     stdout = "|",
     stderr = "2>&1"
   )
-
   if (.wait) {
     # wait for process and capture stdout and stderr
-    p$wait()
+    if (all(c("nonmem", "summary", "--json") %in% .cmd_args)) {
+      # this was originally just a wait, however this caused an infinite
+      # hang with model 510 summary in the tests as of 0.7.0.8009 on mac
+      # it _could_ be due to processes finishing so fast that ps/processx
+      # are not able to adequately track the status of the subprocess while
+      # waiting.
+      # This was attempted to be artifically delayed by making the ext 50k lines
+      # but this still resulted in the hang so not positive. Have not seen
+      # this error in other models.
+      # To get around this false hang we want to timeout.
+      # This should be ok as the json should
+      # be back at that point, in which p$read_all_output_lines will return
+      # valid values.
+      #
+      # in addition to give a debug capability, setting the timeout to
+      # an option default of 3. generally summary should return in 10 ms
+      # or so, however if we move towards providing batch summaries
+      # this process could take longer. This will give us an escape hatch
+      # in such a case that people report errors where summary takes longer
+      # then we can establish a longer potential timeout or more heuristics
+      # around what time might be (eg if see its a summary on a range of models)
+      p$wait(timeout = getOption("rbabylon.summary_wait_timeout", 3))
+    } else {
+      p$wait()
+    }
     # check output status code
     output <- p$read_all_output_lines()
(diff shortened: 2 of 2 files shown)


dpastoor:
FYI this is actually passing after kicking drone to re-run again manually

image

R/rbabylon.R
```diff
@@ -44,18 +44,40 @@ bbi_exec <- function(.cmd_args,
     stdout = "|",
     stderr = "2>&1"
   )
-
   if (.wait) {
     # wait for process and capture stdout and stderr
-    p$wait()
+    if (all(c("nonmem", "summary", "--json") %in% .cmd_args)) {
```
copernican, line 49: It appears we can solve this without special handling for this command, and without introducing a timeout. The following change appears to work on both macOS and Linux:

```r
  if (.wait) {
    if (Sys.info()[["sysname"]] != "Darwin") {
      p$wait()
    }
    output <- p$read_all_output_lines()
    check_status_code(p$get_exit_status(), output, .cmd_args)
  } else {
    output <- "NO STDOUT BECAUSE `.wait = FALSE`"
  }
```

36eab35 notes that `read_all_output_lines()` is non-blocking, which would explain why `wait()` is necessary on Linux. But `wait()` actually causes problems on macOS, and `read_all_output_lines()` is sufficient. I'm not sure if that means that `read_all_output_lines()` _is_ blocking on macOS, but the behavior is definitely different.

DESCRIPTION
```diff
@@ -44,7 +44,8 @@ Suggests:
     knitr,
     rmarkdown,
     testthat (>= 2.1.0),
-    withr
+    withr,
+    tidyselect
```
copernican, line 48: Let's omit this change.

copernican (changes requested):
This issue is present in `develop`, so this PR should be into `develop` (we can subsequently rebase `fix/summary-log-vignette`). Also, there appears to be a solution that does not require special handling for `bbi nonmem summary`.

seth127:
Looks like we need to rebase this onto `develop` if that's still the plan.

This looks good to me in theory. I'll pull it down and run `devtools::check()` on my local Mac to see if it builds. I think I like this solution better than the timeout. Although, if I'm honest, none of this gives me a warm fuzzy feeling. Might be worth submitting a processx bug report.

Still, ultimately I think the chances of this negatively effecting a user are very low so if this makes it build successfully then I'm satisfied for now.

DESCRIPTION
dpastoor: gotta have this change _if_ we are using tidyselect in vignette. If we're not using it then yes this should be dropped

R/rbabylon.R
copernican: We agreed to

1. implement the solution in this PR rather than special handling for macOS

2. dig into the apparent `processx` issue to understand the root cause

DESCRIPTION
seth127: we certainly don't _need_ to use `tidyselect` in the vignette but it gives a nice pattern that I think scientists will like. If we don't think there's much issue with adding this dependency then let's go for it. 

For the record though, the `tidyselect` function in the vignette isn't even evaluated. It's in an `eval=F` block. Not sure if that makes a difference but it means that we have a dependency that we don't actually _use_ anywhere.

seth127: Upon [further research](https://stackoverflow.com/a/9900160/3743213), seems like putting it in `Suggests` is the right move, so let's just do this.

R/rbabylon.R
copernican: We are seeing a variety of errors running `check()` on Linux. It appears that increasing the `timeout`, e.g., to 20, resolves the issue, but the sensitivity to this value on different platforms is unexpected.

copernican:
`check()` also succeeds on macOS.

DESCRIPTION
copernican: Agree that `Suggests` is a good place. Let's include that change in #158.

R/rbabylon.R
seth127: We settled on 1000 ms as the safest route, though it is obviously very very generous. The thinking here is that, in the very unlikely case that a user stumbles onto this, it will still appear to hang and they will submit a bug report. If we set it to 30ms or something it a) could still fail to build and b) could come back just fast enough that a user thinks it's normal. 

For example, running 1000 model summaries for a bootstrap or something might take 2 mins when it should really take 3 seconds if things were working right. But we would never hear about it because the user might think 2 mins was slow but reasonable and _not_ submit a bug or anything.

To reiterate though: this is very much a band-aid solution. We are planning to address this more thoroughly in a future release.

seth127 (approved):

copernican (approved):

PR #195, opened 2020-09-25
Refactor/path getters

### LINKED ISSUE #190: refactor get_*_path() functions
# Summary

The helper functions `get_model_path()`, `get_output_dir()`, and `get_yaml_path()` (collectively, the `get_*_path()` functions) rely on `get_path_from_object()` to dispatch on various objects, currently `bbi_nonmem_model` and `bbi_run_log_df`. Suppose that the user wants to get the model paths for all models in a `bbi_run_log_df` with a call like `get_model_path(rlg)`. The call stack will be

```
get_model_path(rlg)
> get_path_from_object(rlg)
>> get_path_from_object.bbi_run_log_df(rlg)  # map over the models
>>> get_path_from_object.character(path_to_model)
>>>> get_path_from_object.default(mod)
```

We propose changing the pattern as follows:

```
get_model_path(rlg)
> get_model_path.bbi_run_log_df(rlg)
>> get_path_from_log_df(rlg)  # map over the models
>>> get_model_path(mod)
>>>> get_model_path.bbi_nonmem_model(mod)  # could be Stan models, in the future
```

The main benefit of this pattern is that we will change the model structure in #182, which will break the current `get_*_path()` functions, which are required to understand the model structure. If we instead dispatch directly on the model, the details of the structure of a particular object will be encapsulated by its method.

# Technical specification

There is currently a `default` method for `get_path_from_object()`. We intend to remove this method, and do not intend to replace it with similar methods for the `get_*_path()` functions. This will break existing code where `get_*_path()` is called with an object that does not inherit from `bbi_nonmem_model` or `bbi_log_df`. For example, `new_model()` calls `save_model_yaml()`, which in turn calls `get_yaml_path()`. The trouble is that the object passed to `save_model_yaml()` does not inherit from a class for which `get_yaml_path()` has a method.

The core issue is that `create_model_object()` requires that its argument have an element named `yaml_md5`, but this field cannot exist if the YAML file does not exist, which will only be the case after calling `save_model_yaml()`. We can resolve this difficulty with the following pattern:

1. `create_model_object()` assigns the appropriate class, then calls `save_model_yaml()` with the resulting object. This object has a class for which there exists a method, so `get_yaml_path()` will succeed.

2. `save_model_yaml()` writes out the YAML file, and adds or updates the `yaml_md5` element of the model object passed to it.

3. `save_model_yaml()` returns the updated model object.

4. `create_model_object()` checks that the object it received from `save_model_yaml()` has all required elements, including `yaml_md5`, and returns the new model object.

A similar issue affects `copy_nonmem_model_from()`, which calls `get_model_path()` with an object that does not have a corresponding method. The return value of this call is passed to `copy_control_stream()` as the destination for the new model file. But this may not be necessary. `copy_nonmem_model_from()` is called only by `copy_model_from.bbi_nonmem_model()`, which passes the argument `.new_model`, which in turn is the absolute path to the new model `absolute_model_path`. We should be able to set the file extension based on the extension of the parent model file, so that we can determine the path to the new model file without calling `get_model_path()`.

```r
parent_ext <- fs::path_ext(.parent_model_path)
.new_model_path <- fs::path_ext_set(.new_model, parent_ext)
```

1. Add new exported function `get_model_path.bbi_nonmem_model()` that returns the path to the model file for a `bbi_nonmem_model` object. We expect that the specific logic to extract the model path will change as a consequence of #141 and #182, but we prefer to make this change separately.

2. Repeat (1) for `get_output_dir()` and `get_yaml_path()`, i.e., create methods for `bbi_nonmem_model` objects.

3. Add new internal function `get_path_from_log_df()` that reads in a list of models from a `*_log_df` object and returns a character vector of paths using an appropriate helper function, e.g., for the model path.

4. Add new exported function `get_model_path.bbi_run_log_df()` that returns a character vector of model paths for models in a `bbi_run_log_df` object. Internally, this function is expected to call `get_path_from_log_df()`.

5. Repeat (4) for `get_output_dir()` and `get_yaml_path()`.

6. Remove `get_path_from_object()` generic and methods. While exported, users were discouraged from calling these functions, and they are no longer needed.

# Test changes

1. Remove tests for `get_path_from_object()` and its methods.

2. Add tests for new `get_*_path()` methods. The existing tests for (what will become) the generic test a `bbi_nonmem_model` and can be repurposed.

seth127:
closes #190


R/modify-model-field.R
```diff
@@ -39,10 +39,7 @@ modify_model_field <- function(.mod, .field, .value, .append = TRUE, .unique = T
   }
 
   # overwrite the yaml on disk with modified model
-  save_model_yaml(.mod)
-
-  # refresh md5 hash in model object
-  .mod[[YAML_YAML_MD5]] <- digest(file = get_yaml_path(.mod), algo = "md5")
```
seth127, line 45: this was a good move

copernican (approved):
LGTM

PR #197, opened 2020-09-28
Fix/remove mod path from yaml

### LINKED ISSUE #196: do not save model path to YAML
# Summary

`save_model_yaml()` writes a model to a YAML file. It currently excludes certain elements of the model object, including the name of the YAML file (this is sensible: a file doesn't need to store its own name). #141 will unify the location of the model file and YAML file, hence the path to the model will not need to be stored in the YAML. We should add the model path element to the list of elements not to write to the YAML file.

# Technical specification

1. Add `YAML_MOD_PATH` to `YAML_ERASE_OUT_KEYS`, defined in `aaa.R`.

# Test changes

1. Update tests that fail due to excluding this field (whether `model_path` is present in the YAML will change the MD5 sum of the file).

seth127:
closes #196

### DIFF
--- a/R/aaa.R
+++ b/R/aaa.R
@@ -105,8 +105,9 @@ RUN_LOG_REQ_COLS <- c(
 YAML_ERASE_OUT_KEYS <- c(
   WORKING_DIR,
   YAML_YAML_NAME,
-  YAML_YAML_MD5,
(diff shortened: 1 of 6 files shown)


copernican (approved):
LGTM

PR #206, opened 2020-09-30
refactor: import %||% from rlang, closes #204

### LINKED ISSUE #204: import %||% from rlang
# Summary

The binary operator `%||%` returns its second argument if the first is `NULL`, else returns the first argument. Rather than defining it in `rbabylon`, we should import the equivalent operator from `rlang`.

# Technical specification

1. Remove the definition of `%||%` in `utils.R`.

2. Add an `@importFrom` statement to the roxygen comments of a function that calls it, e.g., `create_model_object()`.

# Test changes

There are no test changes.

copernican:
This PR closes #204.


seth127 (approved):
strong work here

Issue #182, opened 2020-09-23
Implementation: combine YAML_YAML_PATH, YAML_MOD_PATH, YAML_OUT_DIR


seth127:
This issue came out of a comment here: https://github.com/metrumresearchgroup/rbabylon/issues/179#issuecomment-697845924

Essentially, because of #141, we will be enforcing that `.mod[[YAML_YAML_PATH]]`, `.mod[[YAML_MOD_PATH]]`, `.mod[[YAML_OUT_DIR]]` are all the same path but with different extensions. This makes it both silly and dangerous to carry all three of them around. Silly for obvious reasons; dangerous because a user could do something like manually change `.mod[[YAML_MOD_PATH]]` and move the control stream, thus breaking our ability to enforce #141 (and inadvertently breaking their model object).

The purpose of this issue will be to reconsider how that is stored in the model object. Some things to consider:

* carry around only `ABS_MOD_PATH` as used in the `_log` functions.
* change all the `get_path_from_object()` functions to add some extension to `.mod[[ABS_MOD_PATH]]` instead of querying some other key.
* do we even need `.mod[[WORKING_DIR]]` if we do this?
  * alternately, should we keep `.mod[[WORKING_DIR]]` and also keep `mod[[MODEL_FILE]]` such that the control stream path would unambigously be `file.path(.mod[[WORKING_DIR]], mod[[MODEL_FILE]])` and the other paths could all be inferred from that?

Need to discuss more... **Should consider doing this before, or in tandem with, #141.**

copernican:
There may be a benefit to storing at least (only?) the absolute path to the model file in the model object. We anticipate that that file will be edited by the user, and 

- do not anticipate that the user will (directly) edit the YAML file
- will not support the user changing the location of the output directory to be inconsistent with the definition in #141

While storing the path to the model file in some sense creates a "leak" in the abstraction of #141, given the outsize importance of the model file, it may be appropriate. If we adopt this, then it should be straightforward to resolve the other meanings of `absolute_model_path`:

- the working directory is obtained by dropping the model file extension
- the YAML file is obtained by replacing the model file extension with `.yaml`

We might imagine helper functions like 

```r
model_file_to_output_dir <- function(path) {
  dirname(path)
}

model_file_to_yaml <- function(path) {
  # replace model file extension with `.yaml`
}
```

[...]

Issue #174, opened 2020-09-23
deprecate get_model_directory() and set_model_directory()


copernican:
`get_model_directory()` and `set_model_directory()` get or set, respectively, the value of the option `rbabylon.model_directory`. We will deprecate this option, so these functions should be removed.

PR #207, opened 2020-09-30
refactor submit_nonmem_models()

### LINKED ISSUE #205: refactor submit_nonmem_models() and build_bbi_param_list()
# Summary

`submit_nonmem_models()` refers to elements of the `bbi_nonmem_model` object that will be deprecated, e.g., `YAML_MOD_PATH`. It would be straightforward to fix these references with the `get_*_path()` functions (refactored as described in #190), except that these references are not made with respect to a model object, so there is no corresponding `get_*_path()` method. Rather, these references are to elements of a list returned by `build_bbi_param_list()`, and are currently a convenient shortcut.

`build_bbi_param_list()` exists because we prefer to make the minimal number of calls to bbi when submitting multiple models. In particular, we prefer that bbi, not `rbabylon`, handle the parallelization of model submission. It is possible to submit multiple models with a single bbi call, provided that the models all share the same set of CLI arguments to bbi and a working directory (the parent directory of `absolute_model_path` as defined in #141). Thus, we need to identify which of a set of models can be submitted together. `build_bbi_param_list()` performs this grouping operation.

But, `build_bbi_param_list()` also refers to elements of the `bbi_nonmem_model` object that will be deprecated. One solution is to modify `build_bbi_param_list()` to be (mostly) agnostic as to the structure of the model object, and at the very least not to refer to elements that will be deprecated. This can be done by returning the model objects passed as `.mods` in the output, but grouped to enable the submission of multiple models as described above. Then, `submit_nonmem_models()` can pass those model objects to the appropriate helper functions, eliminating its direct dependence on the internal structure of the model object.

This issues describes how such a refactor could be done.

# Technical specification

## build_bbi_param_list()

1. Construct a list whose elements correspond to the model objects and hold the "cleaned" set of arguments for the appropriate model, where "cleaned" means that 

    a. conflicts between runtime arguments and arguments in the model YAML file have been resolved

    b. arguments have been checked and formatted in the sense of `check_bbi_args()`

    c. arguments are sorted

2. Construct a list whose elements correspond to the model objects and hold the working directory for the appropriate model.

3. Construct a list of model "keys" whose elements correspond to the model objects and hold the combination of model arguments and working directories. Models that share a key can be submitted in a single bbi call. This list can be created from the respective elements of the lists constructed in (1) and (2).

4. Map over the list of unique keys and construct a list with two elements:

    a. `bbi_args` holds the CLI arguments corresponding to the key

    b. `models` is a list of the model objects that share the key

    Thus, the return value of `build_bbi_param_list()` a list whose elements correspond to the set of unique model keys, where each element is itself a list constructed in the manner described.

    ```
    - key_1
      - bbi_args: bbi_args_1
      - models
        - mod_A
        - mod_C
    - key_2
      - bbi_args: bbi_args_2
      - models
        - mod_B
    ```

## submit_nonmem_models()

1. Modify the assignment to `cmd_args` to be

    ```r
    cmd_args <- c(
      "nonmem",
      "run",
      .mode,
      purrr::map_chr(.run[["models"]], get_model_path),
      .run[["bbi_args"]]
    )
    ```

2. Modify the assignment to `model_dir` to be 

    ```r
    model_dir <- .run[["models"]][[1L]][[WORKING_DIR]]
    ```

   Note that `WORKING_DIR` will be deprecated by #182, and is expected to be replaced by a call like `get_model_working_directory(.run[["models"]][[1L]])`. We might also choose to extract the anonymous function passed to `map()` when building `cmd_args_list`.

# Test changes

1. The tests for `submit_models(dryrun = TRUE)` fail because they expect the bbi call to be of the form

    ```
    bbi nonmem run sge 1.ctl --flags
    ```

    Because `get_model_path()` returns an absolute path, the implementation above results in a call like

    ```
    bbi nonmem run sge /path/to/1.ctl --flags
    ```

    But, the integration tests that actually call bbi pass, which suggests that we might reasonably update the expectation. One complication is that the absolute path may not be reproducible, e.g., include a user's home directory, but that should be obtainable by resolving the path within the test.

2. The tests of `build_bbi_param_list()` should be removed, because it is an internal function.

copernican:
This PR closes #205.


tests/testthat/test-submit-models.R
```diff
@@ -53,14 +59,45 @@ withr::with_options(list(rbabylon.bbi_exe_path = "bbi",
...
               expect_identical(
                 proc_list[[2]][[PROC_CALL]],
-                as.character(glue("cd {file.path(getwd(), MODEL_DIR)} ; bbi nonmem run sge 3.ctl --clean_lvl=2 --overwrite --threads=1"))
+                as.character(
+                  glue("cd {model_dir} ; bbi nonmem run sge {mod_ctl_path[3]} --clean_lvl=2 --overwrite --threads=1"))
               )
             })
 
+  test_that("submit_models() works for models in different directories", {
+    new_dir <- "level2"
```
seth127, line 76: Let's add a TODO here (or somewhere in this test) to the effect of "once #174 is done we should revisit using `create_all_models()` from `setup-workflow-ref.R` instead of manually calling `copy_model_from()` all over this test."

R/utils.R
```diff
@@ -106,41 +106,70 @@ format_cmd_args <- function(.args, .collapse = FALSE) {
...
-
-    # append model path
-    param_list[[arg_md5]][[YAML_MOD_PATH]] <- c(param_list[[arg_md5]][[YAML_MOD_PATH]], .mod[[YAML_MOD_PATH]])
-  }
+  # character vector whose elements correspond to `.mods` and are strings
+  # representing the CLI args to bbi for the corresponding model
+  all_mod_args <-
+    .mods %>%
+    # TODO: consider whether this should be abstracted
+    purrr::map(YAML_BBI_ARGS) %>%
```
seth127, line 134: I don't understand why these are all made as a their own `map` calls? Why not this?
```
map(.mods, function(.m) {
  .m[[YAML_BBI_ARGS]] %>%
  ...
})
```

The way it is strikes me as hard to read, but also doesn't this mean it will iterate over this list 4x instead of once?

R/utils.R
```diff
@@ -106,41 +106,70 @@ format_cmd_args <- function(.args, .collapse = FALSE) {
...
+    # `parse_args_list()` will return an empty list if both are `NULL`
+    purrr::map(~ parse_args_list(.bbi_args, .)) %>%
+    purrr::map(check_bbi_args) %>%
+    purrr::map(sort)
+
+  mod_working_dirs <-
+    .mods %>%
+    # TODO: change this once we have the helper
+    purrr::map(WORKING_DIR) %>%
```
seth127, line 143: same question as above, thought the efficiency argument is less relevant with only two. Is this a popular idiom? I don't like.

R/utils.R
```diff
@@ -106,41 +106,70 @@ format_cmd_args <- function(.args, .collapse = FALSE) {
...
+      keep_mod_idx <- purrr::map2_lgl(
+        all_mod_args,
+        mod_working_dirs,
+        ~ identical(.x, args) && identical(.y, key[dir_idx])
+      )
+      list(
+        bbi_args = args,
+        models = .mods[keep_mod_idx]
+      )
+    }
```
seth127, line 169: nit: the formatting/indentation is a little off here.

seth127 (changes requested):
A few small things but looks good overall

R/utils.R
copernican: The goal of writing it with multiple `map()` calls is actually to improve readability and maintainability. This code can be run line by line, and after each line, the resulting list inspected. Each line is simple/does one thing. In addition, because all the `map()` calls are at the same level as the `.mods` object, we are guaranteed that after each transformation, the resulting list will have the same length as `.mods`. Thus, we can easily see the incremental transformations to `.mods`, which means it is easier to convince ourselves that those transformations are correct.

The pattern you describe means that the code cannot be run line by line interactively, because `.m` exists only within the scope of the anonymous function. For similar reasons, the result of each transformation cannot be inspected in turn.

So, yes, the way it is written results in more `map()` calls, but each call does something simple. Writing it with a single `map()` call and a more complicated anonymous function would also work, but is harder to run interactively, hence possibly harder to read and maintain.

R/utils.R
copernican: This is temporary. When we implement #182, this will become

```r
mod_working_dirs <- purrr::map_chr(.mods, get_model_working_directory)
```

R/utils.R
copernican: What specifically is off? I just let the IDE indent this block, and git shows no changes.

R/utils.R
seth127: ok I'll buy that. If we were iterating over a large list this would be a concern, but it's negligible here. Thanks.

R/utils.R
seth127: well, what first confused me is that the two curly brackets that open and close the anonymous function look like they're on different indentation levels, but I guess that's because the `~` is actually the start of that indentation level. It's fine, especially if there are no other changes.

seth127 (approved):

Issue #141, opened 2020-09-04
Enforce model file lives next to yaml file


seth127:
Enforce that the following paths are always equal to each other for any model, and reflect what is stored in the `absolute_model_path` column of the `_log_df` tibbles.

* the path to the output directory
* the path to the YAML file (without extension)
* the path to the model file (without extension)

In practice this will mean that a user _cannot_ name the model file or output directory anything different from each other, or from the YAML file. This is possible with `bbi` CLI, but it will _not_ be allowed in `rbabylon`.

## Acceptance Criteria

* ~~add a check (similar to `check_yaml_in_sync()`) that verifies the above is true at important points.~~ This won't be necessary because of #182 
  * we may want to use this issue as a chance to make sure paths to non-existing files/dirs get caught reasonably.
* deprecate  `.model_path` argument of `new_model()`
* clean up control flow in [lines 34-40 of create_model_object()](https://github.com/metrumresearchgroup/rbabylon/blob/develop/R/classes.R#L34) so that only lines 38-39 remain
* change warning on [line 243 of find_model_file_path](https://github.com/metrumresearchgroup/rbabylon/blob/develop/R/get-path-from-object.R#L243) to an error like `stop(glue("No model file found at {.ctl_path} or {.mod_path}. Please put relevant model file in that location before creating a model object."))`. Consider renaming this `find_model_file_path()` helper to something more in line with `model_file_exists()` pattern.

# Original post

Original post that the first few comments are replying to has been copied here https://github.com/metrumresearchgroup/rbabylon/issues/141#issuecomment-697630676

seth127:
I'm looking for feedback on this, mostly on the *Counterpoint?* section. @copernican @dpastoor @kylebaron @anarpkpd @timwaterhouse @shairozan tag any others who may want to weigh in.

anarpkpd:
I have noticed that when we run BAYES we tend to do so in the folder that contains the final developed model. Bayes is run 4 times ( 4 chains) . Ofteh the analyst give the same run number to the four chains+letter, the final developed model+letter(i.e, 23a,23b,23c,23d). Is  possible to do this with rbabylon? ( a run within a run directory?) and since we are just perturbing the init est, can we use the same yaml file for the four chains?

seth127:
That is a good question @anarpkpd and we have thought about doing something like this. This is a different issue than what we're discussing here though, although it is connected.

[...]

PR #213, opened 2020-10-01
Refactor/abs mod path

### LINKED ISSUE #182: Implementation: combine YAML_YAML_PATH, YAML_MOD_PATH, YAML_OUT_DIR

copernican:
Closes #182, #141.


R/classes.R
```diff
@@ -32,19 +32,9 @@ create_model_object <- function(res, save_yaml) {
...
-    res[[YAML_MOD_PATH]] <- as.character(fs::path_rel(.mod_path, res[[WORKING_DIR]]))
-  }
-
-  # check for correct NONMEM extension
-  if (.model_type == "nonmem" && (!is_valid_nonmem_extension(res[[YAML_MOD_PATH]]))) {
-    stop(glue::glue("model_path defined in yaml at {res[[YAML_MOD_PATH]]} must have either a .ctl or .mod extension, but found {res[[YAML_MOD_PATH]]}"))
-  }
+  # we won't know the model file extension, so we rely on this helper to check
+  # the possible extensions and throw an error if none exists
+  find_nonmem_model_file_path(res[[ABS_MOD_PATH]])
```
seth127, line 37: I'd propose putting `.check_exists = TRUE` in here. I know it's the default, but we could theoretically change the default someday (or get rid of that arg) and this call would start silently doing nothing. Also, it just makes it more explicit that we're calling this specifically to check it the file exists.

seth127 (approved):
These changes are flawless.

PR #217, opened 2020-10-03
Refactor/new model args

### LINKED ISSUE #179: modify path arguments to new_model()
# Summary

#141 requires that the model file, YAML file, and output directory all have the same path (minus file extensions). #182 adds this path to the model object as `absolute_model_path`. In addition, we will deprecate the `.directory` argument as part of a broader refactor. `new_model()` should reflect these changes, which involves

- removing the `.directory` argument
- renaming the `.yaml_path` argument to `.path`

We also take this opportunity to ensure that `.path` can contain periods.

# Technical specification

## new_model()

1. Remove `.directory` argument, and remove call to `combine_directory_path()`.

2. Rename `.yaml_path` argument to `.path`.

3. Remove warning if `.path` (formerly, `.yaml_path`) does not have a `yaml` file extension, because `.path` will not have an extension.

4. Check that the YAML file implied by `.path` does not exist, and throw an error if it does and `overwrite = FALSE`. The error should (continue to) inform the user that this check can be skipped if `.overwrite = TRUE`.

5. Construct the absolute model path as 

    ```r
    file.path(normalizePath(dirname(.path)), basename(.path)
    ```

    to avoid a warning from `normalizePath()`, and assign the result to `.mod[[ABS_MOD_PATH]]`.

6. Change `.start` argument to `safe_based_on()` to be `dirname(.path)`.

## copy_nonmem_model_from()

1. Modify the call to `new_model()` by

    a. changing `.yaml_path` to `.path = .new_model`

    b. removing the `.directory` argument

## get_yaml_path.bbi_nonmem_model()

1. We want to allow for periods in the `.path` argument to `new_model()`. `new_model()` calls `create_model_object()`, which calls `save_model_yaml()`, which calls `get_yaml_path()`. `get_yaml_path()` calls `yaml_ext()`, but `yaml_ext("foo.bar")` will return `foo.yaml`. Thus, to allow periods in `.path`, we need to modify `get_yaml_path.bbi_nonmem_model()` to append the `yaml` extension to  `.bbi_object[[ABS_MOD_PATH]]` rather than allowing `yaml_ext()` to replace the "extension" (anything after the last period).

## find_nonmem_model_file_path()

1. Similar to `yaml_ext()`, `ctl_ext()` and `mod_ext()` will replace anything after the last period. To support periods in `.path`, we need to append `.ctl` or `.mod` to the `.path` argument rather than calling `ctl_ext()` or `mod_ext()`.

# Test changes

1. Modify tests that call `new_model()` so that they pass an appropriate `.path`.

2. Add a test that `new_model()` supports a `.path` that contains periods.

3. Modify a test of `submit_models()` that calls `copy_model_from()` where `.new_model` has an extension (this was already not supported, but passed because the extension was removed elsewhere).

4. Update vignettes that call `new_model()`.

copernican:
Closes #179.

### DIFF
--- a/R/new-model.R
+++ b/R/new-model.R
@@ -9,8 +9,9 @@
 #' `run_log()` later. Will look for an associated model file (control stream) on
 #' disk and throw an error if it doesn't find one.
 #'
-#' @param .yaml_path Path to save resulting model YAML file to. MUST be either
-#'   an absolute path, or a path relative to the `.directory` argument.
+#' @param .path Path to save the new model. Will be the path to the model file
+#'   and YAML file (both without extension), and the path to the output
+#'   directory.
 #' @param .description Description of new model run. This will be stored in the
 #'   yaml (and can be viewed later in `run_log()`). By convention, it should
 #'   match the $PROBLEM statement in the control stream, but this is not
@@ -28,51 +29,48 @@
 #'   `.yaml_path`. If `TRUE` overwrite existing file, if one exists.
 #' @param .model_type Character scaler to specify type of model being created
 #'   (used for S3 class). Currently only `'nonmem'` is supported.
-#' @param .directory Model directory which `.yaml_path` is relative to. Defaults
-#'   to `options('rbabylon.model_directory')`, which can be set globally with
-#'   `set_model_directory()`.
-#' @importFrom yaml write_yaml
-#' @importFrom fs file_exists
-#' @importFrom digest digest
+#'
 #' @return S3 object of class `bbi_{.model_type}_model` that can be passed to
 #'   `submit_model()`, `model_summary()`, etc.
 #' @export
 new_model <- function(
-  .yaml_path,
+  .path,
   .description,
   .based_on = NULL,
   .tags = NULL,
   .bbi_args = NULL,
   .overwrite = FALSE,
-  .model_type = c("nonmem"),
-  .directory = get_model_directory()
+  .model_type = c("nonmem")
 ) {
 
-  # TODO: update once numeric "dispatch" is formally deprecated
-  if (fs::path_ext(as.character(.yaml_path)) != "yaml") {
-      warning(glue("Did not pass a YAML extension to .yaml_path. Inferred path `{yaml_ext(.yaml_path)}` from `{.yaml_path}`"))
-      .yaml_path <- yaml_ext(.yaml_path)
-  }
-
-  # check for .directory and combine with .yaml_path
-  .yaml_path <- combine_directory_path(.directory, .yaml_path)
-
+  maybe_yaml_path <- paste0(.path, ".yaml")
   # check if file already exists
-  if (fs::file_exists(.yaml_path) && !isTRUE(.overwrite)) {
-    stop(paste(glue("Passed {.yaml_path} to `new_model(.yaml_path)` but that file already exists."),
-               "Either call `read_model()` to load model from YAML or use `new_model(.overwrite = TRUE)` to overwrite the existing YAML."))
+  if (fs::file_exists(maybe_yaml_path) && !isTRUE(.overwrite)) {
+    stop(
+      glue::glue(
+        "File already exists at {maybe_yaml_path}.",
+        "Either call `read_model()` to load model from YAML or use,",
+        "`new_model(.overwrite = TRUE)` to overwrite the existing YAML.",
+        .sep = " "
+      )
+    )
   }
 
-  model_working_directory <- normalizePath(dirname(.yaml_path))
-  model_id <- get_model_id(.yaml_path)
+  # construct the absolute model path in a way that avoids a warning from
+  # normalizePath() if `.path` does not exist (we only require that the model
+  # file exist at `.path`)
+  abs_mod_path <- file.path(
+    normalizePath(dirname(.path)),
+    basename(.path)
+  )
 
   # fill list from passed args
   .mod <- list()
-  .mod[[ABS_MOD_PATH]] <- file.path(model_working_directory, model_id)
+  .mod[[ABS_MOD_PATH]] <- abs_mod_path
   .mod[[YAML_DESCRIPTION]] <- .description
   .mod[[YAML_MOD_TYPE]] <- .model_type
   if (!is.null(.based_on)) {
-    .mod[[YAML_BASED_ON]] <- safe_based_on(model_working_directory, .based_on)
+    .mod[[YAML_BASED_ON]] <- safe_based_on(dirname(.path), .based_on)
   }
   if (!is.null(.tags)) .mod[[YAML_TAGS]] <- .tags
   if (!is.null(.bbi_args)) .mod[[YAML_BBI_ARGS]] <- .bbi_args
--- a/tests/testthat/test-new-model.R
+++ b/tests/testthat/test-new-model.R
@@ -146,4 +146,17 @@ withr::with_options(list(rbabylon.model_directory = NULL), {
     )
   })
 
+  test_that("new_model() supports `.path` containing a period", {
+    temp_ctl <- tempfile(pattern = "file.", fileext = ".ctl")
+    # ensure that `temp_ctl` exists
+    readr::write_file("foo", temp_ctl)
+
+    # this will be created by new_model()
+    temp_yaml <- fs::path_ext_set(temp_ctl, ".yaml")
+    on.exit(fs::file_delete(c(temp_ctl, temp_yaml)))
+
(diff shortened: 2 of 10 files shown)


R/new-model.R
```diff
@@ -28,51 +29,48 @@
...
+    )
   }
 
-  model_working_directory <- normalizePath(dirname(.yaml_path))
-  model_id <- get_model_id(.yaml_path)
+  # construct the absolute model path in a way that avoids a warning from
+  # normalizePath() if `.path` does not exist (we only require that the model
+  # file exist at `.path`)
+  abs_mod_path <- file.path(
+    normalizePath(dirname(.path)),
```
seth127, line 63: should this have `mustWork = TRUE`? Am I right that the _directory_ we are writing into needs to exist? If so, might as well catch it here.

tests/testthat/test-workflow-bbi.R
```diff
@@ -48,13 +48,18 @@ withr::with_options(list(rbabylon.bbi_exe_path = BBI_PATH,
...
+    # created
+
     # create model spec
-    mod1 <- suppressWarnings(new_model(
-      .yaml_path = 1,
-      .description = ORIG_DESC,
-      .tags = ORIG_TAGS,
-      .bbi_args = list(overwrite = TRUE, threads = 4)
-    ))
+    mod1 <- suppressWarnings(
```
seth127, line 55: this is only semi-related, but do we know what warnings this is intended to suppress? I generally don't think we should use `suppressWarnings` in tests (much preferring `suppressSpecificWarning` if we need to catch something unrelated to the test).

I only ask here because I'm wondering if it's related to behavior that changed in this PR.

tests/testthat/test-workflow-bbi.R
```diff
@@ -76,6 +81,7 @@ withr::with_options(list(rbabylon.bbi_exe_path = BBI_PATH,
   })
 
   test_that("copying model works and new models run correctly", {
+    # TODO: this test relies on model-examples-bbi/1.ctl being present
```
seth127, line 84: this is created on line 44, outside of any other test in a "setup" section. Is this TODO intended to put this section inside an explicitly named `setup_` function? 

I guess my point is: many of our tests rely on creating some state at the beginning of the test file and relying on it for multiple different tests. This is different than the tests themselves being dependent on each other, which I _think_ is your concern here.

vignettes/getting-started.Rmd
```diff
@@ -26,6 +27,7 @@ cleanup <- function() {
   if (fs::dir_exists("../inst/nonmem/2")) fs::dir_delete("../inst/nonmem/2")
   if (fs::dir_exists("../inst/nonmem/3")) fs::dir_delete("../inst/nonmem/3")
   if (fs::file_exists("../inst/nonmem/babylon.yaml")) fs::file_delete("../inst/nonmem/babylon.yaml")
+  if (fs::file_exists(file.path(nonmem_path, "1.yaml"))) fs::file_delete(file.path(nonmem_path, "1.yaml"))
```
seth127, line 30: I'm not sure what the point is here? Doesn't this do the same thing as line 22? And if the point is the prefer this patten, then let's go ahead and change all of these lines.

vignettes/getting-started.Rmd
```diff
@@ -113,7 +115,10 @@ To begin modeling, first create a model object. This is an S3 object with an acc
 The `.yaml_path` argument should have the same name (but with a `.yaml` extension) as your base model control stream. For instance, the command below assumes you have a control stream name `1.ctl` or `1.mod` in the model directory you just defined. The `.description` argument is required and corresponds to what would be in the `$PROB` definition in a control stream.
 
 ```{r first model}
-mod1 <- new_model(.yaml_path = "1.yaml", .description = "our first model")
+mod1 <- new_model(
+  system.file("nonmem/1", package = "rbabylon"),
```
seth127, line 119: I feel strongly this should look like a pattern that a user (preferably a Metrum user) would use in real code. Certainly that would _not_ be `system.file`.

I guess this is going to change once we actually deprecate the `set_model_directory()`, so we can cross that bridge when we come to it... can you put a TODO in a vignette? If so, let's do that here. Or actually maybe just put a comment in here about it https://github.com/metrumresearchgroup/rbabylon/issues/165

vignettes/using-based-on.Rmd
```diff
@@ -63,7 +63,10 @@ if (Sys.which("bbi") == "") {
 The modeling process will always start with an initial model, which we create with the `new_model()` call.
 
 ```{r first model}
-mod1 <- new_model(.yaml_path = "1.yaml", .description = "one compartment base model")
+mod1 <- new_model(
+  system.file("nonmem/1", package = "rbabylon"),
```
seth127, line 67: same as above

tests/testthat/test-new-model.R
```diff
@@ -137,7 +137,7 @@ withr::with_options(list(rbabylon.model_directory = NULL), {
     expect_error(
       suppressSpecificWarning({
```
seth127, line 138: this `suppressSpecificWarning` should go away. I'm pretty sure we missed it earlier because it's wrapped in an `expect_error` (that still correctly errors).

seth127 (changes requested):
Just a couple small things. Looks good overall.

R/new-model.R
copernican: What do you mean by 

> the _directory_ we are writing into

If the model file is at `/path/to/foo/bar.ctl`, we require only that `bar.ctl` exists in `/path/to/foo`. This check is performed by `find_nonmem_model_file_path()`, which we call in `create_model_object()`.

tests/testthat/test-workflow-bbi.R
copernican: Good point. We can just remove this `suppressWarnings()` call.

tests/testthat/test-workflow-bbi.R
copernican: Yes, exactly. This test depends on another test, which we should avoid. The point of adding the `TODO` was to remind us to think about how to isolate this test.

vignettes/getting-started.Rmd
copernican: This change is a consequence of deprecating the `.directory` argument to `new_model()`. Now, the `.path` argument needs to indicate the absolute model path. `system.file()` is a way to (reliably) find files in packages using the structure of the installed package on disk, but I agree it may not be user-oriented. We could instead use

```r
mod1 <- new_model(file.path(MODEL_DIR, "1"), .description = "our first model")
```

which might be clearer (though I'm not sure we want to have language that encourages users to think of a "model directory").

vignettes/getting-started.Rmd
copernican: It does not do the same thing as line 22. If we comment this line out and run `check()`, we get the error

```
Error: System command 'R' failed, exit status: 1, stdout + stderr:
E> * checking for file ‘/data/home/seanw/git/rbabylon/DESCRIPTION’ ... OK
E> * preparing ‘rbabylon’:
E> * checking DESCRIPTION meta-information ... OK
E> * installing the package to build vignettes
E> * creating vignettes ... ERROR
E> --- re-building ‘getting-started.Rmd’ using rmarkdown
E> --- finished re-building ‘getting-started.Rmd’
E> 
E> --- re-building ‘parameter-labels.Rmd’ using rmarkdown
E> --- finished re-building ‘parameter-labels.Rmd’
E> 
E> --- re-building ‘using-based-on.Rmd’ using rmarkdown
E> Quitting from lines 66-70 (using-based-on.Rmd) 
E> Error: processing vignette 'using-based-on.Rmd' failed with diagnostics:
E> File already exists at /tmp/RtmpZci0UO/Rinst61a6bd9d387/rbabylon/nonmem/1.yaml. Either call `read_model()` to load model from YAML or use, `new_model(.overwrite = TRUE)` to overwrite the existing YAML.
E> --- failed re-building ‘using-based-on.Rmd’
```

The reason is a consequence of the `system.file()` change below: it creates `1.yaml` at a different location than the previous `new_model()` call. But, if we instead use the pattern `file.path(MODEL_DIR, "1")`, then line 22 will (continue to) work, and we can remove this line.

vignettes/using-based-on.Rmd
copernican: We can resolve this by passing `.path = "../inst/nonmem/1"` rather than calling `system.file()`.

R/new-model.R
seth127: ah yes, I forgot that we check for the model file. That will verify that the directory exists.

seth127 (approved):
Looks good.

PR #222, opened 2020-10-05
Refactor/copy model from args

### LINKED ISSUE #216: remove .directory argument from copy_model_from()
# Summary

#141 will require that the YAML file (for the model, not `babylon.yaml`) and model file have the same path excluding the file extension (`.ctl` or `.mod` for the model file). We will also deprecate the `.directory` argument as part of a broader refactor. Therefore, we should modify `copy_model_from()` by

- removing the `.directory` argument
- changing the way `.new_model` is converted to an absolute path in the S3 dispatches, before being passed through to the implementation functions. Specifically, `.new_model` must have the following properties:
  - no file extension
  - either be an absolute path _or_ be a path relative to the location of the parent model (i.e. `get_model_working_directory(.parent_mod)`)

# Technical specification

## copy_model_from
- update roxygen docs to 
  - remove `.directory` arg
  - update `.new_model` as described above

## copy_model_from.bbi_nonmem_model

- remove `.directory` argument
- remove `combine_directory_path` call and instead do the following with `.new_model`
  - call `new_model <- as.character(.new_model)`
  - check if absolute path and, if not, convert to absolute with `fs::path_rel(.new_model, get_model_working_directory(.parent_mod))`

# Test changes
There are 42 references to copy_model_from in all tests combined.
```
$ grep -rn copy_model_from tests | wc -l
      42
```
The majority of these will change because .new_model will now be relative to .parent_mod instead of relative to the working directory. Unless otherwise noted, that is the required change.

## setup-workflow-ref.R
- add a constant for LEVEL2_SUBDIR <- "level2"
- change LEVEL2_DIR to file.path(MODEL_DIR, LEVEL2_SUBDIR)
- in create_all_models()
  - wrap NEW_MOD2 and NEW_MOD3 in basename
  - change LEVEL2_MOD to file.path(LEVEL2_SUBDIR, "1")
  - in create_rlg_models()
  - wrap NEW_MOD2 and NEW_MOD3 in basename

## test-get-based-on.R
- in "get_based_on works happy path model object" and "get_based_on works happy path character" wrap NEW_MOD2 in basename()
- in "get_model_ancestry works on run_log tibble with more complicated ancestry"
  - change all the file.path(MODEL_DIR, <int>) to just <int>
  - change all the file.path(LEVEL2_DIR, <int>) to just file.path(LEVEL2_SUBDIR, <int>)

## test-model-summary.R
Wrap NEW_MOD2 in basename() in all four occurrences of the following line: mod2 <- MOD1 %>% copy_model_from(NEW_MOD2, .description = "number 2")

## test-model-summaries.R
- Change two copy_model_from calls in the setup() function
  - first try just removing these two lines and calling create_rlg_models() instead. If that doesn't work...
  - remove .directory arg
  - wrap NEW_MOD2 and NEW_MOD3 in basename

## test-config-log.R and test-run-log.R
change line LEVEL2_MOD to file.path(LEVEL2_SUBDIR, "1") in the following line: copy_model_from(MOD1, LEVEL2_MOD, "level 2 copy of 1.yaml", .inherit_tags = TRUE)

## test-submit-models.R
I actually don't think any of these need to change, but the following might need to happen
- in "submit_models() works for models in different directories"
  - remove new_dir <- "level2"
  - change fs::dir_create(file.path(MODEL_DIR, new_dir)) to fs:dir_create(LEVEL2_DIR)
  - change file.path(new_dir, CTL_FILENAME) to file.path(LEVEL2_SUBDIR), CTL_FILENAME)

## test-copy-model-from.R
- add test for integer input of `.new_model`
- change paths in all the other tests

seth127:
closes #216


R/copy-model-from.R
```diff
@@ -9,9 +9,11 @@
 #' ancestry. See ["Using based_on field"
 #' vignette](../articles/using-based-on.html) for details.
 #' @param .parent_mod Model to copy from
-#' @param .new_model Path to write new model files to **without file extension**.
-#'   Function will create both `{.new_model}.yaml` and a new model file based on
-#'   this path.
+#' @param .new_model Path, relative to location of `.parent_model`, to write new
+#'   model files to. Should have **no file extension** because function will
```
copernican, line 13: We want to allow the new model to have a period in its absolute model path, right? There isn't actually any logic to enforce that `.new_model` does not have a file extension, so the issue will arise in `copy_nonmem_model_from()` when `fs::path_set_ext()` is called. That call will _replace_ the existing "extension," so that `copy_control_stream()` will not copy the model file to the correct path, and then `find_nonmem_model_file_path()` will not find it.

```r
library(rbabylon)
parent <- read_model("~/git/rbabylon/tests/testthat/model-examples/1.yaml")
copy_model_from(parent, "foo.bar", "baz")
#> Error in find_nonmem_model_file_path(res[[ABS_MOD_PATH]], .check_exists = TRUE): No model file found at /data/home/seanw/git/rbabylon/tests/testthat/model-examples/foo.bar.ctl or /data/home/seanw/git/rbabylon/tests/testthat/model-examples/foo.bar.mod. Please put relevant model file in that location.
```

R/copy-model-from.R
```diff
@@ -195,3 +191,15 @@ copy_control_stream <- function(.parent_model_path, .new_model_path, .overwrite,
     fs::file_copy(.parent_model_path, .new_model_path)
   }
 }
+
+#' Private helper to build absolute path for copy_model_from
+#' @inheritParams copy_model_from
+#' @keywords internal
```
copernican, line 197: Recommend adding an `@return` tag.

tests/testthat/test-model-summaries.R
```diff
@@ -22,8 +22,9 @@ test_mod_sums <- function(mod_sums) {
 
 setup({
   cleanup()
-  invisible(copy_model_from(MOD1, NEW_MOD2, "model from test-model-summaries.R", .directory = "."))
-  invisible(copy_model_from(MOD1, NEW_MOD3, "model from test-model-summaries.R", .directory = "."))
+  # invisible(copy_model_from(MOD1, basename(NEW_MOD2), "model from test-model-summaries.R", .directory = "."))
```
copernican, line 25: Remove commented-out code.

copernican (changes requested):
This in general looks good, one outstanding issue about supporting `.new_model` with a period.

R/copy-model-from.R
seth127: mmm, interesting. I saw your comment about this somewhere else I must've forgotten to deal with it. I'll add a test for that too.

copernican: We might want to remove the "no file extension" language. It may be clearer to have something like

> Path to the new model, either absolute or relative to the path to `.parent_model`. Represents an absolute model path, which is the path to the YAML file and model file, both without extension, and (possibly) the output directory. Numeric values will be coerced to character. See examples for usage.

and then `@examples` showing the allowed options.

```r
#' @examples
#' \dontrun{
#' parent <- read_model("/foo/parent")
#'
#' # create model file at /bar/child.ctl and YAML at /bar/child.yaml
#' copy_model_from(parent, "/bar/child", "child model with absolute path")
#'
#' # create model file at /foo/child.ctl and YAML at /foo/child.yaml
#' copy_model_from(parent, "child", "relative to parent model path")
#'
#' mod1 <- read_model("/path/to/1")
#'
#' # create model file at /path/to/2.ctl and YAML at /path/to/2.yaml
#' copy_model_from(mod1, 2, "numeric input works")
#'
#' # create model file at /path/to/100.1.ctl and YAML at /path/to/100.1.yaml
#' copy_model_from(mod1, 100.1, "a period is okay")
#' }
```

copernican (approved):
LGTM

PR #234, opened 2020-10-09
Refactor/no output dir arg

### LINKED ISSUE #203: intercept output_dir in bbi_args
`bbi` allows the user to specify an output directory, which defaults to `{{ NAME }}` i.e. the name of the model file, without extension. However, if the users _does_ pass through a different value for this argument (either through `submit_model(.bbi_args)` or `bbi_args` in the model YAML or in their `babylon.yaml` file) this change with _not_ be reflected in the model object and therefore `model_summary()` will not work because it will not be able to find the output directory.

The proposal here is to add a check in `submit_model()` that will check to see if `output_dir` has been passed through to `bbi_args` (either via the argument or the model YAML) and, if it has, error and tell the user that this is not supported in rbabylon. 

Note that it would still be possible for a user to change the value in `babylon.yaml` and rbabylon would have no way of knowing about this. It's not clear that there is any solution to that issue, though it seems relatively unlikely that a user would set a global `output_dir` in their `babylon.yaml`.

Also note: previous to #182 it would be _possible_ for the user to follow this pattern and _manually_ update the `'output_dir'` field of the model object such that `model_summary()` would work. After #182 this will _not_ be possible so it becomes even more perilous to allow `output_dir` to be modified.

copernican:
Closes #203.

### DIFF
--- a/R/aaa.R
+++ b/R/aaa.R
@@ -24,7 +24,6 @@ BBI_ARGS = list(
   no_ext_file = list(type = "logical", flag="--no-ext-file", description = "do not use ext file"),
   no_grd_file = list(type = "logical", flag="--no-grd-file", description = "do not use grd file"),
   no_shk_file = list(type = "logical", flag="--no-shk-file", description = "do not use shk file"),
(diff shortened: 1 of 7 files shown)


seth127 (approved):
Looks good

PR #236, opened 2020-10-09
test: wrap tempfile() and tempdir() in normalizePath()

### LINKED ISSUE #235: temporary files and directories don't match on macOS
Various tests are failing on macOS, all with similar errors:

```r
test-get-path-from-object.R:41: failure: get_model_path() finds .mod path
get_model_path(mod) not identical to `temp_mod`.
1/1 mismatches
x[1]: "/private/var/folders/d3/x8vvcfl93bzdqyjldg5kj6cr0000g
x[1]: p/T/RtmpHa7Y1M/file45e2210917f9.mod"
y[1]: "/var/folders/d3/x8vvcfl93bzdqyjldg5kj6cr0000gp/T/Rtmp
y[1]: Ha7Y1M/file45e2210917f9.mod"
```

In each case, the failure occurs because the result refers to `/private/foo/bar` and the expectation to `/foo/bar`. This is a consequence of the fact that `/tmp` and `/var` are symlinks on macOS. From `?normalizePath`:

> For example, on macOS absolute paths such as ‘/tmp’ and ‘/var’ are symbolic links.

In particular, they are symlinks into `/private`:

```sh
➜  ~ uname
Darwin
➜  ~ ls -ld /tmp /var
lrwxr-xr-x@ 1 root  admin  11 Apr  6  2020 /tmp -> private/tmp
lrwxr-xr-x@ 1 root  admin  11 Apr  6  2020 /var -> private/var
```

But `tempdir()` is not guaranteed to return a canonical path:

> On a Unix-alike, the value will be an absolute path (unless tmpdir is set to a relative path), but it need not be canonical (see normalizePath) and on macOS it often is not.

Thus, any path that goes through `normalizePath()` will be canonical, e.g., include `/private`, and paths that do not, will not necessarily be canonical. We do not observe this behavior on Ubuntu because `/var` and `/tmp` are not symlinks.

To resolve this, we should ensure that calls to `tempfile()` or `tempdir()` in our tests are wrapped in `normalizePath()`, at least the ones that are run on macOS.

copernican:
Closes #235.

### DIFF
--- a/tests/testthat/test-submit-model.R
+++ b/tests/testthat/test-submit-model.R
@@ -61,6 +61,7 @@ test_that("submit_model(.dry_run=T) with bbi_nonmem_model object parses correctl
 test_that("submit_model() creates correct call for non-NULL .config_path", {
   temp_config <- tempfile(fileext = ".yaml")
   readr::write_file("foo", temp_config)
+  temp_config <- normalizePath(temp_config)
   on.exit(fs::file_delete(temp_config))
 
   res <- submit_model(MOD1, .config_path = temp_config, .dry_run = TRUE)
--- a/tests/testthat/test-submit-models.R
+++ b/tests/testthat/test-submit-models.R
@@ -118,6 +118,7 @@ withr::with_options(list(rbabylon.bbi_exe_path = "bbi"), {
   test_that("submit_models() works with non-NULL .config_path", {
     temp_config <- tempfile(fileext = ".yaml")
     readr::write_file("foo", temp_config)
+    temp_config <- normalizePath(temp_config)
     on.exit(fs::file_delete(temp_config))
 
     res <- submit_models(
--- a/tests/testthat/setup-workflow-ref.R
+++ b/tests/testthat/setup-workflow-ref.R
@@ -234,5 +234,6 @@ create_temp_model <- function(path = YAML_TEST_FILE,
   temp_ctl <- fs::path_ext_set(temp_yaml, mod_ext)
   readr::write_file(mod_content, temp_ctl)
   withr::defer(fs::file_delete(c(temp_yaml, temp_ctl)), envir)
-  fs::path_ext_remove(temp_yaml)
+  # normalizePath() needs to be called when the file actually exists
+  fs::path_ext_remove(normalizePath(temp_yaml))
 }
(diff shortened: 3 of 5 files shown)


tests/testthat/test-submit-model.R
```diff
@@ -61,6 +61,7 @@ test_that("submit_model(.dry_run=T) with bbi_nonmem_model object parses correctl
 test_that("submit_model() creates correct call for non-NULL .config_path", {
   temp_config <- tempfile(fileext = ".yaml")
   readr::write_file("foo", temp_config)
```
seth127, line 63: shouldn't the `normalizePath` be _above_ the `write_file` call? Is there a reason you don't just do 
```
temp_config <- normalizePath(tempfile(fileext = ".yaml"))
```
?

tests/testthat/test-submit-models.R
```diff
@@ -118,6 +118,7 @@ withr::with_options(list(rbabylon.bbi_exe_path = "bbi"), {
   test_that("submit_models() works with non-NULL .config_path", {
     temp_config <- tempfile(fileext = ".yaml")
     readr::write_file("foo", temp_config)
+    temp_config <- normalizePath(temp_config)
```
seth127, line 121: same as above

seth127 (approved):
This passes on my Mac so it's approved. I am a little confused about that `temp_config` call I commented on, but I guess that's my bad.

tests/testthat/test-submit-model.R
copernican: The call to `normalizePath()` must happen after `write_file()`, because the path given by `temp_config()` does not exist until after that. `tempfile()` does not _create_ a file, but rather returns

> a character vector giving the names of possible (temporary) files. Note that no files are generated by tempfile.

Thus, calling `normalizePath(tempfile())` will raise a warning, because the temporary file does not yet exist. But, after we write to it (or otherwise create it), it will exist, and `normalizePath()` will work as intended.

```r
x <- tempfile()
x
#> [1] "/var/folders/d3/x8vvcfl93bzdqyjldg5kj6cr0000gp/T//RtmplPceD2/file591b28bc8217"
normalizePath(x)
#> Warning in normalizePath(x): path[1]="/var/folders/d3/
#> x8vvcfl93bzdqyjldg5kj6cr0000gp/T//RtmplPceD2/file591b28bc8217": No such file or
#> directory
#> [1] "/var/folders/d3/x8vvcfl93bzdqyjldg5kj6cr0000gp/T//RtmplPceD2/file591b28bc8217"
file.exists(x)
#> [1] FALSE
writeLines("foo", x)
normalizePath(x)
#> [1] "/private/var/folders/d3/x8vvcfl93bzdqyjldg5kj6cr0000gp/T/RtmplPceD2/file591b28bc8217"
```

PR #237, opened 2020-10-09
Docs/update vignettes post mod dir

### LINKED ISSUE #165: Update vignettes to reflect no rbabylon.model_directory
All of the vignettes make some reference to setting `options('rbabylon.model_directory')` which will need to be taken out and re-written to reflect the new design.

Also, there are places like https://github.com/metrumresearchgroup/rbabylon/blob/develop/vignettes/using-summary-log.Rmd#L177 that reference other functionality (S3 dispatches etc.) that will be deprecated. Those will need to be rewritten too.

seth127:
closes #165


vignettes/getting-started.Rmd
```diff
@@ -232,16 +240,16 @@ mod3 <- mod3 %>%
...
 
 ```{r run_log} 
 log_df <- run_log(MODEL_DIR)
 log_df
 ```
 
-The `run_log()` returns a tibble (called `log_df` in this example) which can be manipulated like any other tibble. However, several of the columns (`tags` and `based_on` for example) are list columns, which complicates how the user interacts with them. Future releases will have  helper functions to more seamlessly interact with `log_df`, but until then, we have provided some sample `tidyverse` code below.
+The `run_log()` returns a tibble which can be manipulated like any other tibble. However, several of the columns (`tags` and `based_on` for example) are list columns, which complicates how you can interact with them. Future releases will have helper functions to more seamlessly interact with these log tibbles, but until then, we have provided some sample `tidyverse` code below.
```
copernican, line 252: typo: "The `run_log()` returns"

copernican (approved):
One possible typo, but otherwise LGTM.

PR #252, opened 2020-10-24
Feat/remove tags

### LINKED ISSUE #249: remove_tags function
We have `add_tags()` and `replace_tags()`, but no way to remove a tag from a model (except by replacing _all_ tags currently on a model). 

# Functional Spec

* API should be just like `add_tags()` except all strings in the vector passed to `.tags` should be removed from `.mod$tags` instead of added.
* Should raise a warning for each string in `.tags` that are _not_ currently in `.mod$tags` (mainly to help catch spookiness due to typos, since this is just free text).

seth127:
closes #249


tests/testthat/test-modify-model-field.R
```diff
@@ -54,6 +61,25 @@ test_that("add_tags() and replace_tags() work correctly", {
   expect_identical(new_mod[[YAML_TAGS]], NEW_TAGS)
 })
 
+test_that("remove_tags() works correctly", {
```
kylebaron, line 64: do you want to reference replace_tags  here too?

R/modify-model-field.R
```diff
@@ -19,16 +19,28 @@
 #' @param .field Character scalar of the name of the component to modify
 #' @param .value Whatever is to be added to `.mod[[.field]]`, typically a character vector
 #' @param .append If `TRUE`, the default, concatenate new values with currently present values. If `FALSE`, new values will overwrite old values.
+#' @param .remove If `TRUE`, `.value` with be removed from the `.field` instead of added. `FALSE` by default. Cannot have both `.append` and `.remove` be true in the same call.
 #' @param .unique If `TRUE`, the default, de-duplicate `.mod[[.field]]` after adding new values. If `FALSE` duplicate values will be kept.
 #' @export
-modify_model_field <- function(.mod, .field, .value, .append = TRUE, .unique = TRUE) {
+modify_model_field <- function(.mod, .field, .value, .append = TRUE, .remove = FALSE, .unique = TRUE) {
```
kylebaron, line 25: Feels like this needs `@keywords internal` in the documentation.  Or don't export it?

kylebaron (approved):
See comments.

R/modify-model-field.R
seth127: yeah, we export it now, and warn people not to use it directly. I don't know, maybe it should be internal, but I think it's fine at this point.

kylebaron: `@keywords internal` would still keep it exported, but drop it from the available functions list.  Might be good compromise if you really don't want people to use it.

Issue #134, opened 2020-09-02
Do NOT replace $PROB with .description in copy_model_from


seth127:
Both @anarpkpd and @kylebaron feel strongly that `copy_model_from()` should _not_ replace the `$PROB` in the new control stream with the `.description` argument. Instead they think it should just (probably optionally) replace `$PROB` with the new run number. 

This is primarily motivated by the pain in keeping the text in sync in `$PROB` (in the control stream) and `description` (in the yaml). The solution is to side step that by just _not_ keeping anything meaningful in `$PROB` and using the `description` in the yaml instead.

seth127:
Note that this may be obsolete due to the change made in #254. See comment here https://github.com/metrumresearchgroup/rbabylon/pull/256#discussion_r512206104

seth127:
New decision: replace $PROB with something like `glue("See {.new_model}.yaml. Created by rbabylon.")`

PR #256, opened 2020-10-26
Feat/desc not required

### LINKED ISSUE #254: Make .description argument optional
# Summary

As a user, I would like to _not_ need to enter a `.description`. It should have no default value, and be empty until something is added, similar to `tags` and `notes`.

# Tests
* tests/testthat/test-copy-model-from.R  
  * copy_from_model creates accurate copy
  * copy_from_model options work
  * copy_from_model .overwrite=TRUE works

seth127:
closes #254


R/copy-model-from.R
```diff
@@ -191,17 +199,13 @@ copy_control_stream <- function(.parent_model_path, .new_model_path, .overwrite,
...
-    new_mod_id <- .new_model_path %>% get_model_id()
-
     # read parent control stream
     mod_str <- .parent_model_path %>% read_file()
 
     # replace the $PROBLEM line(s)
     mod_str <- str_replace(mod_str,
                            "\\$PROB(.|\n)*?\\$",
-                           as.character(glue("$PROBLEM {new_mod_id} {.description}\n\n$")))
+                           as.character(glue("$PROBLEM {.description}\n\n$")))
```
kylebaron, line 208: Just noting here that the update only happens if user opts in.

kylebaron (approved):
WIth comment.

R/copy-model-from.R
seth127: right, although the default is to update when using `copy_model_from()`. As an interesting note, this makes #134 maybe unnecessary. The idea there was to _not_ replace `$PROB` with the `.description` but instead with the run number. But now the `.description` defaults to the run number, so that gets us to the same place (at least by default). We can talk about that on the 134 issue though.

PR #260, opened 2020-10-27
feat/collapse list columns

### LINKED ISSUE #263: Collapse list columns to character
# Summary
As a user, I would like to have a function that can take a tibble and the names of some list columns, and collapse the list columns to a single string per row. An example of this is the `tags` column in the the `bbi_run_log_df`. It is a list of character vectors and this function should convert it to character column with a single string per row which represents pasting together the elements of the character previously in that row.

# Tests
- tests/testthat/test-run-log.R
  - collapse_to_string() works correctly
  - collapse_to_string() warns and errors correctly

seth127:
closes #263


seth127:
Decided on this over #261 for the following reasons:

* there was no clear advantage to implementing the generic
* the docs were harder to find with the generic
* the snake case of this one conforms with the rest of the package

tests/testthat/test-run-log.R
```diff
@@ -91,8 +137,13 @@ test_that("run_log() works correctly with nested dirs", {
 test_that("run_log fails after messing with YAML", {
   # make the description field an array
   rogue_yaml <- yaml::read_yaml(yaml_ext(NEW_MOD3))
-  rogue_yaml[[YAML_DESCRIPTION]] <- c(rogue_yaml[[YAML_DESCRIPTION]], "bad stuff")
+  orig_desc <- rogue_yaml[[YAML_DESCRIPTION]]
```
seth127, line 140: this is unrelated, but this test wasn't cleaning up properly so I fixed it

R/run-log.R
```diff
@@ -32,6 +32,79 @@ run_log <- function(.base_dir, .recurse = TRUE) {
...
+    warning(glue("collapse_to_string() only works on list columns. The following columns are not lists and will be ignored: {paste(bad_cols, collapse = ', ')}"))
+  }
+
+  # collapse together any lists of vectors
+  .data %>%
+    mutate(.collapse_key = row_number()) %>%
+    group_by(.data[[".collapse_key"]]) %>%
+    mutate_at(.vars = vars({{loc}}),
+              function(.vec) {
+                if (inherits(.vec, "list")) {
```
kylebaron, line 90: Does this ever evaluate to FALSE?  Dont' we already check that it is a list?

R/run-log.R
```diff
@@ -32,6 +32,79 @@ run_log <- function(.base_dir, .recurse = TRUE) {
...
+                  if (is.null(.vec[[1]])) {
+                    .vec <- NA
+                  } else if (inherits(.vec[[1]], c("character", "numeric", "logical"))) {
+                    .vec <- paste(unlist(.vec), collapse = .sep)
+                  }
+                }
+                .vec
+              }) %>%
+    ungroup(.data[[".collapse_key"]]) %>%
+    select(-.data[[".collapse_key"]])
```
kylebaron, line 100:
```r
df %>% 
  modify_at(.at = loc, .f = function(x) {
    #if(!is.list(x)) return(x) # this should already be a list, no?
    ans <- map(x, .f = function(.vec) {
      if (is.null(.vec)) return(NA)
      if(inherits(.vec, c("character", "numeric", "logical"))) {
        .vec <- paste0(.vec, collapse = .sep)
      }
      .vec
    })
    unlist(ans, use.names = TRUE)
  })
```

kylebaron (commented):
@seth127 what do you think about this alternate code?

R/run-log.R
kylebaron: This does need some more safety logic ... like do nothing if `x[[1]]` isn't character, numeric, or logical.

kylebaron: Updated the snippet .... but this would guarantee character return; and if it finds something that it can't collapse, sub in `NA_character_`

R/run-log.R
seth127: we do check, for the purpose of warning, but I actually don't subset `loc` above. You're right though, that would be a cleaner place to do it. I'll make that change.

R/collapse-to-string.R
```diff
@@ -0,0 +1,72 @@
...
+  # collapse together any lists of vectors
+  .data %>%
+    modify_at(.at = cols, .f = function(x) {
+      map_chr(x, .f = function(.vec) {
+        if (inherits(.vec, c("character", "numeric", "logical"))) {
+          .vec <- paste0(.vec, collapse = .sep)
+        } else if (is.null(.vec)) {
+          .vec <- NA_character_
+        } else {
+          .vec <- paste(
```
kylebaron, line 64: Nice!

kylebaron (approved):

R/collapse-to-string.R
```diff
@@ -0,0 +1,72 @@
...
+    bad_cols <- names(valid_cols)[!valid_cols]
+    warning(glue("collapse_to_string() only works on list columns. The following columns are not lists and will be ignored: {paste(bad_cols, collapse = ', ')}"))
+  }
+  cols <- cols[valid_cols]
+
+  # collapse together any lists of vectors
+  .data %>%
+    modify_at(.at = cols, .f = function(x) {
+      map_chr(x, .f = function(.vec) {
+        if (inherits(.vec, c("character", "numeric", "logical"))) {
```
kylebaron, line 59: @seth127 working with something similar in pmtables; you might want to include `integer` here; it won't get picked up by this:

```r
> inherits(1L, c("character", "logical", "numeric"))
[1] FALSE
```

You could also look at `is.atomic` but that will pick up NULL as well ... but you could test for `NULL` first then test for `is.atomic` etc.

PR #261, opened 2020-10-27
Feat/toString methods

seth127:
WIP: first draft of toSting.bbi_log_df method, added with docs and vignette appearance. Need to write tests.

This is a competing option with https://github.com/metrumresearchgroup/rbabylon/pull/260


seth127:
Went with #260 instead for the reasons listed here https://github.com/metrumresearchgroup/rbabylon/pull/260#issuecomment-717984749

PR #264, opened 2020-10-28
new replace_tag() function

### LINKED ISSUE #262: rename_tags function
### Problem
The current pattern of using add_tags/remove_tags functions to edit existing tags is useful in specific scenarios. However, a more common pattern would be REPLACING a specific tag with a new tag, representing a subtle change between models.

### Proposed Solution
In this scenario, a one compartment model (101) is inadequate, so a two compartment model is used (mod102).
```
mod101 %>%
copy_model_from(102, .inherit_tags=T, .description = "two-cmpt,  prop RUV") %>%
rename_tags(TAGS$two_compartment=TAGS$one_compartment) 
```

I'm not sure I love the function name, but including "rename" will indicate to users that the syntax in this function matches `dplyr::rename` (e.g. new=old).

seth127:
closes #262


NAMESPACE
```diff
@@ -82,14 +82,18 @@ export(print_bbi_args)
...
+export(remove_based_on)
 export(remove_notes)
 export(remove_tags)
+export(replace_all_based_on)
+export(replace_all_bbi_args)
+export(replace_all_notes)
 export(replace_all_tags)
 export(replace_based_on)
-export(replace_bbi_args)
 export(replace_decisions)
```
callistosp, line 93: is this function deprecated?

R/modify-model-field.R
```diff
@@ -248,10 +269,35 @@ replace_bbi_args <- function(.mod, .bbi_args) {
...
+    .field = YAML_DECISIONS,
+    .value = .decisions,
+    .append = TRUE
+  )
+}
+
+#' @describeIn modify_model_field _Deprecaed_ Replaces `decisions` field in a model object and corresponding YAML with new values
+#' @param .decisions Character vector to use as replacement
+#' @export
+replace_decisions <- function(.mod, .decisions) {
```
callistosp, line 288: any reason we don't have remove_decision/replace_all_decisions functions here also?

callistosp (dismissed):
the new functions all look good. I'm not familiar with the internals of the `modify_model_field` function to verify all of these but the arguments look consistent between tag/note/desc/etc. test script seems sufficient to test new functions

NAMESPACE
seth127: Yes, and this is something we should discuss... what should we do about the old functions? For example:

* We used to have `replace_tags()` which deleted all tags on a model and replaced with new ones
* Now we have `replace_tag()` (singular) which replaces a _single_ tag
* Now we also have `replace_all_tags()` which deletes all tags and replaces with new ones (the same thing `replace_tags()` did before)

What should happen with `replace_tags()`? Options include:

1. Remove it (may break some existing user code, but at this point probably not much, if any)
2. Leave it but have it error and say `"Use replace_all_tags() instead"` (may break code like ^)
3. Leave it but have it warn and say `"This is deprecated, use replace_all_tags() instead"` 

As Sam noticed, as it stands we're doing option 1. Option 3 seems safest in some sense, but I worry that leaving it in (and working) could be dangerous because a user may try to call `replace_tag()` and accidentally call the deprecated `replace_tags()` which will blow away _all_ of their tags (in a way that is _not_ easy to undo, unless they're checked into version control). I think I favor option 2 for that reason, but I'm on the fence. Interested in @copernican @dpastoor and @kylebaron perspective on this too.

Note that we will do the same thing for `_notes`, `_based_on` and `_bbi_args` too.

R/modify-model-field.R
seth127: Because that function is already deprecated. The only reason `add_` and `replace_` are still there is so it doesn't break any existing user code that calls them. (Similar to discussion above, but I don't see any danger here, so when I did these I went with option 3 for those two specific functions)

NAMESPACE
seth127: After discussion offline, the decision is to follow the `tidyverse` model and use the `lifecycle` package and use `deprecate_warn()` (option 3 above) for two releases, then `deprecate_stop()` (option 2) for two releases, then remove it (option 1).

callistosp: that seems like a good approach. let me know when it is ready for re-review

callistosp (approved):
LGTM

PR #265, opened 2020-11-02
Changed copy_model_from to replace $PROB with '...Created by rbabylon' message

### LINKED ISSUE #134: Do NOT replace $PROB with .description in copy_model_from

seth127:
Closes #134 

See: https://github.com/metrumresearchgroup/rbabylon/issues/134#issuecomment-719588739

### DIFF
--- a/R/copy-model-from.R
+++ b/R/copy-model-from.R
@@ -201,7 +201,7 @@ copy_control_stream <- function(.parent_model_path, .new_model_path, .overwrite,
     new_mod_id <- get_model_id(.new_model_path)
     mod_str <- str_replace(mod_str,
                            "\\$PROB(.|\n)*?\\$",
-                           as.character(glue("$PROBLEM See {new_mod_id}.yaml. Created by rbabylon.\n\n$")))
+                           as.character(glue("$PROBLEM From rbabylon: see {new_mod_id}.yaml for details\n\n$")))
     glue("")
 
     # read parent control stream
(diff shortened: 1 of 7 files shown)


R/copy-model-from.R
```diff
@@ -180,32 +179,30 @@ copy_nonmem_model_from <- function(
...
 
     # read parent control stream
     mod_str <- .parent_model_path %>% read_file()
 
     # replace the $PROBLEM line(s)
+    new_mod_id <- get_model_id(.new_model_path)
     mod_str <- str_replace(mod_str,
                            "\\$PROB(.|\n)*?\\$",
-                           as.character(glue("$PROBLEM {.description}\n\n$")))
+                           as.character(glue("$PROBLEM See {new_mod_id}.yaml. Created by rbabylon.\n\n$")))
```
kylebaron, line 204: Not a huge deal but ... "Created by rbabylon." might suggest that rbabylon wrote out the control stream.  Not sure.  But is there a different wording that would only suggest that rbabylon only inserted the text in `$PROBLEM`?

```
$PROBLEM From rbabylon: see {new_mod_id}.yaml for details
```

kylebaron (dismissed):
Approve with comment.  If you could just respond to the comment before merging, I'd appreciate it.

kylebaron (approved):

PR #266, opened 2020-11-02
Fix/update news from 264

seth127:
Updating NEWS.md from #264 and #260

### DIFF
--- a/R/modify-model-field.R
+++ b/R/modify-model-field.R
@@ -69,7 +69,7 @@ replace_model_field <- function(.mod, .field, .old_val, .new_val) {
 
   idx <- which(.mod[[.field]] == .old_val)
   if (length(idx) == 0) {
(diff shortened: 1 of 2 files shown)


R/modify-model-field.R
```diff
@@ -286,7 +286,6 @@ replace_bbi_args <- function(.mod, .bbi_args) {
 #' @export
 add_decisions <- function(.mod, .decisions) {
   warning("The `decisions` field has been replaced by `notes` as of rbabylon 0.10.0 and will be removed in a future release. Please use `add_notes()` going forward.")
-  deprecate_warn("0.10.0", "rbabylon::replace_bbi_args()", "replace_all_bbi_args()")
```
seth127, line 289: this is  unrelated. Just noticed this when I was looking around in here. It's just not supposed to be there.

callistosp (approved):
perfection

PR #267, opened 2020-11-02
Removing default assignment for .description

### LINKED ISSUE #254: Make .description argument optional

seth127:
closes #254


tests/testthat/model-examples/1.yaml
```diff
@@ -1,5 +1,5 @@
-description: original acop model
 model_type: nonmem
+description: original acop model
```
seth127, line 2: the `model_type` and `description` fields got re-ordered by this change, which also accounts for all of the md5 hash changes below.

tests/testthat/setup-workflow-ref.R
```diff
@@ -80,16 +80,16 @@ CONFIG_DATA_MD5 <- "4ddb44da897c26681d892aa7be99f74b"
 CONFIG_MODEL_MD5 <- "6b930119c4224ba077091b47959b0604"
 
 # yaml md5 hashes
-MOD1_YAML_MD5 <- "ee5a30a015c4e09bc29334188ff28b58"
-MOD_LEVEL2_MD5 <- "1c18961a778fc6b71e3990b500f4bca5"
-ALL_MODS_YAML_MD5 <- c(MOD1_YAML_MD5, rep("22c7176b3dbbbf932091c82c2fd9f749", 2), "edd548af82d8556971c442f36a3907e0")
-RUN_LOG_YAML_MD5 <- c(MOD1_YAML_MD5, "62fc26bf3bf941c6bad352f23cc6c2ad", "ebd4861c9d0584b03a23edcfdda9a67e")
+MOD1_YAML_MD5 <- "6ccf206e167485b5adf29bc135197929"
```
seth127, line 83: These were all changed because of this https://github.com/metrumresearchgroup/rbabylon/pull/267/files#r516199955

I was initially annoyed by having to change these for something that simple, but I think having `model_type` first is actually better, so this is the right move.

tests/testthat/setup-workflow-ref.R
```diff
@@ -153,13 +153,14 @@ copy_all_output_dirs <- function() {
...
   # copy models before creating run log
   mod1 <- read_model(MOD1_PATH)
-  copy_model_from(mod1, basename(NEW_MOD2), NEW_DESC, .add_tags = NEW_TAGS)
-  copy_model_from(mod1,
-                  basename(NEW_MOD3),
-                  NEW_DESC,
-                  .based_on_additional = get_model_id(NEW_MOD2),
-                  .inherit_tags = TRUE,
-                  .update_model_file = FALSE)
+  copy_model_from(mod1, basename(NEW_MOD2), .add_tags = NEW_TAGS)
```
seth127, line 156: removed the description from these two so that I could test having null descriptions in the run log

R/modify-model-field.R
```diff
@@ -236,6 +236,19 @@ replace_description <- function(.mod, .description) {
   )
 }
 
+#' @describeIn modify_model_field Fills the description field in a model object and corresponding YAML, if it is currently empty.
+#' @importFrom checkmate assert_scalar
+#' @export
+add_description <- function(.mod, .description) {
```
kylebaron, line 242: Honest opinion: this seems complicated having `add_` and `replace_`.   Can we just do `set_description()` or whatever  and leave it at that?

kylebaron (approved):

R/modify-model-field.R
seth127: This was kind of a legacy thing because we already had `replace_description` when it was a required field.

The reason both exist now is because 1) we already had `replace_` and I didn't want to deprecate it and 2) it seemed like (if I didn't make `add_`) people would instinctively want to type `add_description()` and be confused as to why they had to use `replace_` when they weren't actually replacing anything.

PR #268, opened 2020-11-02
Fix/submit models no bbi args

### LINKED ISSUE #248: BUG: submit_models() fails when no bbi_args passed
# Summary

When a user calls `submit_models()`, and at least one of the models has _no_ `bbi_args` attached, the function will error with `Error in .run[["models"]][[1L]] : subscript out of bounds`. This happens both with `.dry_run = T` and `.dry_run = F`

# Reprex

```r
setwd("tests/testthat")
source("setup-workflow-ref.R")

print(MOD1$bbi_args)
# $overwrite
# [1] TRUE
# 
# $threads
# [1] 4

mod_noargs <- copy_model_from(MOD1, "noargs", "model with NO bbi_args") %>% replace_bbi_args(NULL)
print(mod_noargs$bbi_args)
# NULL

proc1 <- submit_models(list(MOD1, mod_noargs), .dry_run = TRUE)
#  Error in .run[["models"]][[1L]] : subscript out of bounds 

mod_args <- copy_model_from(MOD1, "args", "model with bbi_args") %>% replace_bbi_args(list(threads = 2))
print(mod_args$bbi_args)
# $threads
# [1] 2

proc2 <- submit_models(list(MOD1, mod_args), .dry_run = TRUE)
# Submitting 2 models with 2 unique configurations.
length(proc2)
# [1] 2

# cleanup 
fs::file_delete("model-examples/noargs.yaml")
fs::file_delete("model-examples/noargs.ctl")
fs::file_delete("model-examples/args.yaml")
fs::file_delete("model-examples/args.ctl")
```

# Details

I debugged a little and I'm pretty sure this happens on https://github.com/metrumresearchgroup/rbabylon/blob/develop/R/utils.R#L159 because we're trying to match `NULL` to what ends up being `character(0)` and we don't match, so as a result we return no models for that arg set. (See screen shot)

Screen Shot 2020-10-21 at 11 14 36 AM

Then when we get to https://github.com/metrumresearchgroup/rbabylon/blob/develop/R/submit-models.R#L122 we try to do `.run[["models"]][[1L]]` but `.run[["models"]]` is an empty list, so it dies.

copernican:
Closes #248.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: rbabylon
 Title: R package for babylon
-Version: 0.9.0.8002
+Version: 0.9.0.8003
 Authors@R: 
     c(person(given = "Devin",
              family = "Pastoor",
--- a/NEWS.md
+++ b/NEWS.md
@@ -46,6 +46,9 @@ model object `tags` field. (#252)
   
 ## Bug fixes
 
+* `submit_models()` correctly handles the case where bbi arguments are neither
+  present in the YAML file nor passed at runtime (#248).
+
 * `config_log()` and `add_config()` are supposed to warn a user if models are
 found that do _not_ have a corresponding `bbi_config.json` file (i.e. have not
 yet been run, or did not finish successfully). However, if there is no output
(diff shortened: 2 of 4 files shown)


seth127 (approved):
LGTM

PR #270, opened 2020-11-03
fix: replace_description can now accept NA or NULL

### LINKED ISSUE #269: replace_decisions should take NULL and NA
currently you _can't_ blow away the `description` field by doing `replace_description(NULL)` or `replace_description(NA)` because `assert_scalar` errors out.

seth127:
closes #269


callistosp (commented):
I don't see anything wrong with the code implementation here, but I am not sure if we want to allow description to be NA or NULL... would there be a benefit to just having it default to the model number if it is given NA or NULL?

seth127:
The idea is that you should be able to _remove_ a description of you want (by calling replace_description(NULL)). Currently you can kinda do it by replacing with an empty string, but part of the idea was that you'd want to be able to filter a run log to only models that _have_ a description, which is easiest if you can just use filter(!is.na(description)) (or !is.null works too).

I considered making it so that if you passed NA it would function the same as NULL (wipe the field away) but I didn't follow through. I can do that if you think that makes more sense.

callistosp (approved):
nice `na_to_null` function, clever way to deal with the NULL checks

PR #272, opened 2020-11-04
param_labels.Rd typo fix

callistosp:
change "paramemter" -> "parameter"

not sure if this is how to submit typo fixes, let me know if a different method is preferred

### DIFF
--- a/R/param-labels.R
+++ b/R/param-labels.R
@@ -36,7 +36,7 @@
 #'        and covariance of the parameter on the raw scale will be reported.
 #'      }
 #'   }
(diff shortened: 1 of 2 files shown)


seth127:
Nice catch. The `.Rd` files are generated automatically be roxygen. You need to make the change [here](https://github.com/metrumresearchgroup/rbabylon/blob/develop/R/param-labels.R#L39) and then do `devtools::document()` to re-render the `.Rd` file. Otherwise your change will get overwritten as soon as someone else calls `devtools::document()` to render something else.

callistosp:
okay fixed in the correct place now, thanks @seth127

seth127 (approved):
Lovely

PR #276, opened 2020-11-10
use pkgr + renv

copernican:
This PR configures pkgr and renv. This is not a user-visible change, so there is no need to increment the version.

We have decided to use the tip of CRAN for development, in order to catch any breaking changes as quickly as possible and to stay abreast of (other non-breaking) changes to dependencies.

Note that the 2020-03-24 MPN snapshot is currently the oldest compatible snapshot (for CI purposes). This will be checked in CI, and developers can do the following to develop/debug against it locally:

1. change `https://cran.rstudio.com` to `https://mpn.metworx.com/snapshots/stable/2020-03-24` in the `pkgr.yaml`
2. delete the `renv/library` folder
3. run `pkgr install`


renv/.gitignore
```diff
@@ -0,0 +1,3 @@
+library/
+python/
```
seth127, line 2: ?!

renv/activate.R
```diff
@@ -0,0 +1,185 @@
+
+local({
```
seth127, line 2: this is all just Jenny Bryan (or whoever) code from `renv::init()` right?

.drone.jsonnet
```diff
@@ -214,7 +214,7 @@ local pull_image(image, volumes=[]) = {
 # r_path          path to R executable
 # expr            expression to run
 local run_r_expression(r_path, expr) =
-  std.join(" ", [r_path, "-e", std.escapeStringBash(expr)]);
+  std.join(" ", [r_path, "--no-init-file", "-e", std.escapeStringBash(expr)]);
```
seth127, line 217: this is so drone _doesn't_ use renv right? so that the containers are still controlling the packages?

pkgr.yml
```diff
@@ -0,0 +1,28 @@
...
+        Type: source
+  Packages:
+    - Rcpp:
+        Repo: MPN_Rcpp
+        Type: source
+    - withr:
+        Repo: MPN_withr
+        Type: source
+
+Lockfile:
```
seth127, line 27: I don't really understand how this works. And I assume this is what is telling pkgr where to install? I might need a brief tutorial just so I understand this code better, but it all seems to work the way it should. Well done.

seth127 (changes requested):
All of my comments are really just curiosity, but then I tried some things with `devtools::check()` and I'm _pretty_ sure that it's _not_ using the `renv` when we run check. If that's true... not so good. 

Maybe you knew that, but to me being able to run `check` with the known/controlled package versions is kinda the main draw here, right? I'll try to attach some documentation to the PR but if github fails me I'll slack it to you.

seth127 (approved):
Ahh, just kidding! It does work! I had messed something up when I was running `check` before. All looks good. I'll drop in the code I used to test. Well done.

seth127:
I dropped this into the top of the first test file (`test-collapse-to-string.R`)

```
cat(c(
    "Hello, I'm devtools::check()",
    "\n working dir: ", getwd(),
    "\n .libPaths(): ", .libPaths(),
    "\nThe first .libPath contains:\n",
    system(paste("ls -lh", .libPaths()[1]), intern = TRUE)
  ),
  sep = "\n", file = "~/Documents/renv-check-devtools.txt"
)
```
And here's what we got back:

[renv-check-devtools.txt](https://github.com/metrumresearchgroup/rbabylon/files/5534501/renv-check-devtools.txt)

Looks good

.drone.jsonnet
dpastoor: this is not about drone "not using renv" per say - its about R not sourcing the .Rprofile when it starts up, therefore renv does not get activated

renv/activate.R
dpastoor: yes from kevin Ushey mostly - its the bootstrapping script renv uses to activate itself

renv/.gitignore
dpastoor: default stuff renv creates - renv can help manage python packages as well

pkgr.yml
```diff
@@ -0,0 +1,28 @@
...
+
+Repos:
+  - MPN: https://mpn.metworx.com/snapshots/stable/2020-03-24
+  - MPN_Rcpp: https://mpn.metworx.com/snapshots/stable/2020-04-18
+  - MPN_withr: https://mpn.metworx.com/snapshots/stable/2020-05-09
+
+Customizations:
+  Repos:
+    - MPN:
+        Type: source
```
dpastoor, line 16: why are you forcing source on all of these?

dpastoor (changes requested):
can you expand a little on the logic around having the default/only pkgr built around the oldest version - especially given the oldest snapshot is the one that is "least important" from a day-to-day development perspective. Is the idea to not "accidentally" use new stuff since its not there in the first place? I'm not connecting the ultimate scenarios we want the developer to consider, how this will be expected to work day-to-day, and align with the CI environment

.drone.jsonnet
copernican: Exactly. We want the package set to be dictated by the container, not renv. renv is activated in `.Rprofile`, but because the renv library is not committed to git, renv won't find any of the package dependencies (which we don't want to use in any case), so `check()` will fail. So as @dpastoor says, we bypass activating renv by skipping `.Rprofile`.

pkgr.yml
copernican: This `Lockfile` statement tells pkgr to use an renv-style (project-specific) library.

pkgr.yml
copernican: Here's the thought process:

1. The default MPN snapshot should be the oldest compatible, precisely to avoid inadvertently relying on new features or backwards-incompatible behavior, as you say. 

2. The current oldest compatible snapshot is 2020-03-24.

3. We want to test against the latest R version for both the current and previous major.minor releases. At the moment, those are 4.0 and 3.6.

4. We test against Linux (Ubuntu) in CI, and we also want to test against macOS, especially given recent platform-specific issues.

5. (macOS) binaries for R 4.0 do not exist for 2020-03-24.

6. pkgr does not fall back to source if binaries are unavailable.

7. Therefore, to have a single `pkgr.yml` that works for both Linux and macOS development, we need to specify `Type: source` for each of the `Repos`.

You also commented that the oldest snapshot should be the "least important" for day-to-day development. Why do you think this? If the concern is that we want to mitigate the risk of (newer versions of) dependencies introducing breaking changes, that should be addressed by the `cran-latest` CI image.

So, the idea would be that development is done against the "oldest compatible" package versions, and if CI reveals a breaking change in a dependency, we will need to decide how to handle, whether by introducing/bumping a version constraint, adding a shim, etc. A similar process would play out if we wanted to take advantage of new, non-breaking functionality in a dependency. The general philosophy is conservative: take the minimal set of version constraints possible.

seth127:
I think we should add something to the README about how you need to use pkgr instead of `renv::restore()` or whatever if you want to develop on this package. Also, probably add a comment in the pkgr.yml about the weird parts that are necessary for CI.

seth127:
I'm good with having the snapshot be the oldest supported, especially because:

* it's relatively easy to change the snapshot date in pkgr.yml and develop against the newest version when you want to test things.
* we test against the newest snapshot and the tip of CRAN in CI, so any _breaking changes_ will get caught.

This also provides the easier dev flow because you will have changes much less frequently in your development environment. If we do it the other way, then you need to run `pkgr install --update` every time the newest snapshot changes and, more importantly, if you forgot to do that then you're actually developing against something older than you might think you are.

I hear the argument for developing against a newer version to see the new changes/features/patterns sooner, but I don't necessarily think package development is the right place to be doing that. Anything that gets deprecated, we will certainly find out (via CI, if not before) because it will begin throwing warnings or errors (at least all of tidyverse uses `lifecycle` for this), and likewise I don't think that we want to begin incorporating new functionality and patterns into package code until they have matured and stabilized a bit anyway.

copernican:
1. Per offline discussion, pkgr now points directly at CRAN.

2. As a consequence, the build is failing, due to both #286 and #287.

The implication is that we should add the fixes for those issues to this branch, so that the build will pass.

copernican:
All tests now pass. To reproduce, you can run `pkgr install --update` followed by `check()`.

dpastoor (approved):

PR #277, opened 2020-11-11
Docs/vignettes 0.10.0 edits

seth127:
Making edits to the vignettes to reflect some better practices now available in 0.10.0

This is _not_ a bugfix, just a vignette refactor to showcase newer patterns.

### DIFF
--- a/vignettes/using-based-on.Rmd
+++ b/vignettes/using-based-on.Rmd
@@ -25,11 +25,10 @@ If you are new to `rbabylon`, the ["Getting Started with rbabylon"](https://metr
 
 There is some initial set up necessary for using `rbabylon`. Please refer to the ["Getting Started"](https://metrumresearchgroup.github.io/rbabylon/articles/getting-started.html#setup) vignette, mentioned above, if you have not done this yet. Once this is done, load the library.
 
-```{r}
+```{r results = 'hide', message=FALSE, warning=FALSE}
 library(rbabylon)
-suppressPackageStartupMessages(library(dplyr))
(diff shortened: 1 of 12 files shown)


vignettes/using-based-on.Rmd
```diff
@@ -96,10 +95,10 @@ mod6 <- copy_model_from(mod5, 6)
 # ...submit mod6...look at diagnostics...decide you're done!
 ```
 
-Now that you have arrived at your final model, you can add a tag to it, which will be used shortly for filtering the `run_log()` tibble.
+Now that you have arrived at your final model, you can add description to identify it, which will be used shortly for filtering the `run_log()` tibble.
```
copernican, line 98: nit: do you mean "a description"?

vignettes/using-based-on.Rmd
```diff
@@ -138,14 +137,12 @@ log_df <- run_log(MODEL_DIR)
 log_df
 ```
 
-## Filtering tags example
-
-Among other things, the run log contains the tags that have been assigned to each model. Here we use `purrr::map_lgl` to filter the run log to only the final model.
+Among other things, the run log contains any descriptions that have been assigned to each model. Here we use `filter` and `pull` to get the path to the final model.
```
copernican, line 140: nit: consider using `filter()` rather than `filter` and `pull()` rather than `pull` to match usage elsewhere.

copernican (approved):
Two very minor comments, otherwise LGTM.

copernican (approved):
LGTM

PR #281, opened 2020-11-13
Fix/replace model field yaml bug

### LINKED ISSUE #280: replace_model_field does not modify YAML as it should
Any modifications to the model object should also be made in the accompanying YAML file. Currently that does not happen with any of the functions that call `replace_model_field()`.

This test (which I just added to test this)
```
  new_mod <- replace_tag(new_mod, ORIG_TAGS[1], NEW_TAGS[1])
  mod_yaml <- new_mod %>% get_yaml_path() %>% yaml::read_yaml()
  expect_identical(new_mod[[YAML_TAGS]], mod_yaml[[YAML_TAGS]])
```
fails with this
```
test-modify-model-field.R:77: failure: replace_tag() works correctly
new_mod[[YAML_TAGS]] not identical to mod_yaml[[YAML_TAGS]].
1/2 mismatches
x[1]: "new tag 1"
y[1]: "acop tag"
```

seth127:
closes #280

### DIFF
--- a/R/modify-model-field.R
+++ b/R/modify-model-field.R
@@ -85,6 +85,8 @@ replace_model_field <- function(.mod, .field, .old_val, .new_val) {
 
   .mod[[.field]][idx] <- .new_val
 
+  .mod <- save_model_yaml(.mod)
+
   return(.mod)
 }
 
--- a/tests/testthat/test-modify-model-field.R
+++ b/tests/testthat/test-modify-model-field.R
@@ -71,6 +71,10 @@ test_that("replace_tag() works correctly", {
   # test_replacing
   new_mod <- replace_tag(new_mod, ORIG_TAGS[1], NEW_TAGS[1])
   expect_identical(new_mod[[YAML_TAGS]], c(NEW_TAGS[1], ORIG_TAGS[2:length(ORIG_TAGS)]))
+
+  # check that YAML is modified
+  mod_yaml <- new_mod %>% get_yaml_path() %>% yaml::read_yaml()
+  expect_identical(new_mod[[YAML_TAGS]], mod_yaml[[YAML_TAGS]])
 })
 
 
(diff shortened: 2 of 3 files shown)


R/modify-model-field.R
```diff
@@ -85,6 +85,8 @@ replace_model_field <- function(.mod, .field, .old_val, .new_val) {
 
   .mod[[.field]][idx] <- .new_val
 
+  .mod <- save_model_yaml(.mod)
```
copernican, line 88: This change adds a side effect to `replace_model_field()`. This is consistent with the documentation for `modify_model_field()`

> After the object has been modified they will write the modified object back to the YAML

but worth commenting on because in general we should try to isolate side effects to as few functions as possible. This may not be achievable unless we change the abstraction so that the model is entirely in memory (versus portions in both memory and on disk).

tests/testthat/test-modify-model-field.R
```diff
@@ -71,6 +71,10 @@ test_that("replace_tag() works correctly", {
   # test_replacing
   new_mod <- replace_tag(new_mod, ORIG_TAGS[1], NEW_TAGS[1])
   expect_identical(new_mod[[YAML_TAGS]], c(NEW_TAGS[1], ORIG_TAGS[2:length(ORIG_TAGS)]))
+
+  # check that YAML is modified
```
copernican, line 75: Strictly speaking, this is a test of `replace_tag()`, not `replace_model_field()`. Because `replace_model_field()` is exported, it may be better to test it directly.

copernican (changes requested):
Let's talk about the test for this change.

R/modify-model-field.R
seth127: :eye_roll: You know I don't disagree, but you also know that ain't happening. (the "entirely in memory" comment)

copernican (approved):
LGTM

PR #282, opened 2020-11-13
Fix/add summary run col error

### LINKED ISSUE #279: add_summary errors if all summaries fail
If a user calls `run_log() %>% add_summary()` and _none_ of the summaries complete it will throw this very obtuse error
```
Error: Can't subset columns that don't exist.
x Column `run` doesn't exist.
Run `rlang::last_error()` to see where the error occurred.
In addition: Warning message:
In .impl_func(mod_list, ...) :
```

## situation

This will have in both of the following plausible situations

* the user has not correctly set `options('rbabylon.bbi_exe_path')` so model_summary() can't find it and all the summaries fail
* the user is running a batch of models and none of them has finished yet

In both cases, some kind of error _may_ be appropriate. But not the one currently being thrown.

## potential cause
I'm pretty sure this happens because, when no summaries complete, the `bbi_summary_log_df` comes back with only two columns: `absolute_model_path` and `error_msg`.  The most obvious way to fix this is to construct the `run` column even if everything fails, but there may be a better approach too.

seth127:
closes #279


R/config-log.R
```diff
@@ -86,7 +88,9 @@ config_log_impl <- function(.mods) {
 
   if (all(missing)) {
     warning(glue("Found no bbi_config.json files for {length(.mods)} models."), call. = FALSE)
-    return(tibble())
+    return(
+      c(ABS_MOD_PATH, RUN_ID_COL) %>% map_dfc(~ tibble(!!.x := character()))
```
copernican, line 92: I agree that the symptom is the same (the `run` column is missing from the object returned to `add_log_impl()`), but I'm not sure the solution should be the same. In particular, if all models are missing configuration files, there is not a separate error message to present to the user (as might be the case if all model summaries fail), so I'm not sure returning an (effectively) empty tibble gives the user information they wouldn't get by throwing an error instead.

R/summary-log.R
```diff
@@ -2,7 +2,8 @@
 #'
 #' Runs [model_summaries()] on all models in the input and returns a subset of the each resulting summary as a tibble.
 #'
-#' @return An object of class `bbi_summary_log_df`, which includes the fields described below.
+#' @return An object of class `bbi_summary_log_df`, which includes the fields described below. If _all_ model summaries
```
copernican, line 5: The returned object will also include the `run` column, which is not described in `@details`.

tests/testthat/test-config-log.R
```diff
@@ -197,3 +197,23 @@ test_that("config_log() works with missing output dirs", {
   expect_equal(ncol(log_df), CONFIG_COLS)
   expect_false(any(duplicated(log_df[[ABS_MOD_PATH]])))
 })
+
+test_that("config_log() works with no json found", {
```
copernican, line 201: I'm not sure we want to allow this behavior, so we may not need to test it.

copernican (approved):
Approving to unblock the merge. I do think we need to have a broader design discussion around the behavior we want to allow.

PR #294, opened 2020-12-04
Feat/process print method

### LINKED ISSUE #293: Print method for babylon_process object
# Summary

Implement a print method for `babylon_process()` objects. Not sure what this should be, but when someone calls `submit_model(...)` without assigning it to anything, it will print this object, which is currently a hideous list. We should print something nicer.

seth127:
closes #293


R/print.R
```diff
@@ -0,0 +1,52 @@
...
+  }
+  cli::cat_line(paste("In", run_dir), col = "green")
+
+  # format and print status string
+  if (length(x[[PROC_STDOUT]]) > 1) {
+    cli::cat_line("Process finished.", col = "green")
+  } else if (x[[PROC_STDOUT]] == "DRY_RUN") {
+    cli::cat_line("DRY RUN! Process not actually run.", col = "red")
+  } else if (str_detect(x[[PROC_STDOUT]], ".wait = FALSE")) {
+    cli::cat_line("Not waiting for process to finish.", col = "blue")
```
annacnev, line 48: What does `"Not waiting for process to finish."` mean? Done?

annacnev (approved):
looks good! no major changes requested- just some if you want to clean up code.

R/print.R
seth127: No, it means you passed `.wait = FALSE` so it just kicked off the process to run in the background. When this happens, we get no STDOUT.

tests/testthat/setup-workflow-ref.R
```diff
@@ -61,6 +61,8 @@ SUM_CLASS_LIST <- c(SUM_CLASS, "list")
 MOD_CLASS_LIST <- c(NM_MOD_CLASS, "list")
 PROC_CLASS_LIST <- c(PROC_CLASS, "list")
 
+PROC_HELP_STR <- c("Running:", as.character(glue("  {read_bbi_path()} --help")), as.character(glue("In {getwd()}")))
```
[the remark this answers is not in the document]
seth127: yea... but it actually returns some kind of `fs_[something]` object (which is also a string) so our `expect_equal` tests fail when you compare it to a regular string. Annoying, but this is the best workaround I've found. And this file is all refs for tests, so we gotta do it here (or in the tests, which would be even more annoying).

PR #297, opened 2020-12-17
Feat/summary getters

### LINKED ISSUE #285: Make get_*_path helpers work on summary object too
# Summary

As a user, I would like to be able to call the following methods on a `bbi_nonmem_summary` object:

* `get_model_id()`
* `get_model_path()`
* `get_output_dir()`
* `get_yaml_path()`
* `check_grd()`
* `check_ext()`

# Tests
- tests/testthat/test-get-path-from-object.R
  - get_model_path() builds the right path from summary object
  - get_output_dir() builds the right path from summary object
  - get_yaml_path() builds the right path from summary object
  - get_model_id parses summary object
- tests/testthat/test-read-output.R
  - check_ext() summary object
  - check_grd() summary object

seth127:
Closes #285


tests/testthat/test-rbabylon.R
```diff
@@ -30,56 +30,53 @@ test_that("bbi_dry_run() correctly returns object", {
   expect_identical(class(res), PROC_CLASS_LIST)
 })
 
-if (Sys.getenv("METWORX_VERSION") == "" && Sys.getenv("DRONE") != "true") {
-  skip("bbi_init only runs on Metworx or Drone")
-} else {
 
-  test_that("check_bbi_exe() correctly errors or finds paths", {
-    FAKE_BBI_PATH <- "/tmp/fake/bbi"
+skip_if_not_drone_or_metworx("bbi_init")
```
seth127, line 34: all of these "changes" below are just un-indenting because I took this out of an `else` block.

tests/testthat/test-param-labels.R
```diff
@@ -83,55 +83,52 @@ for (MODEL_PICK in MODEL_PICKS) {
...
-        expect_equal(nrow(join_df), nrow(.new_df))
-        expect_equal(join_df$value, join_df$estimate, tolerance = 0.01)
-        expect_equal(join_df$se, join_df$stderr, tolerance = 0.01)
-        expect_equal(join_df$c, join_df$random_effect_sd, tolerance = 0.01)
-        expect_equal(join_df$cse, join_df$random_effect_sdse, tolerance = 0.01)
-      })
-    }
-  })
-}
+withr::with_options(list(rbabylon.bbi_exe_path = read_bbi_path()), {
```
seth127, line 86: all of these "changes" below are just un-indenting because I took this out of an `else` block.

R/get-path-from-object.R
```diff
@@ -20,7 +33,13 @@ get_model_path <- function(.bbi_object, .check_exists = TRUE) {
 #' @rdname get_path_from_object
 #' @export
 get_model_path.bbi_nonmem_model <- function(.bbi_object, .check_exists = TRUE) {
-  find_nonmem_model_file_path(.bbi_object[[ABS_MOD_PATH]], .check_exists)
+  get_model_path_bbi(.bbi_object, .check_exists)
```
copernican, line 36: The methods for `bbi_nonmem_model` and `bbi_nonmem_summary` are identical for each of the `get_*` functions. Another approach to this problem could therefore be to create an abstract (parent) class from which both sub-classes inherit, and then define the method for the abstract class rather than writing the same method twice. The methods are all small enough that it's not adding a lot of code to write them twice, but it does complicate keeping the methods in sync if and when they change.

R/get-path-from-object.R
```diff
@@ -63,6 +83,76 @@ get_yaml_path <- function(.bbi_object, .check_exists = TRUE) {
...
+####################################################
+
+#' @keywords internal
+get_model_path_bbi <- function(.bbi_object, .check_exists = TRUE) {
+  find_nonmem_model_file_path(.bbi_object[[ABS_MOD_PATH]], .check_exists)
+}
+
+#' @keywords internal
+get_output_dir_bbi <- function(.bbi_object, .check_exists = TRUE) {
+.path <- .bbi_object[[ABS_MOD_PATH]]
```
copernican, line 145: style: indentation

R/model-summary.R
```diff
@@ -12,6 +12,9 @@
 #' **NONMEM**
 #'
 #' The returned list for a NONMEM model will contain the following top-level elements:
+#'  * **absolute_model_path** -- Absolute path to the model that generated this summary.
+#'    Note, using this directly is discouraged in favor of using the "getters"
+#'    described in `?get_path_from_object`.
```
copernican, line 17: Link to documentation rather than `?get_path_from_object`.

R/model-summary.R
```diff
@@ -137,10 +140,15 @@ nonmem_summary <- function(
     }
   )
 
-  res_list <- res$stdout %>%
+  res_list <- list()
+  res_list[[ABS_MOD_PATH]] <- tools::file_path_sans_ext(get_model_path(.mod))
```
copernican, line 144: Given that we already import rlang, this is a location where you might prefer

```r
res_list <- rlang::list2(
  !!ABS_MOD_PATH := tools::file_path_sans_ext(get_model_path(.mod))
)
```

R/read-output.R
```diff
@@ -222,9 +222,13 @@ check_grd.character <- function(.mod, .iter_floor = 0) {
 #' @describeIn check_nonmem_table_output Checks .grd file from a `bbi_nonmem_model`
 #' @export
 check_grd.bbi_nonmem_model <- function(.mod, .iter_floor = 0) {
-  grd_path <- build_path_from_mod_obj(.mod, "grd")
-  df <- check_nonmem_table_output(grd_path, .x_var = "ITERATION", .x_floor = .iter_floor)
-  return(df)
+  check_nonmem_table_bbi(.mod, .iter_floor, .extension = "grd")
```
copernican, line 225: These methods also support creating an abstract parent class.

tests/testthat/setup-workflow-ref.R
```diff
@@ -19,7 +19,13 @@ ABS_MODEL_DIR <- system.file("model", "nonmem", "basic",   package = "rbabylon")
 MOD_ID <- "1"
 MODEL_DIR <-   fs::path_rel(ABS_MODEL_DIR, getwd()) %>% as.character()
 MOD1_PATH <- file.path(MODEL_DIR, MOD_ID)
-MOD1 <- MOD1_PATH %>% read_model()
+MOD1 <- read_model(MOD1_PATH)
+# if we're on Metworx or Drone, run the summary
+if (!Sys.getenv("METWORX_VERSION") == "" || Sys.getenv("DRONE") == "true") {
```
copernican, line 24: Consider abstracting this logic into two functions, `is_metworx()` and `is_drone()`, and then calling them here and in the test helper.

copernican (changes requested):
A few minor comments, and a suggestion about an abstract class.

R/get-path-from-object.R
```diff
@@ -4,8 +4,21 @@
...
+#' **`get_model_path()`** returns the path the model definition file.
+#'   For NONMEM models, this is the control stream.
+#'
+#' **`get_output_dir()`** returns the path the directory containing
+#'   output files created when the model is run.
+#'
+#' **`get_yaml_path()`** returns the path the YAML file created by rbabylon.
+#'   This file contains metadata like tags, etc. and should, generally speaking,
+#'   _not_ be interacted with directly. The helper functions mentioned in
+#'   `?modify_model_field` modify this file for you.
```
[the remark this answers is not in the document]
seth127: that wasn't working for me. `check()` complained about it for some reason. I think maybe because it's not a real function.

R/model-summary.R
seth127: same as https://github.com/metrumresearchgroup/rbabylon/pull/297/files#r545953884

R/get-path-from-object.R
seth127: Per discussion offline, I think this is the right idea, but we are going to wait on it for now. The code as is works and will not be difficult to refactor whenever we cross this bridge. Adding a parent class seems like a good idea moving forward, but it will require some forward-thinking design that isn't necessary for this PR.

copernican (approved):
LGTM

PR #299, opened 2020-12-22
Feat/print bbi nonmem model

### LINKED ISSUE #292: Print method for bbi_nonmem_model object
# Summary

Implement a print method for `bbi_nonmem_summary()` objects. Doesn't need to be anything fancy, just something to make things more pleasant for scientists to interact with these objects. The `bbi_nonmem_model` object is just a named list with a lot of internal information and currently printing it is pretty ugly and intimidating.

# Proposed Specification

Print the following, into easy-to-read bullet-pointed subsections:

* The absolute model path, or possibly the absolute path to the model working directory
* The name of the associated yaml and model (control stream) files
* Whether the model has been run, maybe a trinary option?
  * Not run (no output dir)
  * Incomplete run (output dir but no bbi_config.json)
  * Finished running (output dir with bbi_config.json)
* `description` (if any)
* Bullet point list of `tags` (if any)

Could also consider the following, but I think probably unnecessary and will just make this crowded and hard to read:

* Comma separated print out of `based_on` field (if any)
* Bullet point list of `notes` (if any)
* Bullet point list of `bbi_args` (if any)

annacnev:
closes #292


seth127:
I think we should sit on this for now (as in, for this release) because there are some outstanding issues described in #300 .

Primarily this has to do with rendering in `.Rmd`, which I now realized was _not_ mentioned anywhere in the associated issue or spec. Oops. Still, I don't think we should merge this until we iron that out.

PR #300, opened 2020-12-23
DO NOT MERGE: ideas for fixing Rmd printing

seth127:
I will attach some screenshots in comments below. Basically, the way it is currently implemented in #299 the printing doesn't work well in Rmd docs. This is primarily because of `cli_h1` (and `cli_h2`) which seem to do something weird with stdout.


seth127:
The current implementation (as of 21c88ed506babc533480b0abb9ce52e9e405e742) prints nicely on the console as shown here:
 Screen Shot 2020-12-23 at 11 27 34 AM

Unfortunately, it does _not_ print nicely in a rendered `.Rmd` document, either with or without `results = 'asis'` (which we mention in the docs, because that's how you can get the table in `print.bbi_nonmem_summary` to print nicely).

Screen Shot 2020-12-23 at 11 26 50 AM

Screen Shot 2020-12-23 at 11 26 57 AM

seth127:
I tried doing some simple fixes (switching `cli_h1` to `cat_rule` and adding `"\n"` at the end of everything) and that made it a little better. But it's still not great, especially for `'asis'`. I think we need revisit this and decide on a best option that will compromise console vs Rmd printing.

Screen Shot 2020-12-23 at 11 48 25 AM

Screen Shot 2020-12-23 at 11 48 33 AM

seth127:
Abandoned this in favor of https://github.com/metrumresearchgroup/bbr/pull/307

PR #305, opened 2020-12-28
version constraint

dpastoor:
when using development version(s) of bbi, the version does not hold to standard semver, rather a date stamp, thus causing the version check to fail.

This suppresses it, and allows that suppression to be set via either option or env variable.

The convention, as this is really for development only, is to also add the prefix `DEV_` to the option name

thus, given a normal option of no_min_version, the choice is `DEV_no_min_version` to strongly communicate that most people should not be setting this.


R/zzz.R
```diff
@@ -1,5 +1,12 @@
 .onLoad <- function(libname, pkgname) {
 
+  # if the option set before load, keep it, else look for an env var
+  # if it isn't one of the false-y settings, then will set to true
+  if (is.null(getOption("rbabylon.DEV_no_min_version")) &&
+      !Sys.getenv("RBABYLON_DEV_NO_MIN_VERSION", unset = "FALSE") %in% c("0", "FALSE", "false")) {
```
seth127, line 6: Why do you want to just check the env var on load and set the option? Why not just check the env var in the function? 

Doing this with an environment variable makes sense to me, but I don't understand why we bring options into it. If we do it this way then you could set the env var to TRUE but then it doesn't take effect until you reload the package. Why is that desirable?

seth127 (changes requested):
This is a good idea, but I'm a little unclear on the implementation. See comment.

R/zzz.R
dpastoor: This way you can more easily manage it in CI only, by setting an environment variable, but can manage it interactively via options to change the behavior at any point in time. This keeps us still providing the same "interface" to users (set options), makes the option extra visible as a dev option, but also gives us a clean CI/global/system implementation via env vars.

This is actually the pattern that I'd like to move to for basically all options/pkg settings

seth127: Per discussion offline, I think if we're going to implement a system of configuration inheritance like this we should do it in a package so that it remains consistent. Adding this complexity already worries me, but adding it without enforcing consistency _really_ worries me.

I think we can go ahead and merge this and we'll link to it as we develop the new package so that we replace this code with the new package code once it exists. 

I'm not too worried about any breaking changes (to the `no_min_version` API) that could potentially happen, especially since this is only intended for developers.

seth127 (dismissed):

seth127:
Mmm, actually we need to make a change to skip this test if the option is set:
```
x |  15 2     | bbr exec functions [0.3 s]                                                                                                      
──────────────────────────────────
Failure (test-bbr.R:49:3): check_bbi_exe() errors on too low version
`check_bbi_exe(BBI_EXE_PATH)` did not throw an error.

Failure (test-bbr.R:49:3): check_bbi_exe() errors on too low version
CACHE_ENV$bbi_exe_paths[[BBI_EXE_PATH]] is not NULL

`actual` is a logical vector (TRUE)
`expected` is NULL
──────────────────────────────────
```

Currently it fails whenever you use this option and then run the full test suite. I can make that change and then push it.

seth127:
well, now our [drone build is failing](https://github-drone.metrumrg.com/metrumresearchgroup/rbabylon/1597/1/4) because of something with `renv`. I'll have to look at that later. [This](https://github.com/metrumresearchgroup/rbabylon/pull/305/commits/7c560e3d19c512cb80554c6f29497fdeef23cf7f) was the only code change I made, so I don't think that's what caused it...

seth127 (approved):
Great stuff. See comments above about the future of this pattern.

PR #307, opened 2021-01-13
ideas for fixing Rmd printing #2

annacnev:
Alternative idea to PR #300  to address the issues in #299

- detect if `knitr` is in progress, determine if the chunk is asis or markup & print a certain way depending on that. 
- if not `knitr`,  print using the original implementation in #299. 

Here are the outputs for cli, markup, & asis.

Screen Shot 2021-01-13 at 1 19 47 PM
Screen Shot 2021-01-13 at 1 19 57 PM
Screen Shot 2021-01-13 at 1 19 30 PM

**Note:** These outputs were made with `cli` version = 2.2.0 (latest on cran & installed by pkgr in the renv). When I built and installed this and ran the vignette outside the package environment, I had an old `cli` (2.0.2) where I had set my libpaths to & the markup output on this version is kinda wonky. I didn't want to add a version constraint in the DESCRIPTION file for `cli` because it still works on this older version, but it is just not as pretty. If there is a way we can recommend a version... we should do so.  

Screen Shot 2021-01-13 at 1 57 55 PM


seth127 (approved):
Looks good

PR #308, opened 2021-01-14
Stop checking in Renv folder

annacnev:
Adopting a new development workflow so we stop breaking the drone:

- renv/ now in the .gitignore file 
- .Rprofile set to source renv/activate.R only if it exists

To develop:
- clone repo
- open project in R session & run `renv::init()` (install renv into default lib path if not already installed)
- run `pkgr install` in terminal within project dir.


seth127 (approved):
Looks good to me. I pulled it and messed around with stuff in the `renv/` folder (including `activate.R`) just to make sure it wasn't still getting tracked and all is working as expected.

PR #310, opened 2021-01-18
release: initial bbr version

dpastoor:
This PR bumps to 0.12.0. My preference would be to 1.0.0 this lib after some final testing to make sure we have cohabitation and migration locked down, thus the version is somewhat arbitrary compared to prior version - given the rename it doesn't feel like a "patch" release - but also need to give it a round of testing before 1.0.0 to iron out any tweaks.

In addition, this PR removes the baked in docs folder, in preference for migrating to using a gh-pages branch as to not pollute the commit history with doc rebuilds


seth127 (approved):
LGTM

PR #318, opened 2021-01-29
DO NOT MERGE: cmdstan dev1

seth127:
This is a dev branch for trying out ideas for how to create and use the newly conceived `bbi_stan_model` object.

Most new functions are sitting in `R/cmdstan.R`, though they will likely be broken out into different files before a real PR is created.


seth127:
This was a dead end that was superseded by #358 (and then later superseded by `bbr.bayes` package).

PR #322, opened 2021-02-04
Feat/change bbi default path

### LINKED ISSUE #321: Change default bbi path
# Summary

As a user, I would like to have to `bbr` look for `bbi` in my `$PATH` by default. I would also like the `use_bbi()` installer function to first try to install to whatever path I have set in `options("bbr.bbi_exe_path")`, falling back to the `bbi` in my `$PATH` (accessed via `Sys.which("bbi")`) and then an OS-dependent default, in that order.

# Tests
- tests/testthat/test-use-bbi.R
  - use-bbi works on linux pulling from options
  - use-bbi works on linux with path specified

seth127:
Closes #321


dpastoor:
not tested on windows yet - will follow up on that as that is lower priority but should not block this release.

dpastoor (approved):
good to go - lets bump and 🚀

PR #332, opened 2021-02-22
Refactor/bbi model parent class

### LINKED ISSUE #331: bbi_model parent class
# Summary

We are going to add a `bbi_model` parent class to all objects that correspond to some kind of `bbr` model object. Specifically, right now these are `bbi_nonmem_model` and `bbi_nonmem_summary` objects, but will also include `bbi_stan_model` and `bbi_stan_summary` objects in the near future. 

Part of the point of this is in anticipation for the Stan models and wanting to have a parent class for attributes they will share in common. For example, having common dispatches that pull `"absolute_model_path"` from different kinds of models. 

The other part of this is cleaning up some complexity that has become unnecessary and will be more cumbersome as we add Stan (and other types of) models in the future.

# Spec

* Add `bbi_model` class to both `create_model_object()` and `create_summary_object()`
  * Remove `bbi_args` as a required key because it won't be required for Stan, and honestly there's no reason it should be required for NONMEM either.
* Refactor the following helpers to dispatch on `bbi_model`
  * `get_yaml_path()`
  * `get_model_id()`
  * `get_data_path()`
  * `build_path_from_model()`
  * `check_output_dir()`
* _Note:_ `get_model_path()` and `get_output_dir()` will _not_ dispatch on `bbi_model` because we already know those will be different for NONMEM and Stan models

seth127:
closes #331


R/get-path-from-object.R
```diff
@@ -352,35 +311,6 @@ yaml_ext <- function(.x) {
 ###################################
 
 
-#' Build absolute path from a directory and path
```
seth127, line 355: This is unrelated to the specific change, I just noticed that this isn't used anywhere and should've been deleted back when we did the `model_directory` refactor.

inst/model/nonmem/basic/1/1.cpu
```diff
@@ -1 +1 @@
-       3.141
+       2.595
```
seth127, line 1: most of these `inst` changes are just date/time stamps etc. but I reran the test models and refs so that the classes of the summary objects would be updated to include `bbi_model`

callistosp (approved):
LGTM. lots of overlap with the review I did on the [tags_diff PR](https://github.com/metrumresearchgroup/bbr/pull/337)

seth127:
> LGTM. lots of overlap with the review I did on the [tags_diff PR](https://github.com/metrumresearchgroup/bbr/pull/337)

Ahhh, that PR was branched off of this one but I bet the diff was still showing a lot of these changes (hence why you already saw them). Thanks!

PR #334, opened 2021-02-23
add NO_COLOR env var to print tests and refs

### LINKED ISSUE #333: print tests failing because of color codes
This isn't really a bug, it's just that our print tests are now failing, likely because one of our dependencies changed.

This issue is just to document.

# Solution
I'm pretty sure this is the issue https://github.com/r-lib/crayon/blob/master/NEWS.md#140 (they now "correctly" support ANSI colors that were previously being ignored)

In [this commit](https://github.com/r-lib/crayon/commit/07635d7b0bc1bbf0ea061623ecbe6ab563b2107c) they added support for ANSI colors (which the `cli` package uses in our `print` methods). Per [docs here](https://github.com/r-lib/crayon/blob/master/R/aab-num-ansi-colors.R#L19), it can be turned off.

See comment here where I got it to work: https://github.com/metrumresearchgroup/bbr/issues/333#issuecomment-784491858

seth127:
closes #333

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: bbr
 Title: R package for bbi
-Version: 1.0.0
+Version: 1.0.0.7000
 Authors@R: 
     c(person(given = "Devin",
              family = "Pastoor",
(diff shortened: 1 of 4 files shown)


copernican:
`check()` results in one note:

```
> checking R code for possible problems ... NOTE
  File ‘bbr/R/zzz.R’:
    .onLoad calls:
      packageStartupMessage("setting `rbabylon.DEV_no_min_version` to TRUE so no validation of bbi version will occur")
  
  See section ‘Good practice’ in '?.onAttach'.
```

In particular, such messages [should be emitted](https://r-pkgs.org/r.html#when-you-do-need-side-effects) by `.onAttach()` rather than `.onLoad()`. This is not related to this PR, but should be fixed. Do you want to add a fix here, or fix first with a separate PR?

copernican (approved):
LGTM

PR #335, opened 2021-02-24
Refactor/new model stan prep

seth127:
Some small refactoring of `new_model()` to prepare for adding Stan functionality.

* Simplified how we check if the model exists
* Made `.overwrite = TRUE` delete any old output directory as well the old .yaml (**this is the biggest change**)
* Changed how we fill the new model object from the arguments. 
  * We used to build the list manually and then create the model object
  * Now we create the bare bones model object first use the `replace_all_` helpers because these are more agnostic to model structure and have their own input checking built in
* Added a NONMEM-specific error message to `find_nonmem_models()` to warn Stan users who stumble onto that path


copernican:
This PR has the same issue as described [here](https://github.com/metrumresearchgroup/bbr/pull/334#issuecomment-786324076), and presumably the resolution will be the same in both cases.

R/classes.R
```diff
@@ -26,29 +26,23 @@ create_model_object <- function(res, save_yaml) {
...
-  if (is.null(res[[YAML_BBI_ARGS]])) {
-    res[[YAML_BBI_ARGS]] <- list()
-  } else {
-    # check that unique named list was passed
-    tryCatch(
-      checkmate::assert_list(res[[YAML_BBI_ARGS]], names="unique"),
-      error = function(e) { stop(glue("`{YAML_BBI_ARGS}` must be a unique, named list: {e}")) }
-    )
+  # look for appropriate model files on disk and then write out YAML.
+  # must be done AFTER assigning the class so that associated helpers can dispatch correctly
```
copernican, line 39: This comment is somewhat unclear, because on literally the next line we check `.model_type` rather than the class of `res`. I think this mostly relates to `get_yaml_path()`, which is called by `save_model_yaml()`?

R/get-path-from-object.R
```diff
@@ -37,13 +37,13 @@ get_model_path <- function(.bbi_object, .check_exists = TRUE) {
 #' @rdname get_path_from_object
 #' @export
 get_model_path.bbi_nonmem_model <- function(.bbi_object, .check_exists = TRUE) {
-  get_model_path_bbi(.bbi_object, .check_exists)
+  get_model_path_nonmem(.bbi_object, .check_exists)
```
copernican, line 40: That the implementations of this method, and the method for `bbi_nonmem_summary`, are identical, suggests that they share an abstract parent class, possibly `bbi_model`?

R/new-model.R
```diff
@@ -64,23 +57,18 @@ new_model <- function(
...
-    .mod[[YAML_BASED_ON]] <- safe_based_on(dirname(.path), .based_on)
-  }
-  if (!is.null(.tags)) .mod[[YAML_TAGS]] <- .tags
-  if (!is.null(.bbi_args)) .mod[[YAML_BBI_ARGS]] <- .bbi_args
-
-  # make list into S3 object
   .mod <- create_model_object(.mod, save_yaml = TRUE)
 
+  # update model from passed args
+  if (!is.null(.description)) .mod <- replace_description(.mod, .description)
```
copernican, line 67: Is the plan for these functions (and possibly `modify_model_field()`) to eventually become generics, with methods for NONMEM and possibly other models?

copernican (approved):
LGTM, just a few comments to consider.

R/new-model.R
seth127: I kind of doubt it because the idea of this refactor is that all of these will inherit from `bbi_model` and my current thinking is that all of these `modify_model_field()` helpers should probably _all_ work on any `bbi_model` because they will all have the same metadata structure (in terms of `tags`, `notes`, etc.).

As you note though, this does set us up to make those divergent methods like that if we want to, which is nice.

PR #338, opened 2021-02-26
Feat/check up to date

### LINKED ISSUE #339: check_up_to_date() function
# Summary

As a user, I would like a function to check whether the model file or data file associated with a model have changed since the model was run. I would also like to be able to pass a `bbi_log_df` tibble to check this same thing for all of the models it contains.

# Tests

- tests/testthat/test-check-up-to-date.R
  - check_up_to_date.bbi_nonmem_model() happy path
  - check_up_to_date.bbi_nonmem_model() with mismatched model
  - check_up_to_date.bbi_nonmem_model() with mismatched data
  - check_up_to_date.bbi_nonmem_model() with missing data
  - check_up_to_date.bbi_nonmem_model() with mismatched both
  - check_up_to_date.bbi_nonmem_summary() with mismatched model
  - check_up_to_date.bbi_log_df() works as expected

seth127:
closes #339

Also refactored `config_log()` to call the new function internally.


R/config-log.R
```diff
@@ -205,27 +187,6 @@ config_log_entry <- function(path,
   config[out_fields]
 }
 
-#' Compare a file to an MD5 sum
```
seth127, line 208: this moved to utils.R

tests/testthat/test-new-model.R
```diff
@@ -56,7 +56,7 @@ test_that("new_model() throws an error if the model file does not exist", {
 })
 
 test_that("compare read_model() and new_model() objects", {
-  temp_mod_path <- create_temp_model(YAML_TEST_FILE)
```
seth127, line 59: this is unrelated cleanup

R/check-up-to-date.R
```diff
@@ -0,0 +1,146 @@
...
+  changed_files <- c(
+    config[[CONFIG_MODEL_MD5]] != tools::md5sum(model_file),
+    config[[CONFIG_DATA_MD5]] != tools::md5sum(data_file)
+  )
+
+  any_changes <- any(changed_files)
+  if(isTRUE(any_changes)) {
+    message(paste(
+      glue("The following files have changed in {get_model_id(.mod)}"),
+      paste("*", names(which(changed_files)), collapse = "\n"),
```
annacnev, line 127: maybe this is just a me thing but it's kinda weird to mix `paste` and `glue`.. especially because `glue` has the same base functionality as `paste`. I.e. this could be rewritten:

```
message(glue(
glue("The following files have changed in {get_model_id(.mod)}"),
glue_collapse("*", names(which(changed_files)), sep = "\n"),
sep = "\n"
))
```

paste works so this isn't wrong I just don't see why we are using it if we have already imported the glue package

annacnev (approved):
All looks good- minor suggestions for code aesthetics but nothing really that would change functionality so completely optional.

R/check-up-to-date.R
seth127: yea, I see that. I always just use paste unless I explicitly need the `{}` glue thing. I would change it here but... I do this all over the place so consistency is really not within reach.

R/check-up-to-date.R
```diff
@@ -0,0 +1,146 @@
...
+#' calling the function for this side effect without explicitly handling the
+#' returned value.
+#'
+#' @export
+check_up_to_date <- function(.bbi_object, ...) {
+  UseMethod("check_up_to_date")
+}
+
+#' @export
+check_up_to_date.bbi_nonmem_model <- function(.bbi_object, ...) {
```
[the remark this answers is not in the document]
seth127: Part of the point of this is prepare for adding Stan stuff so, even though these are the same, they're _not_ gonna be the default methods. I considered adding another parent class that they would share, but that seemed a little excessive to have something like
```r
class(.mod)
# "bbi_nonmem_model"  "bbi_nonmem_object"  "bbi_model"  "list"
```
(`"bbi_model"` is already the parent class that they'll share with Stan and other kinds of models)

Anyway, that would be a totally valid way to do it, but I opted for just duplicating a couple dispatches instead (and made them use the same private helper to avoid copy/paste).

PR #344, opened 2021-03-04
fix: Bayes param_estimates() est_methods bug

### LINKED ISSUE #330: param_estimates() not catching Bayesian methods correctly
There is a bug [here](https://github.com/metrumresearchgroup/bbr/blob/master/R/param-estimates.R#L51) that models where the Bayesian estimation method is _not first_ are not picked up.

`str_detect(est_method, "Bayesian")` should be changed to `any(str_detect(est_method, "Bayesian"))`

seth127:
closes #330


R/param-estimates.R
```diff
@@ -43,17 +43,16 @@ param_estimates <- function(.summary) {
 #' @importFrom stringr str_detect
 #' @export
 param_estimates.bbi_nonmem_summary <- function(.summary) {
-  num_methods <- length(.summary[[SUMMARY_PARAM_DATA]])
```
seth127, line 46: there were two problems, only [one of which](https://github.com/metrumresearchgroup/bbr/issues/330#issue-813667615) was diagnosed in the original issue. 

The other is this line. Specifically for `METHOD=CHAIN` there is nothing in `.summary[[SUMMARY_PARAM_DATA]]` so `num_methods` was coming up with one method _less_ than it should've had. This was causing problems further down, namely that when we thought we were checking the last estimation method we were actually checking the _second to last_. The change on lines 55-56 is also to account for this.

inst/model/nonmem/complex/acop-fake-bayes.ctl
```diff
@@ -0,0 +1,41 @@
+$PROBLEM PK model 1 cmt base
```
seth127, line 1: these are all just the added test case

tests/testthat/test-summary-log.R
```diff
@@ -91,13 +91,13 @@ withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
     sum_df2 <- summary_log(MODEL_DIR_X, .fail_flags = list(ext_file = "1001.1.TXT"))
 
     # check fail flag parsed
-    expect_equal(sum_df2[[SL_FAIL_FLAGS]], c(TRUE, FALSE, FALSE, FALSE))
+    expect_equal(sum_df2[[SL_FAIL_FLAGS]], c(TRUE, FALSE, FALSE, FALSE, FALSE))
```
seth127, line 94: had to change all these because now our new test case gets pulled in here

timwaterhouse (approved):
Looks good, thanks.

PR #343, opened 2021-03-04
Add heuristics descriptions to model_summary() docs

seth127:
closes #326

PR #342, opened 2021-03-02
add model_diff() function

seth127:
closes #324

PR #337, opened 2021-02-26
add tags_diff()

seth127:
closes #325

PR #345, opened 2021-03-05
bbr 1.1.0 release

seth127:
# Release Notes
This release adds a few helper functions and makes a few small patches, but it is mostly internal refactoring in preparation for the impending `bbr_stan_alpha` branch where initial development to support Stan modeling will begin soon.

## New features and changes

* Added `tags_diff()` function for comparing the tags between different models. (#337)

* Added `model_diff()` function for comparing the model files between different models. (#342)

* Added `check_up_to_date()` function for checking whether the model file(s) and data file(s) associated with a model have changed since the model was run.  (#338)

* Added more documentation about the heuristics returned from `model_summary.bbi_nonmem_model()` (#343)

## Bug fixes

* `param_estimates()` now correctly errors when a Bayesian method is used but is _not_ the final method. (#344)

## Developer-facing changes

* Added a `bbi_model` parent class to `bbi_nonmem_model` and `bbi_nonmem_summary` objects. Many of the helpers in `get-path-from-object.R` now dispatch on this class. This had been discussed in the past but was primarily done now in preparation for beginning development for Stan modeling, which will create `bbi_stan_model` and `bbi_stan_summary` objects that will also inherit from this parent class. (#332)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: bbr
 Title: R package for bbi
-Version: 1.0.0.8100
+Version: 1.1.0
 Authors@R: 
     c(person(given = "Devin",
              family = "Pastoor",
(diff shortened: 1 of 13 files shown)


dpastoor (approved):
🚀

PR #361, opened 2021-03-10
feat/stan copy_model_from

### LINKED ISSUE #357: copy_model_from for Stan models
# Summary
As a user, I would like to be able to use `copy_model_from()` to create a new `bbi_stan_model` object from a parent model. The parent model should be required to be a "complete" (as determined by `check_stan_model()`) model.

# Tests
- tests/testthat/test-copy-model-from.R
  - copy_from_model.bbi_stan_model creates accurate copy

seth127:
Closes #357 

Also changed `.inherit_tags` to default to `TRUE`. We've talked about doing this in the past and this seems like a good time to do it because whenever this release eventually merges it will be a major one.


seth127:
Screen Shot 2021-03-10 at 4 08 18 PM

PR #374, opened 2021-03-29
bbi_exec() stdout and stderr to file

seth127:
There was a bug where submitting more than roughly 250 models at time via `submit_models()` (e.g. for bootstrapping) 
would hang indefinitely. This had something to do with [a bug in processx](https://github.com/r-lib/processx/issues/286).
It was fixed (in `bbr`) by routing the stdout and stderr to a temp file and then reading from it when necessary, instead 
of relying on `processx` to poll the process.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: bbr
 Title: R package for bbi
-Version: 1.1.0
+Version: 1.1.0.7000
 Authors@R: 
     c(person(given = "Devin",
              family = "Pastoor",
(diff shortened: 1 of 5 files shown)


seth127:
Intending to make a `1.1.1` release with this. Shouldn't effect anyone, except people trying to do big bootstrap runs.

`check()` passes on Metworx for the "oldest supported" and tip of CRAN. (I did this because it doesn't run all the NONMEM tests in Drone.)

**oldest**
Screen Shot 2021-03-29 at 3 56 40 PM

**tip**
Screen Shot 2021-03-29 at 4 07 44 PM

seth127:
@kylebaron @callistosp I'm trying to get this out as a patch release to drop into the example project. Lemme know if either of you want to jump on the line and do a live code review or anything.

callistosp (approved):
LGTM. tried submitting 300 test models, seemed to work fine

PR #377, opened 2021-04-09
Feat/bbi exe mode option

### LINKED ISSUE #364: Move .mode argument to .bbi_args
# Summary

As a user, I would like to be able to set the `bbi` execution mode (e.g. "local" vs "sge") in the `bbi.yaml` file, as a global default, and attach it or pass it through as `bbi_args` for models that need to override the global setting.

# Tests

### LINKED ISSUE #379: Set .mode with global options
# Summary
As a user, I want to be able globally set the `.mode` argument to `submit_model()` and `submit_models()` by using `options("bbr.bbi_exe_mode")`. This should be able to be overriden by passing something to the `.mode` argument directly.

# Tests

- tests/testthat/test-submit-model.R
  - submit_model(.mode) inherits option
  - submit_model(.mode) errors when NULL
  - submit_model(.mode) errors when invalid
- tests/testthat/test-submit-models.R
  - submit_models(.mode) inherits option

seth127:
Closes #379


R/zzz.R
```diff
@@ -11,6 +11,10 @@
     options("bbr.bbi_exe_path" = BBI_DEFAULT_PATH)
   }
 
+  if (is.null(getOption("bbr.bbi_exe_mode"))) {
+    options("bbr.bbi_exe_mode" = BBI_DEFAULT_MODE)
```
callistosp, line 15: when is `zzz.R` called? is this a safe assumption to make?

callistosp (approved):

R/zzz.R
seth127: it's called when you load the package (after the rest of the files in `R` are sourced). That's where we do the rest of these so I think it's right. It's basically "if this isn't already set (in `.Rprofile` or something) then set it to the default." Then you can overwrite that in your code later too.

callistosp: okay that seems fine, just wanted to verify

PR #384, opened 2021-05-18
Update for testthat 3e

### LINKED ISSUE #155: replace setup() and teardown() with 3e test fixtures
https://github.com/metrumresearchgroup/rbabylon/blob/921b91094afbe193ba71d2c85e19af3cac4b2e85/tests/testthat/test-config-log.R#L53-L56

---

###### This issue was generated by [todo](https://todo.jasonet.co) based on a `TODO` comment in 921b91094afbe193ba71d2c85e19af3cac4b2e85 when #148 was merged. cc @metrumresearchgroup.

### LINKED ISSUE #290: Remove hacks from all_equal bug
There was a bug in previous testthat versions (that is mentioned [here](https://www.tidyverse.org/blog/2020/10/testthat-3-0-0/#3rd-edition) as "a bug in testthat’s comparison of floating point values") that necessitated some manual rounding code. This should now be removed.

Look for any mention of "testthat 3e" in comments in 
* `test-read-output.R`

seth127:
Closes #155

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: bbr
 Title: R package for bbi
-Version: 1.1.2
+Version: 1.1.2.7000
 Authors@R: 
     c(person(given = "Devin",
              family = "Pastoor",
--- a/tests/testthat/test-collapse-to-string.R
+++ b/tests/testthat/test-collapse-to-string.R
@@ -1,11 +1,10 @@
 context("Collapse columns to string representation")
 
-setup({
-  cleanup()
-  create_rlg_models()
-})
-
-teardown({ cleanup() })
+# setup
+cleanup()
+create_rlg_models()
+# teardown
+withr::defer(cleanup())
 
 test_that("collapse_to_string() works correctly", {
   # add a note to collapse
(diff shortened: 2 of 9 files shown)


seth127:
Was gonna close #290 in here too (346cdd74110c1d4d96fc8b02a0bcc0683d6c093b) but it broke our ["oldest supported snapshot"](https://github-drone.metrumrg.com/metrumresearchgroup/bbr/2010/2/4) so I [rolled it back](346cdd74110c1d4d96fc8b02a0bcc0683d6c093b).

copernican:
`check()` results in one note:

> Imports includes 22 non-default packages. Importing from so many packages makes the package vulnerable to any of them becoming unavailable.  Move as many as possible to Suggests and use conditionally.

Is this something we might think about resolving? For example, forcats and ggplot2 are needed for only one function, `plot_nonmem_table_df()`. It seems reasonable to move them to `Suggests` and have the function check for them.

copernican (approved):
LGTM

seth127:
Thanks, the note about `check` and the dependencies is something I've been ignoring for awhile, but you're right that we should probably clean that up a bit. the `ggplot2` one is an obvious example.

Issue #387, opened 2021-06-24
handling lists in add_tags function


callistosp:
# Summary

As a user, I want to be able to pass a list or a vector to `add_tags()` and have the correctly formatted tags persisted to the model.

# Tests
- test/testthat/test-tags.R
  - some test (this will break mrgvalidate because this test doesn't exist)

seth127:
# Original Post

# Problem
When adding tags as a list instead of a character vector the `collapse_to_string` function doesn't work properly

## example code
```
> mod %<>% add_tags(list(TAGS$bio, TAGS$one_compartment_absorption))
> run_log(MODEL_DIR) %>% collapse_to_string(tags) %>% select(run, tags)
# A tibble: 5 x 2
  run   tags                                                                                                                      
  <chr> <chr>                                                                                                                     
1 100   "one-compartment + absorption, additive RUV, ETA-CL, ETA-KA, ETA-V"                                                       
2 101   "two-compartment + absorption, additive RUV, ETA-CL, ETA-KA, ETA-V2, CLWT-allo, V2WT-allo, V3WT-allo, QWT-allo"           
3 102   "list(\"one-compartment + absorption\", \"additive RUV\", \"ETA-CL\",     \"ETA-KA\", \"ETA-V\", \"CLWT-allo\", NULL)" 
```

# Proposed Solution
`unlist` all inputs to `add_tags` (and potentially other `add_*` functions which may be affected) when received.

PR #389, opened 2021-06-30
modify_model_field helpers: unlist and type check character vector values

kyleam:
This PR fixes gh-387 by implementing @callistosp's suggestion: call `unlist` on the `.tags` argument of `add_tags`.  Other `modify_model_field` helpers should get the same treatment, so this change is made at the level of `modify_model_field` (through an optional parameter).  In addition, there's now a check that these values (after unlisting) are character vectors.

### DIFF
--- a/R/modify-model-field.R
+++ b/R/modify-model-field.R
@@ -30,7 +30,7 @@
 #' @param .append If `TRUE`, the default, concatenate new values with currently present values. If `FALSE`, new values will overwrite old values.
 #' @param .remove If `TRUE`, `.value` with be removed from the `.field` instead of added. `FALSE` by default. Cannot have both `.append` and `.remove` be true in the same call.
 #' @param .unique If `TRUE`, the default, de-duplicate `.mod[[.field]]` after adding new values. If `FALSE` duplicate values will be kept.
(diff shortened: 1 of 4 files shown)


R/modify-model-field.R
```diff
@@ -30,8 +30,9 @@
 #' @param .append If `TRUE`, the default, concatenate new values with currently present values. If `FALSE`, new values will overwrite old values.
 #' @param .remove If `TRUE`, `.value` with be removed from the `.field` instead of added. `FALSE` by default. Cannot have both `.append` and `.remove` be true in the same call.
 #' @param .unique If `TRUE`, the default, de-duplicate `.mod[[.field]]` after adding new values. If `FALSE` duplicate values will be kept.
+#' @param .char_value If `TRUE`, check that `.value` (after unlisting) is a character vector.
 #' @export
-modify_model_field <- function(.mod, .field, .value, .append = TRUE, .remove = FALSE, .unique = TRUE) {
+modify_model_field <- function(.mod, .field, .value, .append = TRUE, .remove = FALSE, .unique = TRUE, .char_value = TRUE) {
```
kyleam, line 35: Happy to reflow this line, not sure what the project conventions are here.

seth127:
Closing this and moving to #393

PR #390, opened 2021-07-07
Fix/processx cleanup bug

seth127:
Fixes a bug where submitting multiple models in a loop with `submit_model(.mode = "local", .wait = FALSE)` would cause the models to never finish because the `bbi` processes would get [killed by `processx`](https://processx.r-lib.org/reference/process.html#cleaning-up-background-processes) when the R objects were garbage collected.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: bbr
 Title: R package for bbi
-Version: 1.1.2.7000
+Version: 1.1.2.7001
 Authors@R: 
     c(person(given = "Devin",
              family = "Pastoor",
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,11 @@
+# bbr develop
+
+## Bug fixes
+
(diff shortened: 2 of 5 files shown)


seth127:
Here's some ad-hoc test code. 

```r
mod_dir <- system.file("model", "nonmem", package = "bbr")
test_dir <- file.path(mod_dir, "test")

if (fs::dir_exists(test_dir)) fs::dir_delete(test_dir)
fs::dir_create(test_dir)
bbi_init(test_dir, "/opt/NONMEM", "nm75")

mod1 <- read_model(file.path(mod_dir, "basic", 1))

for (i in 1:20) {
  mod1 %>%
    copy_model_from(file.path("..", "test", i)) %>% 
    add_bbi_args(list(threads = 1, parallel= F)) %>%
    submit_model(.mode = "local", .wait = FALSE)
  Sys.sleep(1)
}

Sys.sleep(20)
.s <- summary_log(test_dir)
sum(is.na(.s$error_msg))
```
The final line tells you how many models successfully completed (should be all 20). On `develop` this was coming back with 15-17 typically. On this branch it should always come back 20.

I don't think we need to put this in the actual test suite because it's going to take awhile to run and it's a pretty rare edge case. If people disagree though, I can add something like this as a unit test.

dpastoor (approved):
excellent!

kyleam (approved):
Looks good to my not-yet-familiar-with-bbr eyes.

PR #391, opened 2021-07-08
1.1.3 release

seth127:
This is a minor bug fix release. In addition to what's below, we also updated some of the test code to use the more up-to-date testthat "fixtures" style of setup and teardown, and updated to the new bbr logo in the README. Full diff [here](https://github.com/metrumresearchgroup/bbr/compare/main...release/1.1.3).

# Release Notes

## Bug fixes

* Fixed a bug where submitting multiple models in a loop with `submit_model(.mode = "local", .wait = FALSE)` would cause the models to never finish because the `bbi` processes would get [killed by `processx`](https://processx.r-lib.org/reference/process.html#cleaning-up-background-processes) when the R objects were garbage collected. (#390)

* `bbr` now checks for a valid configuration file _before_ calling out to `bbi` to avoid the situation where `.wait = FALSE` and the "no config file" error from `bbi` is swallowed. Note, this check is skipped if `.dry_run = TRUE`. (#390)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: bbr
 Title: R package for bbi
-Version: 1.1.2.7001
+Version: 1.1.3
 Authors@R: 
     c(person(given = "Devin",
              family = "Pastoor",
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,10 +1,10 @@
-# bbr develop
+# bbr 1.1.3
 
 ## Bug fixes
 
 * Fixed a bug where submitting multiple models in a loop with `submit_model(.mode = "local", .wait = FALSE)` would cause the models to never finish because the `bbi` processes would get [killed by `processx`](https://processx.r-lib.org/reference/process.html#cleaning-up-background-processes) when the R objects were garbage collected. (#390)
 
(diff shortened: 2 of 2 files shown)


kyleam (approved):
I'm of course not familiar with the specifics of the bbr release process (what needs to be updated, cadence, etc.), but looks sensible to me.

dpastoor:
The final piece is we can just add a tag... (will complete this sentence shortly)

dpastoor (approved):

dpastoor:
ok so last part that I also did is then tag 1.1.3 - this will kick off the CD process: https://github-drone.metrumrg.com/metrumresearchgroup/bbr/2060/5/1

so now 1.1.3 is available as a cran-like repo at https://s3.amazonaws.com/mpn.metworx.dev/releases/bbr/1.1.3 for anyone that wants it now

so a projects pkgr.yml file might look like: 

```
Repos:
  - bbr: https://s3.amazonaws.com/mpn.metworx.dev/releases/bbr/1.1.3
  - MPN: https://mpn.metworx.com/snapshots/stable/2021-06-20
```

and will be pulled into the next MPN snapshot next weekend.

This way people can get completely stable snapshots that have been pretested as a cohort - or go out of bands "at their own risk" as needed.

PR #393, opened 2021-07-15
Fix/modify model unlist

### LINKED ISSUE #387: handling lists in add_tags function

seth127:
Closes #387 

Also note that this is identical to #389 that was closed because it was on a fork and therefore wasn't triggering the CI build.


seth127:
Merging this with my privileges because its actually changes @kyleam made in #389 (which I approve)

PR #407, opened 2021-09-27
Adding NONMEM Parallel tips and tricks vignette

seth127:
S

### DIFF
--- a/vignettes/nonmem-parallel.Rmd
+++ b/vignettes/nonmem-parallel.Rmd
@@ -1,0 +1,330 @@
+---
+title: "Running NONMEM in Parallel: bbr Tips and Tricks"
+author: "Seth Green"
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{nonmem-parallel}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+# 
+
+```{r setup, include=FALSE}
+knitr::opts_chunk$set(
+  echo = TRUE,
+  eval = FALSE
+)
+```
+
+## Introduction
+
+This file shows some helpful patterns for using `bbr` with NONMEM, mostly relating to running models with parallel execution, either multiple cores on a local machine (head node, laptop, etc.) or by executing on a grid.
+
+```{r message=FALSE}
+library(bbr)
+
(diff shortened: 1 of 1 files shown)


dpastoor:
do we have any prior art on how other packages handle this type of thing? eg things like tidymodels or keras or tensorflow where vignettes could take a potentially long time/need lots of data. This honestly is starting to feel like a separate docs site could give us the better ergonomics - a la what @kylebaron did for mrgsolve

seth127:
@dpastoor good thought. I did a little looking around this evening. Some thoughts are below. 

TL;DR is: my gut is to push it now as is, mostly because I'd like to be able to point people towards it while it's still fresh in their minds from the JET presentation, and then we can add some data and maybe move the whole thing later (and redirect to it) if we decide we want to. 

# Thoughts 

* This [move that keras does](https://github.com/keras-team/keras/tree/master/keras/datasets) is pretty slick: storing the data elsewhere out on the web, and then having functions to pull it in. Here's a simple example for a [large image dataset](https://github.com/keras-team/keras/blob/master/keras/datasets/fashion_mnist.py). Not sure it's worth it at this point, but I like this pattern. I know there are already some repos of public models and we could put some more up if necessary.

* For `mrgsolve` are you talking about [this](https://github.com/mrgsolve) (or more specifically [this](https://github.com/mrgsolve/mrgsolve.github.io)) or is there another site I missed? I don't think we're at the point of needing this yet, but good to keep in mind. `tidymodels` has a [similar thing](https://www.tidymodels.org/learn/).

* Most of the `tidymodels` vignettes I glanced through were just using `iris` and `mtcars` and other simple stuff that's already included and super quick to fit models on. 

* Similarly, the `keras` R package uses generated random data (e.g. [here](https://github.com/rstudio/keras/blob/master/vignettes/sequential_model.Rmd#L162) and [here](https://github.com/rstudio/keras/blob/master/vignettes/custom_models.Rmd#L92).

* We can also just put it up as a vignette for now and then redirect (like this [broom article](https://github.com/tidymodels/broom/blob/master/vignettes/bootstrapping.Rmd)) if we end up with enough material to merit its own separate docs repo later.

dpastoor:
cool deal - sounds good then - i'm fine to proceed forward as is - worst case we can adjust later.

dpastoor (approved):

PR #425, opened 2021-11-02
test: use same test ID for same print methods

seth127:


seth127 (commented):
I'm reviewing this myself because it's a tiny change, mostly just to make the validation docs slightly less obnoxious and verbose.

Issue #406, opened 2021-09-24
nm_join() (the artist formerly known as redataset() and superset())


seth127:
Returns a single data frame with model output (from specified table files) and input data.

Similar to the old `metrumrg::superset()` function, resurrected and redesigned by @kylebaron in 2020.

# Proposed code

```r
#' Return a vector of absolute paths to table files to look for
#' 
#' @param mod the `bbi_nonmem_model` object
#' 
#' @details
#' Called by redataset
#' 
tabfiles <- function(mod) {
  output_dir <- get_output_dir(mod)
  c(
    file.path(output_dir, "PAR_TAB"), 
    file.path(output_dir, "TAB"),
    build_path_from_model(mod, "par.tab"), 
    build_path_from_model(mod, ".tab")
    
  )
}


#' Return a single data frame with model output and input data
#' 
#' @param mod the `bbi_nonmem_model` object or a path to a NONMEM run
#' @param join character column name to use to join tab files
#' @param files absolute paths to table files to try to join
#' @param more absolute paths to other files to try to join
#' @param verbose if TRUE, messages are printed as files are read
#' 
#' @details
#' Required packges: data.table, dplyr, bbr. 
#' 
#' Note that you can get the absolute path to the model output directory 
#' with `bbr::get_output_dir(mod)` if needed to build paths for passing to 
#' `files` or `more`.
#' 
redataset <- function(mod, join = "NUM", files = tabfiles(mod), more = NULL, 
                      verbose = TRUE) {
  stopifnot(requireNamespace("data.table"))
  stopifnot(requireNamespace("dplyr"))
  stopifnot(requireNamespace("bbr"))
  if (!inherits(mod, "bbi_nonmem_model")) {
    mod <- read_model(mod)
  }
  files <- c(files,more)
  verbose <- isTRUE(verbose)
  sum <- bbr::model_summary(mod)
  nid <-  sum$run_details$number_of_subjects
  nrec <- sum$run_details$number_of_data_records
  data_loc <- get_data_path(sum)
  if(verbose) message("data file: ", basename(data_loc))
  
  data <- data.table::fread(data_loc, na = '.', data.table = FALSE)
  data <- dplyr::as_tibble(data)
  if(verbose) {
    message("  rows: ", nrow(data))
    message("  cols: ", ncol(data))
  }
  chk <- file.exists(files)
  files <- files[chk]
  if(length(files)==0) {
    message("  zero table files found; returning")
    return(data)
  }
[...]

PR #426, opened 2021-11-03
Feat/nm file

seth127:
Refactoring some of the helpers for loading NONMEM files and deprecating some old and soon-to-be-replaced functionality in `read-output.R`. Mostly this is prep for changes that will be made in implementing #406


seth127:
(Optionally) rename columns in `nm_grd()`

* theta, sigma, omega
*

seth127:
possible changes

* remove `TAB` and `PAR_TAB`
* Uppsala table naming method

seth127:
Scientific/user review already provided by @callistosp @kylebaron

R/utils.R
```diff
@@ -358,7 +369,10 @@ check_required_keys <- function(.list, .req) {
 #' @keywords internal
 check_model_object <- function(.mod, .mod_types = VALID_MOD_CLASSES) {
   if (!inherits(.mod, .mod_types)) {
-    stop(glue("Must pass a model object, but got object of class: `{paste(class(.mod), collapse = ', ')}`"))
+    stop(paste(
+      glue("Must pass a model object with one of the following classes: `{paste(.mod_types, collapse = ', ')}`"),
+      glue("Got object of class: `{paste(class(.mod), collapse = ', ')}`")
```
kyleam, line 374: There should be a newline or at least a period before "Got object ...":

```R
> check_model_object("foo")
Error in check_model_object("foo") :
  Must pass a model object with one of the following classes: `bbi_nonmem_model, bbi_stan_model` Got object of class: `character`
```

R/nm-file.R
```diff
@@ -0,0 +1,136 @@
...
+#' @param .mod Either a `bbi_nonmem_model`, `bbi_nonmem_summary`, or a path to a
+#'   file to read in. If passing model object to `nm_file()`, must also pass `.suffix` that
+#'   will be passed through to [build_path_from_model()].
+#' @param .est_method If `NULL`, the default, pulls the data from the final
+#'   estimation method. If an integer, pulls the data from that estimation
+#'   method. Note that, if more than one table is written to a given file, this
+#'   will also mean it pulls the final table by default, and this argument can
+#'   be used to pull a different table instead.
+#' @inheritParams build_path_from_model
+#' @param ... arguments passed through to methods. (Currently none.)
```
kyleam, line 17: What's the advantage of adding `...` now rather than waiting to see if a need arises? I haven't been able to come up with why adding it later would be a breaking change.

R/nm-file.R
```diff
@@ -0,0 +1,136 @@
...
+#'   `bbi_nonmem_summary` object
+#' @param .sep Single character used to separate fields within a record. Passed
+#'   through to `readr::read_delim()`. Defaults to `","`.
+#' @importFrom readr read_delim cols
+#' @export
+nm_data <- function(.mod, .sep = ",") {
+  check_model_object(.mod, c(NM_MOD_CLASS, NM_SUM_CLASS))
+  .path <- get_data_path(.mod)
+  verbose_msg(glue("Reading {.path}"))
+  read_delim(.path, delim =.sep, na = ".", col_types = readr::cols())
```
kyleam, line 100: `cols` is in the `importFrom` line.

NEWS.md
```diff
@@ -12,6 +12,18 @@ This release adds a number of helper functions, primarily for use with NONMEM mo
...
+
+* Deprecated `check_nonmem_table_output()`, `check_grd()`, and `check_ext()` and
+replaced them with `nm_file()` variants. These functions will warn about this
+for two more releases and then begin to error for two more releases before being
+removed altogether. (#426)
+
+* Deprecated `plot_nonmem_table_df()`, `plot_grd()`, and `plot_ext()` in an
+effort to more tightly define the scope of `bbr`. These functions will warn
+about this for two more releases and then begin to error for two more releases
+before being removed altogether. (#426)
```
kyleam, line 25: I'm not familiar with what the policy/practice is for bbr, but I was expecting this to say that these would be removed in a major release step (i.e. SemVer X.y.z bump).

R/nm-file.R
```diff
@@ -0,0 +1,136 @@
...
+
+#####################################
+# PRIVATE HELPERS AND IMPLEMENTATION
+
+#' Implementation function for reading NONMEM files
+#' @importFrom readr read_table read_table2 cols read_lines
+#' @importFrom stringr str_detect
+#' @keywords internal
+nm_file_impl <- function(.path, .est_method) {
+  checkmate::assert_integerish(.est_method, lower = 1, null.ok = TRUE)
```
kyleam, line 112: Isn't it safe to drop this `assert_integerish` call given the one below?

R/nm-file.R
```diff
@@ -0,0 +1,136 @@
...
+  checkmate::assert_string(.mod)
+  if (!is.null(.suffix)) {
+    stop("Cannot pass .suffix to use nm_file.character; pass only file path to .mod")
+  }
+  nm_file_impl(.mod, .est_method)
+}
+
+#' @describeIn nm_file Reads `.grd` file from a `bbi_nonmem_model` or
+#'   `bbi_nonmem_summary` object
+#' @param .rename If `TRUE`, the default, will rename `.grd` columns to the
```
kyleam, line 44: a final doc refresh is missing for this

NEWS.md
seth127: This is what we've done in the past, based on `tidyverse` practices. In practice, I don't think these functions are getting much use, so I'd be surprised if many (or any) users even notice.

R/nm-file.R
seth127: I don't think it's a big deal either way. I always do it by habit because it means, when you extend the method to other classes later, you don't have to change the signature of the original generic (other existing methods). This is obviously more important for methods that are likely to be extended in other packages, but it felt like a good practice to follow. Based on [this section](https://adv-r.hadley.nz/s3.html) of the Hadley book:

Screen Shot 2021-11-08 at 8 57 11 AM

NEWS.md
kyleam: > This is what we've done in the past, based on `tidyverse` practices.

Okay, thanks for the context.  I'll keep that in mind (even if I think it's worth trying to adhere closer to semantic versioning rules/expectations than tidyverse practices).

kyleam (approved):

PR #427, opened 2021-11-04
Fix/fix refresh scripts

seth127:
Updating test references, mostly in prep for some new helper functions that will necessitate modifying the test refs:

* write reference values to json instead of hard-coding constants in test files
* modifying test model to write table outputs

Sidenote: there was an issue with the `print.*` tests that I wanted to fix here as well, but now I can't reproduce it. I think it's worth merging this though, and then readdressing those issues if they crop back up.


seth127:
A lot of changes here that are incidental to refreshing the ref files. Mainly take a look at how I'm doing capturing this information in [here](https://github.com/metrumresearchgroup/bbr/pull/427/files#diff-75063d2219452bcc98201b03a8b47b8b2ed1dfa5f229a87887460187917cfab9) and then loading it [here](https://github.com/metrumresearchgroup/bbr/pull/427/files#diff-4576526f8ee3073a248a942c5d927309d713a5c4e1cdbf660e7bc4ad6225d310) for the tests.

I'll also note up top that I feel a little silly about how I load that json and then reassign all the pieces to constants, but going through all the tests that use those constants and putting `ref_json$` in front of them just felt a little too annoying to bite off. I can do it though if someone feels strongly about it.

inst/model/nonmem/basic/1.ctl
```diff
@@ -1,7 +1,8 @@
 $PROBLEM PK model 1 cmt base
 
-$INPUT ID TIME MDV EVID DV AMT  SEX WT ETN
+$INPUT ID TIME MDV EVID DV AMT  SEX WT ETN NUM
 $DATA ../../../../extdata/acop.csv IGNORE=@
+IGNORE(ID.EQ.2)
```
kyleam, line 5: Hmm, so presumably the motivation for these changes will be clear once the test for the new helpers arrive?

kyleam (commented):
The approach looks okay to me.

kyleam (approved):

inst/model/nonmem/basic/1.ctl
seth127: heh, yea that's right. This is ignoring one of the subjects in the trial data, essentially so that we can test this functionality: https://github.com/metrumresearchgroup/bbr/blob/bfb6eb2df69ae7bc2e250aee471b748bd0680695/tests/testthat/test-nm-join.R#L16-L37

PR #422, opened 2021-10-28
DO NOT MERGE: Feat/redataset

seth127:
Moving this functionality to a fresh branch for cleanliness. This turned into a dev/experimentation branch that ended up spawning #426 and #429. 

# Original description

Closes #406 

Manual testing can be done with the following:

```r
# clone the repo and checkout `feat/redataset` branch
# make sure you have bbi installed and `options('bbr.bbi_exe_path')` points to it
devtools::load_all()
setwd("tests/testthat")
source("setup-workflow-ref.R")
?nm_join
df <- nm_join(MOD1)
```

Also check out:

* the [`nm_join() code](https://github.com/metrumresearchgroup/bbr/blob/feat/redataset/R/nm-join.R)
* any of the tests in [`tests/testthat/test-nm-join.R`](https://github.com/metrumresearchgroup/bbr/blob/feat/redataset/tests/testthat/test-nm-join.R)
* the control stream and output files in [`inst/model/nonmem/basic/1`](https://github.com/metrumresearchgroup/bbr/tree/feat/redataset/inst/model/nonmem/basic)

PR #429, opened 2021-11-05
Feat/nm tables

seth127:
Building helpers in advance of refactoring #422 (originally intended to address #406)


R/nm-tables.R
```diff
@@ -0,0 +1,131 @@
...
+  # read in input data
+  res <- list(
+    data = nm_data(.mod)
+  )
+
+  # build names for table elements
+  .n <- .files %>%
+    basename() %>%
+    str_replace(glue("^{get_model_id(.mod)}"), "") %>%
+    str_replace(glue("{get_model_id(.mod)}$"), "") %>%
```
kyleam, line 43: * I think the model ID return value is arbitrary enough input that it should be protected with `\\Q...\\E`.  

 * I wasn't able to figure out the scenario where the `id$` case applies.  That case doesn't seem to be covered by the tests because those pass when I remove line 43.

 * Assuming that second call is needed: mostly a matter of taste that you can ignore if you don't agree, but I think it'd be good to lift the two `get_model_id` calls into one upstream one stored in a variable.

R/nm-tables.R
```diff
@@ -0,0 +1,131 @@
...
+    make.names() %>%
+    str_replace(glue("^\\Q.\\E"), "") %>%
```
kyleam, line 45: QE is handy for arbitrary input, but to my eyes regular escaping is nicer for a hard-coded period.

R/nm-tables.R
```diff
@@ -0,0 +1,131 @@
...
+nm_table_files <- function(.mod, .check_exists = TRUE) {
+  .p <- get_model_path(.mod)
+  .l <- parse_ctl_to_list(.p)
+
+  # get file names from table statements and construct paths
+  .f <- .l[names(.l) == "TABLE"] %>%
+    map_chr(~paste(.x, collapse = " ")) %>%
+    str_extract("\\bFILE\\s*=\\s*([^ ]+)") %>%
+    str_replace("\\bFILE\\s*=\\s*", "") %>%
+    str_replace("^\\Q./\\E", "") %>%
```
kyleam, line 70: I don't love the nearly repeated regular expression, but from a readability standpoint I can also see why you might prefer that to this:

```diff
diff --git a/R/nm-tables.R b/R/nm-tables.R
index d16cbba..f44d77a 100644
--- a/R/nm-tables.R
+++ b/R/nm-tables.R
@@ -65,9 +65,8 @@ nm_table_files <- function(.mod, .check_exists = TRUE) {
   # get file names from table statements and construct paths
   .f <- .l[names(.l) == "TABLE"] %>%
     map_chr(~paste(.x, collapse = " ")) %>%
-    str_extract("\\bFILE\\s*=\\s*([^ ]+)") %>%
-    str_replace("\\bFILE\\s*=\\s*", "") %>%
-    str_replace("^\\Q./\\E", "") %>%
+    str_match("\\bFILE\\s*=\\s*(?:\\./)?([^ ]+)") %>%
+    `[`(, 2) %>%
     file.path(get_output_dir(.mod, .check_exists = .check_exists), .)

   if(isTRUE(.check_exists)) {
```

Same comment as above about literal periods and QE.

R/nm-tables.R
```diff
@@ -0,0 +1,131 @@
...
+    file.path(get_output_dir(.mod, .check_exists = .check_exists), .)
```
kyleam, line 71: Similar subjective comment as above: I'd find it more readable to have  the `get_output_dir` call out of the loop since it doesn't change.

tests/testthat/test-nm-tables.R
```diff
@@ -0,0 +1,40 @@
...
+
+test_that("nm_tables() works", {
+  res <- nm_tables(MOD1)
+  expect_equal(length(res), length(MOD1_TABLE_FILES) + 1)
+  expect_equal(res$data, nm_data(MOD1))
+  expect_equal(res$tab, nm_tab(MOD1))
+  expect_equal(res$par.tab, nm_par_tab(MOD1))
+})
+
+test_that("nm_tables() works with .files argument [BBR-NMT-001]", {
```
kyleam, line 12: You might be planning to add test IDs later, but seeing this one made me think that perhaps you forgot to add them to the other tests in this file.

kyleam (commented):
Nice.

Left one comment about arbitrary input that should be protected in regex.  Other comments are subjective; feel free to ignore.

R/nm-tables.R
seth127: yea, I felt the same way as you, but this syntax does scare me a bit, from a readability/maintainability perspective, so I think I'll keep it. The repetition looks a bit dumb, but it does make it very clear what's happening.

R/nm-tables.R
seth127: the second one is necessary, but a good point that we never test it. I'll add one.

tests/testthat/test-nm-tables.R
```diff
@@ -0,0 +1,61 @@
...
+test_that("nm_tables() naming works correctly [BBR-NMT-002]", {
+  # construct fake tables with test case names
+  table_lines <- c(
+    "TABLE NO. 1",
+    "NUM VAR",
+    "1 2"
+  )
+  table_names <- file.path(
+    get_output_dir(MOD1),
+    c("satab1", "catab.1", "par-tab")
```
kyleam, line 28: > "satab1", "catab.1"

Are these motivated by actual naming conventions/practice?  Glancing bbr-nonmem-poppk-foce, I didn't spot anything following this pattern.

kyleam (approved):

tests/testthat/test-nm-tables.R
seth127: Yes. Example in [here](https://uupharmacometrics.github.io/xpose4/reference/read.nm.tables.html) where the show the roots, but @callistosp said they often also append the "run number" (`model_id` for us) on the _end_ instead of the beginning.

seth127: (to your point, not a convention we use at Metrum but Sam says it's pretty wide-spread.)

PR #430, opened 2021-11-08
feat: nm_join()

### LINKED ISSUE #406: nm_join() (the artist formerly known as redataset() and superset())

seth127:
Closes #406


seth127:
The moment we've all been waiting for! (At least since #406, #426, #429, etc...) 

There are a lot of changes in this branch, but most of the action happens in [`R/nm-join.R`](https://github.com/metrumresearchgroup/bbr/pull/430/files#diff-64ba2b4a8cc0d6d95419076e518bb0ce9c250b59e88a26403bd2113977a0624a) and [its tests](https://github.com/metrumresearchgroup/bbr/pull/430/files#diff-9c25528f115c57dc1dbda0f719c3867ec7d2c90caf28efcde7cf0bed530e92bf)

A brief summary of the other changes:

* All the changes in [`inst/model/nonmem/basic/...`](https://github.com/metrumresearchgroup/bbr/pull/430/files#diff-f40b8be75a1b298222359b55b5b1b563634e40e3d45ae721d95befbd26d7c8e7) involve adding some new table outputs to our test model and re-running it to generate the tables. All other changes in there (and [`inst/test-refs`](https://github.com/metrumresearchgroup/bbr/pull/430/files#diff-88381ad425f3867a5722eff55b1a60d8a6db166417bb5efb6007dd42ec9aa540)) are incidental annoying NONMEM stuff like runtime and datestamp.
* `nm_tables()` now takes a file path, to be consistent with the desired behavior (from scientists) for `nm_join()`
* `nm_file()` now makes all column names uppercase for consistency with downstream functions which expect that
* Some docs cleanup in related functions
* A few small bug fixes in related functions (e.g. [this one](https://github.com/metrumresearchgroup/bbr/pull/430/files#diff-9f6054e9c6b119e1ed6758eeae15716510e28be5b4f670ae7f8772e3e5254203R126-R128)) which were uncovered in testing

kylebaron:
## Parsing

Current project. 

### Using `readr`

```r
> .libPaths("/data/Rlibs")
> library(bbr)
> source("functions-model.R")
> options(bbr.verbose = FALSE)
> 
> x <- nm_join("model/pk/508")
Warning: 6056 parsing failures.
  row      col           expected               actual                                                 file
37678 DATETIME 1/0/T/F/TRUE/FALSE 2006-05-24T07:00:00Z '/data/home/kyleb/bbr-test/data/derived/tran103.csv'
37679 DATETIME 1/0/T/F/TRUE/FALSE 2006-05-24T07:30:00Z '/data/home/kyleb/bbr-test/data/derived/tran103.csv'
37680 DATETIME 1/0/T/F/TRUE/FALSE 2006-05-24T08:00:00Z '/data/home/kyleb/bbr-test/data/derived/tran103.csv'
37681 DATETIME 1/0/T/F/TRUE/FALSE 2006-05-24T08:30:00Z '/data/home/kyleb/bbr-test/data/derived/tran103.csv'
37682 DATETIME 1/0/T/F/TRUE/FALSE 2006-05-24T08:32:00Z '/data/home/kyleb/bbr-test/data/derived/tran103.csv'
..... ........ .................. .................... ....................................................
See problems(...) for more details.
```

### Using `data.table`
```r
> y <- redataset("model/pk/508", verbose = FALSE)
```

kylebaron:
# Benchmark

We can improve performance by 
1. Just read the file; opt into this scan for multiple tables
2. Switch to data.table

```r
Unit: milliseconds
                                            expr      min       lq     mean   median       uq       max neval cld
 x <- redataset("model/pk/508", verbose = FALSE) 110.1276 112.4106 173.0565 117.3880 140.3944  515.1812    30  a 
  y <- suppressWarnings(nm_join("model/pk/508")) 791.8433 806.6572 832.8808 820.2619 832.1596 1196.6709    30   b
```

seth127:
@kylebaron good points. Some requests:

* Could you provide some samples of difficult-to-parse tables? We're just [reading the lines as raw text](https://github.com/metrumresearchgroup/bbr/blob/feat/nm_join/R/nm-file.R#L121-L163) and then parsing tables manually, so we should definitely be able to handle edge cases where necessary.
* Can you give me some orders of magnitude (number of columns and rows) for what these tables might realistically be, so I can do some benchmarking? At this point we still don't have `data.table` anywhere, but we put it in `Suggests` and use it optionally if it's installed.
  * The problem with "Just read the file; opt into this scan for multiple tables" is that, if you try to "just read" a file with multiple tables it will fail. We could catch the error, but then I'm not sure what we do with that... Skip the table?  Automatically go back and try it the other way?
* In terms of multiple tables in a single file, the mechanics are already there (via `nm_file(.est_method)`) if we want to do that, but I thought the decision was to fail on that scenario. That said:
  * If you think we should do something else we can. Default behavior for non-table files is currently to read the final table in the file and let you pass an int to choose a different one. I [short-circuited that](https://github.com/metrumresearchgroup/bbr/blob/feat/nm_join/R/nm-tables.R#L62-L65) in `nm_tables.R` but we could change that. 
  * I'm not sure how we'd have an elegant interface for telling it "pick this table from this file and this other table from this other file" though. That feels kinda like an edge case we might just not support, but I'm open to feedback if you think it would be needed often enough.

kylebaron:
The test I did

```r
> x <- nm_tables("model/pk/508")
Warning: 6056 parsing failures.
  row      col           expected               actual                                                 file
37678 DATETIME 1/0/T/F/TRUE/FALSE 2006-05-24T07:00:00Z '/data/home/kyleb/bbr-test/data/derived/tran103.csv'
37679 DATETIME 1/0/T/F/TRUE/FALSE 2006-05-24T07:30:00Z '/data/home/kyleb/bbr-test/data/derived/tran103.csv'
37680 DATETIME 1/0/T/F/TRUE/FALSE 2006-05-24T08:00:00Z '/data/home/kyleb/bbr-test/data/derived/tran103.csv'
37681 DATETIME 1/0/T/F/TRUE/FALSE 2006-05-24T08:30:00Z '/data/home/kyleb/bbr-test/data/derived/tran103.csv'
37682 DATETIME 1/0/T/F/TRUE/FALSE 2006-05-24T08:32:00Z '/data/home/kyleb/bbr-test/data/derived/tran103.csv'
..... ........ .................. .................... ....................................................
See problems(...) for more details.

> lapply(x, dim)
$data
[1] 44069    44

$tab
[1] 37685     8

$par.tab
[1] 37685    15
```

R/nm-join.R
```diff
@@ -0,0 +1,179 @@
...
+#'
+#' **Join column**
+#'
+#' The `.join_col` is the name of a single column that should appear in both the
+#' input data set and any tables you want to join. We recommend you make this
+#' column a simple integer numbering the rows in the input data set (for example
+#' `NUM`). When this column is carried into the output table files, there will
+#' be unambiguous matching from the table file back to the input data set.
+#'
+#' The one exception to this are `$FIRSTONLY` tables. If a table file has the
```
timwaterhouse, line 39: Replace all `$FIRSTONLY` with `FIRSTONLY`

R/nm-join.R
```diff
@@ -0,0 +1,179 @@
...
+#' same number of rows as the there are individuals in the input data set, it
```
timwaterhouse, line 40: "same number of rows as the there are individuals in the input data set (accounting for any filtering of data in the NONMEM control stream)."

dpastoor:
putting in my voice after a brief convo with seth on slack - IMO data.table would be 👍  

I think the dep is such a small and contained dependency that even if these functions might not be "needed" but would not worry about jumping thorugh potential suggested hoops/checks.

R/nm-join.R
kylebaron: That was probably my doing. 👎

seth127: nah it was me. I'll change.

seth127:
Alright, I'm on board for `data.table` if both of you are.

My next big question though: how do we handle multiple tables in a single file. The biggest concern here is `.ext` files, but it's certainly possible for other files too. 

I like the behavior we describe [here](https://github.com/metrumresearchgroup/bbr/blob/feat/nm_join/R/nm-file.R#L14-L19) (possibly not the `fail` part, but the other part), but [the implementation](https://github.com/metrumresearchgroup/bbr/blob/feat/nm_join/R/nm-file.R#L121-L163) is a bit messy. I'll have to look at how to best do this with `data.table`, but any ideas are welcome.

We need to settle on a spec though. Proposal:

* `nm_file()` (and `nm_grd()` and `nm_ext()`, which call it internally) will default to final method, but let you pick a different one
* `nm_tables()` (and `nm_join()`, which calls it) will... what? 
  * Pick the final one? 
  * Fail and skip the table if there is more than one?

R/nm-join.R
```diff
@@ -0,0 +1,179 @@
...
+  }
+  col_order <- names(.d)
+
+  .join_col <- toupper(.join_col)
+  if (!(.join_col %in% names(.d))) {
+    stop(glue("couldn't find `.join_col` {.join_col} in data with cols: {paste(names(.d), collapse = ', ')}"))
+  }
+
+  if (.superset) {
+    join_fun <- function(x, y, ...) left_join(y, x, ...)
```
kyleam, line 106: Is it just a stylistic choice to swap the `left_join` args rather than use `right_join`?  (The tests pass for me if I use `right_join`, so I think it might be.) If it makes a practical difference, it might be worth a comment/test case to avoid a regression by someone "simplifying" it later.

R/nm-join.R
```diff
@@ -0,0 +1,179 @@
...
+      # toss .join_col, if present, because we're joining on ID
+      tab[[.join_col]] <- NULL
+
+      # do the join
+      tab <- drop_dups(tab, .d, "ID", .n)
+      col_order <- union(col_order, names(tab))
+      .d <- join_fun(tab, .d, by = "ID")
+    }
+
+    if (nrow(tab) == nrec) {
```
kyleam, line 151: The `== nid` and `== nrec` should never both be followed, right?  In that case, I think it'd be clearer to restructure the three outer `if`s above into `if ... else if ... else ...` to make invalid states inaccessible.

R/nm-join.R
```diff
@@ -0,0 +1,180 @@
...
+  ))
+
+  return(select(.d, !!col_order))
+}
+
+
+#' Drop duplicate columns to prepare for join
+#' @keywords internal
+drop_dups <- function(.new_table, .dest_table, .join_col, .table_name) {
+  new_cols <- c(setdiff(names(.new_table), names(.dest_table)))
```
kyleam, line 171: superfluous `c()`?

tests/testthat/test-nm-file.R
```diff
@@ -97,3 +103,16 @@ test_that("nm_par_tab() works [BBR-NMF-007]", {
   expect_equal(ncol(.d), 6)
   expect_equal(nrow(.d), DATA_TEST_ROWS_IGNORE)
 })
+
+test_that("nm_file() error with no 'TABLE NO' found [BBR-NMF-008]", {
```
kyleam, line 107: s/error/errors/ for consistency with other descriptions

tests/testthat/test-nm-join.R
```diff
@@ -0,0 +1,156 @@
...
+  withr::with_options(list(bbr.verbose = TRUE), {
+    expect_message({
+      test_df <- nm_join(MOD1)
+    }, regexp = "join stats")
+    expect_equal(nrow(test_df), DATA_TEST_ROWS_IGNORE)
+    expect_equal(ncol(test_df), DATA_TEST_COLS + TAB_NEW_COLS + PARTAB_NEW_COLS + 1)
+  })
+})
+
+test_that("nm_join() works correctly summary object [BBR-NMJ-001]", {
```
kyleam, line 18: This is minor and probably not worth worrying about, but I stumble when parsing the tail part of descriptions like this (there are a number in this PR).  I'm pretty sure you're wording it like that for the benefit of mrgvalprep (so that it can grab the common prefix), but I think ideally it could be done in a way that would still flow for readers of the test file.  Perhaps it could be something like "nm_join() works correctly: summary object", and mrgvalprep could learn to strip a trailing ":".

R/nm-join.R
seth127: I think it had something to do with the column ordering, but since we're doing that manually at the bottom then this shouldn't matter at all. I agree: let's switch it to prevent confusion down the road.

R/nm-join.R
seth127: I can never decide what is the least annoying way to do this in R. It's basically a case statement, but those are hideous in R. I was kinda using the "left justified path" mentality, but I think you're right that if/else is probably cleaner here.

seth127:
> Alright, I'm on board for data.table if both of you are.

Going to address this (and other parts of https://github.com/metrumresearchgroup/bbr/pull/430#issuecomment-964291763) in a separate PR before this is released, but after we merge this into `develop`

kyleam (approved):
> Going to address this (and other parts of #430 (comment)) in a separate PR before this is released, but after we merge this into develop

Handling the `data.table` conversion in a follow-up PR works for me.  This one is good to go from my pov.

kylebaron:
It might be worth a check for `ONEHEADER` before reading. 

```r
$TABLE
[1] "NOPRINT ONEHEADER NOAPPEND FILE=foobar.tab"        
[2] "       NUM CL V2 Q V3 KA ETAS(1:LAST) "
```
You can exit if you don't find `ONEHEADER` and refer to `nm_file` or one of the other available read functions that will look for text in the middle of a table. 

Also: since you are parsing the the ctl, you can tell if it is a simulation run or not:

```r
any(grepl("^SIM", labs))
```

That's how you get `TABLE NO.` in the table. Just ask to use something else. I'm not against developing something on our own to do this; but should be a different thing better suited for that (we don't do a lot of simulation with nonmem).

PR #431, opened 2021-11-09
fix: use_bbi() version message respects bbr.verbose

seth127:
Setting `version_message()` (called by `use_bbi()`) to respect `options("bbr.verbose")` and `local_options(bbr.verbose = FALSE)` in `test-use-bbi.R` to suppress obnoxious messages that filter through test suite output.

### DIFF
--- a/R/use-bbi.R
+++ b/R/use-bbi.R
@@ -336,6 +336,7 @@ bbi_version <- function(.bbi_exe_path = getOption('bbr.bbi_exe_path')){
 #' @param release_v Character scalar for version number of current release
 #' @keywords internal
 version_message <- function(local_v, release_v){
+  if (!isTRUE(getOption("bbr.verbose"))) return(invisible(NULL))
 
   if (local_v == "") {
     cat(cli::col_red("No version currently installed "))
--- a/R/use-bbi.R
+++ b/R/use-bbi.R
@@ -2,7 +2,9 @@
 #' @description Identifies system running and pulls the relevant tarball for the
 #'   current release of bbi from GitHub, and then installs it in the directory
 #'   passed to `.dir`. If used in an interactive session, will open an
-#'   installation menu confirming the installed version.
+#'   installation menu confirming the installed version. This function will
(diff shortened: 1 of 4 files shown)


kyleam (commented):
The base chosen for this PR (dc9ad2e, 2021-11-04) is confusing because there are no other uses of `bbr.verbose` yet. `bbr.verbose` was introduced into `develop` with a65d080 (Merge pull request #426 from metrumresearchgroup/feat/nm_file, 2021-11-08).

---

So this gets rid of these messages:

```
# base (dc9ad2e)
 > devtools::test(filter = "use-bbi")
 ℹ Loading bbr
 ℹ Testing bbr
 ✔ | F W S  OK | Context
 ⠏ |         0 | test-use-bbi
 No version currently installed  - Current release: 3.1.0
 ⠋ |         1 | test-use-bbi
 No version currently installed  - Current release: 3.1.0
 ⠙ |         2 | test-use-bbi
 No version currently installed  - Current release: v2.1.2
 ✔ |         5 | test-use-bbi [1.8s]
[...]


# this PR
> devtools::test(filter = "use-bbi")
 ℹ Loading bbr
 ℹ Testing bbr
 ✔ | F W S  OK | Context
 ✔ |         5 | test-use-bbi [1.6s]

[...]
```

I like the end result, though I'm not sure if `version_message` is the best spot to handle it.  `version_message` is called only within `install_menu`, and `install_menu` is called only within `use_bbi`.  `use_bbi` has a `.quiet` parameter.  Should that default to `bbr.verbose` and control whether `version_message` is called?

seth127:
Good points. I must not have pulled `develop` first, because `bbr.verbose` came in with #426. I pull `develop` and this rebase on it.

In terms of the structure of it, in `nm_file()` and friends we're going with the pattern of controlling verbosity only with `options(bbr.verbose)` (and no argument) so I think my preference would be to begin deprecating that `.quiet` arg and doing the same thing here. I'll take a look at that and commit something for review.

kyleam (approved):

PR #432, opened 2021-11-09
silence recent 'R CMD check' notes about undefined variables

kyleam:
The decision to use globalVariables() for "." is based on discussion
in magrittr's issue 29.


seth127 (approved):
Well done

Issue #408, opened 2021-09-27
friendlier tail_output() when run completed


callistosp:
# Problem

Currently, if you call `tail_output()` when a NONMEM model has finished running, it simply says that the file is not found. It would be more easily interpreted if we could catch that error and just tell the user that the model has finished running.

seth127:
This is a good idea. If you or @barrettk wants to knock this out sometime, you are welcome to it. Either way, we'll put it on the docket for the next release.

PR #433, opened 2021-11-09
tail_output: when OUTPUT doesn't exist, check model state

kyleam:
As described in gh-408, calling `tail_output` with a completed model gives an unintuitive error.

```
> tail_output(m)
Error: '.../inst/model/nonmem/basic/1/OUTPUT' does not exist.
```

This PR (second commit) changes that to

```
> tail_output(m)
Model already finished running
```


seth127 (approved):
good stuff

PR #434, opened 2021-11-11
Refactor/add data.table

seth127:
Refactoring the internals of `nm_file()` (and potentially some other functions) based on the discussion in #430


seth127:
@kylebaron I think this should address all the concerns you mentioned in #430 

there are some changes across files in this though, so I think you need to pull this whole branch (instead of copy/pasting files) to try it out.

R/nm-file.R
```diff
@@ -115,16 +122,33 @@ nm_data <- function(.mod, .sep = ",") {
...
 #' @keywords internal
 nm_file_impl <- function(.path, .est_method) {
   # read file and find top of table
   verbose_msg(glue("Reading {basename(.path)}"))
-  .txt <- read_lines(.path)
-  .est <- which(str_detect(.txt, "^ *TABLE NO"))
+
+  # get line numbers of TABLE lines
+  checkmate::assert_file_exists(.path)
+  .est <- processx::run(
```
kylebaron, line 135: I don't think we need to do this every time. Let's talk about my previous comments and how to determine this.

seth127: I timed it on some pretty big files (100k lines, I think) and it was super quick. I hear your point, but I think any other solution is going to end up being more complicated than it's worth, especially given that this doesn't really cause problems.

kylebaron: What is the use case here? Why are you looking?

seth127: although, @dpastoor I meant to have a conversation about how we should do this on Windows (or any system without `grep` I guess). Any thoughts? Our "support" for windows is a bit thin, but at the very least we should give a reasonable warning or something. Thoughts?

kylebaron: @seth127 answer this question first: why are you looking through the file in the first place?

seth127: to see if there's more than one table in it

kylebaron: Ok, good.  Why might there be more than one table there? .... in this use case?

kylebaron: The main case where you might see multiple tables would be a simulation run; we should refuse to process this and can be easily checked from the names in the control stream. As you note, someone could (possibly) could write the same table file by mistake (we can say "mistake" b/c I don't think anyone is really doing this intentionally). This is an easy check of duplicate table file names. Once you got that far, I think you're in the clear and you don't need to be grepping the file every time. I think grepping the file is great when you have a large simulation run to process but we already took that option off the table, right?  

I'll put it another way: this function is to make more convenient something that we are already doing: reading in the data and a straightforward read in of the table file (with fread or readr). We are already doing it in this straightforward, read-the-table way and I wanted to capture that workflow.  

I can't guarantee that no one will ever come across a problematic situation where reading the table gets mangled. If you wanted to be super careful to exit in a nice way no matter what, then wrap in `try()`, let the user know something failed (unlikely IMO if we rule out $SIM and duplicate table files), and ask them to go back and ready the intended use of the function.

R/nm-file.R
```diff
@@ -115,46 +109,47 @@ nm_data <- function(.mod, .sep = ",") {
...
+    if (str_detect(.w$message, "Stopped early.+TABLE NO")) {
+      W <<- paste(
+        "nm_file() does not support files with multiple tables in a single file.",
+        glue("{.path} appears to contain multiple tables and will be skipped."),
+        sep = "\n"
+      )
+      invokeRestart("muffleWarning")
     } else {
-      .est_method <- 1
+      warning(.w)
```
kyleam, line 142: Neat.  

I think you want the `invokeRestart` even if you go down the `else` arm.  Otherwise the warning is repeated.

R/nm-tables.R
```diff
@@ -7,6 +7,16 @@
 #' the related [nm_join()] function. This function will print the number of rows
 #' and columns when each file is loaded. This **printing can be suppressed** by
 #' setting `options(bbr.verbose = FALSE)`.
+#'
+#' @details
+#' As described in Value, `nm_tables()` returns a named list of tibble. To
```
kyleam, line 12: s/tibble/tibbles/

kyleam (changes requested):

kyleam (approved):
Thanks for the update.  LGTM.

PR #443, opened 2021-12-01
print.bbi_nonmem_summary: account for ONLYSIM

kyleam:
This is the sister PR to metrumresearchgroup/bbi#258.  Before that PR, feeding an `ONLYSIM` run to `print.bbi_nonmem_summary` makes the underlying bbi call fail.  After that PR, `print.bbi_nonmem_summary` itself will fail:

```
> m %>% model_summary()
Dataset: ../1001.csv

Records: 2645	 Observations: 2620	 Subjects: 20

Objective Function Value (final est. method): NA

Estimation Method(s):

–

No Heuristic Problems Detected

Error: All columns in a tibble must be vectors.
✖ Column `parameter_names` is NULL.
✖ Column `estimate` is NULL.
✖ Column `fixed` is NULL.
```

---

I'm marking this as a draft for a few reasons:

 - [x] the bbi PR hasn't yet settled (edit: now merged)

 - [x]  this leads to `summary_log` test failures

   I haven't looked into it yet to determine if that's purely a test issue or if `summary_log` also needs to account for `ONLYSIM` models. (edit: just test tweaks, in particular adjusting assertions to expect the new model in `inst/model/nonmem/complex`, were needed)

  - [x] perhaps other spots (e.g., `param_estimates`?) need to be adjusted to give a more informative error or to avoid an error (edit: `param_estimates` is the main thing I spotted; updated)


seth127:
when I run this with `bbi 3.1.1-rc1` I get
```
Error (test-print.R:194:5): print.bbi_nonmem_summary ONLYSIM [BBR-PRNT-003]
Error: invalid version specification '3.1.1-rc1'
Backtrace:
 1. bbr::skip_if_old_bbi("3.1.0") test-print.R:194:4
 2. base::package_version(v_bbi) /data/home/sethg/bbr/R/helpers-for-tests.R:47:6
 3. base::.make_numeric_version(...)
 ```

I think we want to treat this (and anything that doesn't look like a proper numeric version) the same as `develop`, i.e. run the test.

seth127:
as part of this PR, let's bump the `bbi` version in `.drone.yml` to `3.1.1` (once that bbi version is released) so that the new tests will run in CI.

basically change everywhere it says `bbi -v 'v3.1.0'` in the `.drone.yml` (5 places, I think)

kyleam:
> (once that bbi version is released) 

Released....

> so that the new tests will run in CI.

... and done.

kyleam:
> Was there an issue that this closes? I looked briefly and didn't see one.

No, it was driven by the issue in bbi: metrumresearchgroup/bbi#224

PR #444, opened 2021-12-01
test: summary-log: load dplyr

kyleam:
The unqualified `pull` calls added in d27ac23 (test: updating test
refs for eigenvalue_issues heuristic, 2021-10-20) lead to a
test-summary-log failure on my end:

    $ Rscript -e 'devtools::test(filter = "summary-log")'
    [...]
    Error in `pull(., minimization_terminated)`: could not find function
    "pull"
    Backtrace:
     1. testthat::expect_false(filter(sum_df2, run == "1001") %>%
    [...]

Rather than qualify those calls [*], follow what test-param-labels.R
does and just load dplyr.

[*] ... as well as the other unqualifed dplyr function calls in this
    file that don't error because they happen to be in NAMESPACE due
    to imports elsewhere

### DIFF
--- a/tests/testthat/test-summary-log.R
+++ b/tests/testthat/test-summary-log.R
@@ -1,5 +1,7 @@
 context("Test creating summary logs")
 
+suppressPackageStartupMessages(library(dplyr))
+
 skip_if_not_drone_or_metworx("test-summary-log")
 
 # helper to run expectations


tests/testthat/test-summary-log.R
```diff
@@ -1,5 +1,7 @@
 context("Test creating summary logs")
 
+suppressPackageStartupMessages(library(dplyr))
```
dpastoor, line 3: my preference would be to directly namespace the pull rather than globally load dplyr if we do it, or just simplify the code to not need pull (as convenient as it is for quick writing

```
filter(sum_df2, run == "1001") %>% pull(minimization_terminated)
```

to just

```
filter(sum_df2, run == "1001")$minimization_terminated
```

dpastoor: example in #445 for consideration

kyleam: > my preference would be to directly namespace the pull rather than globally load dplyr if we do it

Sure, that's fine by me.  As noted in my initial message ("Rather than qualify those calls [*], follow what test-param-labels.R does and just load dplyr"), my decision to load was pretty much based on the fact that this same line is already elsewhere in the tests.

kyleam:
superseded by #445

PR #445, opened 2021-12-02
drop pull from test

dpastoor:
a quick refactor to no longer need pull in the tests

### DIFF
--- a/tests/testthat/test-param-labels.R
+++ b/tests/testthat/test-param-labels.R
@@ -65,9 +65,10 @@ for (MODEL_PICK in MODEL_PICKS) {
       select(-param_type)
 
     # join against reference to make sure they're the same
-    ref_df <- ref_df %>% dplyr::mutate(!!SUMMARY_PARAM_NAMES := ifelse(param == "THETA",
-                                                                paste0(param, var1),
-                                                                paste0(param, "(", var1, ",", var2, ")")))
+    ref_df <- ref_df %>% dplyr::mutate(
+      !!SUMMARY_PARAM_NAMES := ifelse(param == "THETA",
+                                      paste0(param, var1),
+                                      paste0(param, "(", var1, ",", var2, ")")))
 
     join_df <- ref_df %>%
       dplyr::full_join(.label_df, by = names(.label_df))
@@ -106,15 +107,19 @@ withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
 
       # join to constuct full parameter table
       suppressSpecificWarning({
-        .new_df <- dplyr::inner_join(.param_df,
-                              .label_df %>% apply_indices(.omega = MODEL_PICK$omega, .sigma = MODEL_PICK$sigma) %>% dplyr::select(-param_type),
-                              by = SUMMARY_PARAM_NAMES)
+        .new_df <- dplyr::inner_join(
+          .param_df,
+          .label_df %>%
+            apply_indices(.omega = MODEL_PICK$omega, .sigma = MODEL_PICK$sigma) %>%
+            dplyr::select(-param_type),
+          by = SUMMARY_PARAM_NAMES)
       }, .regexpr = "Column .+ has different attributes on LHS and RHS of join")
 
       # join against reference to make sure they're the same
-      ref_df <- ref_df %>% dplyr::mutate(!!SUMMARY_PARAM_NAMES := ifelse(param == "THETA",
-                                                                  paste0(param, var1),
-                                                                  paste0(param, "(", var1, ",", var2, ")")))
+      ref_df <- ref_df %>% dplyr::mutate(
+        !!SUMMARY_PARAM_NAMES := ifelse(param == "THETA",
+                                        paste0(param, var1),
+                                        paste0(param, "(", var1, ",", var2, ")")))
 
       join_df <- ref_df %>%
         dplyr::full_join(.new_df, by = names(.label_df))
--- a/tests/testthat/test-param-labels.R
+++ b/tests/testthat/test-param-labels.R
@@ -1,8 +1,5 @@
 context("test parsing labels for parameter table")
 
-suppressPackageStartupMessages(library(glue))
-suppressPackageStartupMessages(library(dplyr))
-
 # constants
 PL_MODEL_DIR <- file.path(REF_DIR, "param-labels", "tidynm_extdata")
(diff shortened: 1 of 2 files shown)


kyleam (commented):
Rewrite looks good to me.  However, since the motivation is to avoid loading dplyr, shouldn't this drop the `library` call added in the last commit?  Or better yet, use `develop` rather than `smlg-pull` as the base to avoid adding the line just to revert it?

---

Fwiw, continuing the spirit of this change, here's what it'd look like to remove the other `suppressPackageStartupMessages(library(dplyr))` that I mentioned in gh-444:

```diff
diff --git a/tests/testthat/test-param-labels.R b/tests/testthat/test-param-labels.R
index fd6208e..d7ca429 100644
--- a/tests/testthat/test-param-labels.R
+++ b/tests/testthat/test-param-labels.R
@@ -72,7 +72,7 @@ test_that("param_labels.character errors on vector [BBR-PLB-003]", {
                                                                 paste0(param, "(", var1, ",", var2, ")")))
 
     join_df <- ref_df %>%
-      full_join(.label_df, by = names(.label_df))
+      dplyr::full_join(.label_df, by = names(.label_df))
 
     # test that the full join didn't add any rows (i.e. there are no mismatches)
     expect_equal(nrow(join_df), nrow(ref_df))
@@ -108,8 +108,8 @@ test_that("param_labels.character errors on vector [BBR-PLB-003]", {
 
       # join to constuct full parameter table
       suppressSpecificWarning({
-        .new_df <- inner_join(.param_df,
-                              .label_df %>% apply_indices(.omega = MODEL_PICK$omega, .sigma = MODEL_PICK$sigma) %>% select(-param_type),
+        .new_df <- dplyr::inner_join(.param_df,
+                                     .label_df %>% apply_indices(.omega = MODEL_PICK$omega, .sigma = MODEL_PICK$sigma) %>% select(-param_type),
                               by = SUMMARY_PARAM_NAMES)
       }, .regexpr = "Column .+ has different attributes on LHS and RHS of join")
 
@@ -119,7 +119,7 @@ test_that("param_labels.character errors on vector [BBR-PLB-003]", {
                                                                   paste0(param, "(", var1, ",", var2, ")")))
 
       join_df <- ref_df %>%
-        full_join(.new_df, by = names(.label_df))
+        dplyr::full_join(.new_df, by = names(.label_df))
 
       # tests
       expect_equal(nrow(join_df), nrow(ref_df))
```

dpastoor:
yep good point, i'll repoint to develop, and also 👍 to the additional patch

tests/testthat/test-summary-log.R
```diff
@@ -1,5 +1,7 @@
 context("Test creating summary logs")
 
+suppressPackageStartupMessages(library(dplyr))
```
kyleam, line 3: This line/commit from gh-444 is still sneaking in because the base was changed on github, the commit still needs to be dropped from the branch.

tests/testthat/test-param-labels.R
```diff
@@ -101,26 +98,26 @@ withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
...
-        .new_df <- inner_join(.param_df,
-                              .label_df %>% apply_indices(.omega = MODEL_PICK$omega, .sigma = MODEL_PICK$sigma) %>% select(-param_type),
+        .new_df <- dplyr::inner_join(.param_df,
+                              .label_df %>% apply_indices(.omega = MODEL_PICK$omega, .sigma = MODEL_PICK$sigma) %>% dplyr::select(-param_type),
                               by = SUMMARY_PARAM_NAMES)
       }, .regexpr = "Column .+ has different attributes on LHS and RHS of join")
 
       # join against reference to make sure they're the same
-      ref_df <- ref_df %>% mutate(!!SUMMARY_PARAM_NAMES := ifelse(param == "THETA",
+      ref_df <- ref_df %>% dplyr::mutate(!!SUMMARY_PARAM_NAMES := ifelse(param == "THETA",
```
kyleam, line 115: Not something that needs to be coupled with this clean-up (and perhaps not something to bother with at all?), but just a note for completeness:  the tests include other unqualified `mutate`s as well as `select`s and `filter`s that make it through without an error because they happen to be in NAMESPACE.

kyleam (changes requested):
The `suppressPackageStartupMessages(library(dplyr))` from gh-444 is still sneaking in.

tests/testthat/test-summary-log.R
dpastoor: ah - yeah - what is the best way of dropping that commit from the branch?

kyleam: Given it's the initial commit, you could

  * use interactive rebase (`git rebase -i origin/develop` and then drop the first line when the editor pops up)
  * a non-interactive rebase of just the top two commits: `git rebase --onto origin/develop HEAD~2`

Then force push.

---

Or just say "why bother" and add a new commit that reverts that `library` call :)

dpastoor: i'm all about racking up those commit numbers, especially with how many @seth127 steals from me 😆

kyleam: Ha, in this case it's actually _my_ commit that would be dropped, but all right, forward-only it is.  I'll push a commit that drops/reverts the `library(dplyr)` call that initially prompted this PR, and I'll also push a commit to cleanup a few function calls whose alignment was thrown off by dd8578c.

kyleam (approved):

kyleam:
Looks like both builds failed due to a failing wget call in one of the "install bbi" phases.  I'll restart the jobs to see if the issue clears up, though `install_bbi.sh` should probably be tweaked to retry/wait to reduce transient failures.

- https://github-drone.metrumrg.com/metrumresearchgroup/bbr/2324/3/3
- https://github-drone.metrumrg.com/metrumresearchgroup/bbr/2323/1/3

kyleam:
I said:
> I'll restart the jobs to see if the issue clears up

Well kind of.  The `wget` call didn't fail, but now some `use_bbi` tests failed (a version check on one build, a mode check on another).  So, it's not consistent across builds and I can't trigger it locally...

- https://github-drone.metrumrg.com/metrumresearchgroup/bbr/2325/1/4
- https://github-drone.metrumrg.com/metrumresearchgroup/bbr/2326/1/5

Issue #446, opened 2021-12-07
Setup bbi from pre-built binaries helper


seth127:
something a regulator could "just run" and it would have bbi set up and ready to work (_not_ for NONMEM, but for things like `model_summary()`)

this "something" could be a makefile or shell script or something, or could be a bbr function (@seth127 leans towards a bbr function, so it could be included in script)

# One idea: `bbr::setup_bbi(.dir)`

* This would be called by the regulator who has the eappendix with the pre-built binaries in a specified place
* We could include this call at the top of our simulation scripts
  * Maybe even do this automatically, via the [create-electronic-appendix.R](https://ghe.metrumrg.com/templates/msproject/blob/master/script/e-appendix/create-electronic-appendix.R) script

Point to `.dir` and it:

* looks for pre-built binaries in that dir
* checks the OS
* sets `options(bbr.bbi_exe_path)` to the right binary (maybe it moves/renames?)

@seth127 initial idea was to add this `.pre_built_dir` argument to `use_bbi()`, but this is probably a better path because it actually _sets_ the `bbi_exe_path` instead of _reading_ from it and installing to it (because that would mean we need to set `bbi_exe_path` first)

seth127:
# High-level steps

PR #447, opened 2021-12-08
fix: bbi_{help,init}: restore relay of bbi output

kyleam:
When on a call today with @dpastoor working on gh-446, `bbi_help` didn't provide any output.  At first, we thought it might be Windows-specific, but it looks like a longstanding regression for all systems.

---

bbi_help() and bbi_init() are supposed to relay bbi's combined
stdout/stderr via a call to cat().  However, both functions always
produce empty output because they refer to a non-existent "output"
element.

This regression happened in dbe29fd (more debugging and cleanup,
2020-02-26), which updated bbi_exec() to return a custom list rather
than a process.  As a result, the output element name was changed from
"output" to "stdout" (indirectly through RES_PROCESS, which would
later become PROC_STDOUT).

Update bbi_help() and bbi_init() to use PROC_STDOUT.

### DIFF
--- a/tests/testthat/test-bbr.R
+++ b/tests/testthat/test-bbr.R
@@ -96,3 +96,10 @@ test_that("bbi_init passes .bbi_args [BBR-BBR-008]", {
   expect_equal(bbi_yaml$threads, 36)
   expect_true(bbi_yaml$no_shk_file)
 })
+
+test_that("bbi_help captures and relays bbi output [BBR-BBR-009]", {
+  withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
+    res <- paste(capture.output(bbi_help()), collapse = "\n")
+  })
+  expect_true(stringr::str_detect(res, "bbi cli version"))
+})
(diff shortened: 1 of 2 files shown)


tests/testthat/test-bbr.R
```diff
@@ -96,3 +96,10 @@ test_that("bbi_init passes .bbi_args [BBR-BBR-008]", {
   expect_equal(bbi_yaml$threads, 36)
   expect_true(bbi_yaml$no_shk_file)
 })
+
+test_that("bbi_help captures and relays bbi output [BBR-BBR-009]", {
+  withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
+    res <- paste(capture.output(bbi_help()), collapse = "\n")
+  })
+  expect_true(stringr::str_detect(res, "bbi cli version"))
+})
```
kyleam, line 105: It'd be good to test `bbi init` too, but that command doesn't look very chatty.  I wasn't able to come up with a way to invoke `bbi_init()` that would lead to output from `bbi init` (versus a quiet success or a failure caught by `bbi_init()` _before_ calling `bbi init`).

seth127: makes sense. I can't think of any messages that `bbi init` _should_ relay in any obvious scenarios.

seth127 (approved):
LGTM

PR #449, opened 2021-12-10
fix: model-summary: abort if more than one lst file is found

kyleam:
If a model whose directory contains more than one lst is fed to
model_summary(), check_lst_file() will return all the lst files,
leading to a bbi call like this from within the model directory:

    bbi nonmem summary 1 2 --json

The output of that call isn't the expected structure of

    {
      "run_details": {...},
      "run_heuristics": {...}
      ...
    }

Instead it is

    {
      "Results": [
        {
          "run_details": {...},
          "run_heuristics": {...},
          ...
      ]
    }

which causes the downstream create_summary_object() call to fail with
an error about missing keys.

It's probably pretty rare for a user to accidentally create multiple
lst files.  However, it still seems worth guarding against given that 1) 
this case was hit into in the wild and led to a fair amount of head
scratching and 2) it's easy to do.

---

cc @timwaterhouse

### DIFF
--- a/R/model-summary.R
+++ b/R/model-summary.R
@@ -237,11 +237,16 @@ nonmem_summary <- function(
 
 #' Private helper function to look for .lst function in a directory
 #' @param .x The directory path to look in for the lst file
+#' @importFrom glue glue
 #' @keywords internal
 check_lst_file <- function(.x) {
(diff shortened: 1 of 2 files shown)


seth127 (approved):
LGTM

PR #457, opened 2022-01-31
feat/ param_estimates_compare() 

### LINKED ISSUE #453: Bootstrap helper for param_estimates_batch()
There are two things here (that should just be done together, though potentially in different PR's)

1. We need to change the `_` to `,` in the column names that come back from `param_estimates_batch()`. This is annoying and we might need a more specific spec for it, but the important part is that it inhibits downstream stuff (table code, etc.) which expects column names like `OMEGA(1,1)` (which matches what comes from `model_summary()` and `param_estimates()`) instead of `OMEGA(1_1)`.
2. We should add a little helper for comparing the results of `param_estimates_batch()` to a single model. This is primarily for bootstraps, but that's a primary use case of `param_estimates_batch()`.

An ad-hoc implementation (from Example Project) for the helper in point 2:

```r
#' Compare parameter estimates
#' 
#' Let's you compare the parameter estimates from  a single model 
#' against those in tibble like what's returned from 
#' [param_estimates_batch()].
#' 
#' @param .mod `bbi_model` object
#' @param .comp A tibble with columns for each parameter and rows for each model (like what's returned from 
#' [param_estimates_batch()])
#' @param .probs Numeric vector with between 0 and 1 to be passed through to `quantile`.
#' Represents the quantiles to calculate for parameter estimates in `.comp`.
#' 
#' @export
param_estimates_compare <- function(.mod, .comp, .probs = c(0.025, .5, 0.975)) {
  
  boot_qt <- .comp %>% 
    summarise(across(-run, quantile, probs = .probs)) %>% # better select only the params
    t() %>%
    as.data.frame() %>%
    rownames_to_column()
  colnames(boot_qt) <- c("parameter_names", paste0(.probs * 100, "%"))
  
  .mod %>%
    model_summary() %>% # maybe we optionally take `bbi_model_summary` too, so need to check this
    param_estimates() %>%
    select(parameter_names, estimate) %>%
    rename(original_estimate = estimate) %>%
    left_join(boot_qt, by = "parameter_names")
  
}
```

(This currently doesn't work though, because of the `_` vs. `,` problem mentioned in point 1)

seth127:
Closes #453


seth127:
This is possibly finished, but want to get some scientist feedback first, to avoid a repeat of the #453 issue that made this necessary. @timwaterhouse @kylebaron @callistosp I'll send you an Example Project branch where you can try this out, but feel free to comment on this PR if you have input on the code/usage. Thanks.

Also, drone doesn't seem to be working right now...

timwaterhouse (commented):
Looks good. No comments on code, just a couple of typos.

seth127:
Scientific/user review provided by @timwaterhouse @kylebaron @KatherineKayMRG @riggsmm and updates made in bec7c8e80684ab7269fbf56ee0581bf8df1b5c1a

Ready for final code review.

kyleam (dismissed):
Left comment about `R CMD check` warnings and a doc typo.  Otherwise looks good to me.

seth127:
@kyleam I had a thought last night about [this line](https://github.com/metrumresearchgroup/bbr/blob/01b2f287fc203deac8c63b9c99b34c84b4187fa7/R/param-estimates-batch.R#L150) where we select the columns to summarize.

My thinking was that this is overly inflexible and we should let the user pass in a vector that's passed to `starts_with()`, defaulting to `c("THETA", "SIGMA", "OMEGA")`. But then I thought maybe it would actually better to use [`matches()`](https://tidyselect.r-lib.org/reference/starts_with.html) so the user can pass in a regex, and have it default to `"^(THETA|SIGMA|OMEGA)"`. 

Basically I'm thinking this will mostly be used for NONMEM, in which case no one would change the default, but could also be useful for Stan or whatever else _if_ you could specify which columns to summarize and pivot. Any thoughts on all of that?

kyleam:
In terms of adding the argument or not, either way seems fine to me.  If you hard code it and focus on the NONMEM use case, it's easy (and backward compatible) to add an argument later if there ends up being interest.  

re `starts_with` vs `matches`: what about just letting any `select(...)` expression be passed?

seth127:
Hmmm, that's an interesting point. And it could just default to something like `.compare_cols = starts_with(c("THETA", "SIGMA", "OMEGA"))`? I like that. Max flex.

seth127:
@kyleam made final edits and added the `.compare_cols` arg. Ready for final review.

My plan is to make a `1.3.0` release with this next week, so that it can get on the new MPN.

R/param-estimates-batch.R
```diff
@@ -103,3 +106,101 @@ param_estimates_batch <- function(.path,
...
+param_estimates_compare <- function(
+  .param_df,
+  .orig_mod = NULL,
+  .compare_cols = starts_with(c("THETA", "SIGMA", "OMEGA")),
+  probs = c(.5, 0.025, 0.975),
+  na.rm = FALSE
+) {
+
+  comp_df <- .param_df %>%
+    select(.compare_cols)
```
kyleam, line 158: I think you'll need to use `{{ .compare_cols }}` to get the expected evaluation.

https://dplyr.tidyverse.org/articles/programming.html#indirection-3

```
> param_estimates_compare(param_df, .compare_cols = THETA2)
Error in `instrument_base_errors()`:
! object 'THETA2' not found
Caused by error in `env_get()`:
! object 'THETA2' not found
Run `rlang::last_error()` to see where the error occurred.

# with `{{ .compare_cols }}`
> devtools::load_all()
ℹ Loading bbr
setting `bbr.DEV_no_min_version` to TRUE so no validation of bbi version will occur
> param_estimates_compare(param_df, .compare_cols = THETA2)
  parameter_names     50%    2.5%   97.5%
1          THETA2 54.6151 54.6151 54.6151
```

kyleam (changes requested):

R/param-estimates-batch.R
seth127: > In my view, the expected evaluation for the tidy selection should be used. Is there some downside of using {{ .compare_cols }} that I'm overlooking?

No, I guess there isn't. And considering these other cases, I think you're right we should switch to embrace. Would you change the docs for the argument (from what they are in the PR now) or do you think it's accurate and clear as is?

kyleam: The description seems clear to me as is.

kyleam (approved):

PR #462, opened 2022-02-09
pkgdown: include param_estimates_compare in index

kyleam:
Generating the docs for the 1.3.0 release, I noticed a warning message about `param_estimates_compare` missing from the index.  This adds that function to the index and silences another message about an internal function.

I should have taken a look before the release, but, this is the only change on top of develop, so I'll regenerate the 1.3.0 docs once this is merged in.


seth127 (approved):
good catch

PR #464, opened 2022-02-16
Smaller logo png

seth127:
Made the logo the recommended 250x275 pixels because the new pkgdown seems to print it at actual full size (and ours was previously huge)


seth127:
pkgdown site previously looked like this:
Screen Shot 2022-02-15 at 8 35 54 PM

Now it looks like this:
Screen Shot 2022-02-15 at 9 17 04 PM

Nicer, I think.

kyleam (approved):

Issue #465, opened 2022-02-22
Partial matching names when parsing ctl file


kylebaron:
Can't count on full names being there; grep out `^TAB` in names. 

```r
  # get file names from table statements and construct paths
  .f <- .l[names(.l) == "TABLE"] %>%

```

barrettk:
Hey @kylebaron do you have any examples for when this fails? Would be happy to jump on a call when you have time.

barrettk:
Yes thIs was addressed and ok to close. Thanks

PR #466, opened 2022-03-04
fix per Kyle Baron's suggestion

barrettk:
Reference Issue: https://github.com/metrumresearchgroup/bbr/issues/465
 - some people use "TAB" instead of "TABLE"
    - same with "EST" instead of "ESTIMATION"

Fix includes grep-ing out "TAB" instead of matching "TABLE"


barrettk:
Conversation with KyleB pasted below for convenience:

> Hey Kyle, do you have any time today or next week to go over [this issue](https://github.com/metrumresearchgroup/bbr/issues/465)? If I understand correctly, I believe this will be a simple fix. Seems like you're suggesting some of the names could be "TAB" rather than "TABLE", which would be missed by the call you mentioned.

> Yeah; same with EST and ESTIMATION … nonmem does partial matching and people use both.

> does bbr have an issue with EST vs ESTIMATION as well?

> It might not be looking there.

R/nm-tables.R
```diff
@@ -88,7 +88,7 @@ nm_table_files <- function(.mod, .check_exists = TRUE) {
   out_dir <- get_output_dir(.mod, .check_exists = .check_exists)
 
   # get file names from table statements and construct paths
-  .f <- .l[names(.l) == "TABLE"] %>%
+  .f <- .l[grep("TAB", names(.l))] %>%
```
seth127, line 91: I'd say make this `grep("^TAB", names(.l))` so that we don't snag any names that happen to have `TAB` in the middle somewhere. I don't know if that's even a possibility with NONMEM, but if they allow `$TAB` or `$TABLE` for specifying it then there shouldn't be any others that _start_ with `TAB`

seth127 (changes requested):
minor change to the regex

R/nm-tables.R
barrettk: Yeah that's a good call - taken care of

barrettk:
Last commit addressed issue https://github.com/metrumresearchgroup/bbr/issues/467

seth127 (approved):
Looks good.

PR #469, opened 2022-03-07
bbr validation stories to YAML

barrettk:
Convert validation stories to yaml

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -71,7 +71,9 @@ Suggests:
     knitr,
     rmarkdown,
     testthat (>= 2.1.0),
-    withr
+    withr,
+    mrgvalidate,
+    mrgvalprep (>= 0.0.4)
 VignetteBuilder: 
     knitr
 Encoding: UTF-8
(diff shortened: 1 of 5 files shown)


seth127 (dismissed):
Stories look good. I should've put this in the original issue, but can you add a version of [this `build-validation-docs.R` script](https://github.com/metrumresearchgroup/pmforest/blob/main/inst/validation/build-validation-docs.R) in this same folder too. You'll need to change some of the constants at the top, but other than that it should work I think.

You'll also have to add the following to `Suggests` in `DESCRIPTION`

```
    mrgvalidate,
    mrgvalprep (>= 0.0.4),
    yaml
```

barrettk:
@seth127 `yaml` is already an import but I added the other two

seth127 (approved):
I had to add MPN to pkgr to get mrgvalprep and mrgvalidate, and I also dropped in a gitignore so we don't accidentally check in rendered validation docs. See [8315a8d](https://github.com/metrumresearchgroup/bbr/pull/469/commits/8315a8d0707104a8c32ab0f451e2de35e2421a5c)

Issue #472, opened 2022-04-05
test_threads() helper


seth127:
# Summary

Helper function to run (a short version of) a model several times with different numbers of threads, time each run, and report the results.

Basically we want to implement a version of what's shown in this vignette: https://metrumresearchgroup.github.io/bbr/articles/nonmem-parallel.html#testing-different-numbers-of-threads

# Specification

## Arguments

- `test_threads()`
  - model
  - threads
  - iterations
    - "cap the iterations to 10 or 20 by changing MAX_EVAL or NITER in your control stream"
  - .bbi_args
- `check_threads()`
  - a list of models
 
## Return values
- `test_threads()` returns the list of models
- `check_threads()` returns tibble of times for the models that have finished (with `NA` for the models that haven't)

## Error handling
- `check_threads()` should warn and continue if some of the models can't be parsed

barrettk:
Stories and requirements can be found [here](https://github.com/metrumresearchgroup/bbr/issues/488#issuecomment-1145358906)

PR #473, opened 2022-04-07
test_threads() helper function

barrettk:
Closes https://github.com/metrumresearchgroup/bbr/issues/472


barrettk:
FYI I know I have to delete the cleanup section. It's only there for testing at the moment. At some point Ill find an easier way to delete those files when appropriate

barrettk:
The functions can be tested as follows:
```
# Setup
devtools::load_all()
REF_DIR <- system.file("test-refs",   package = "bbr")
ABS_MODEL_DIR <- system.file("model", "nonmem", "basic",   package = "bbr")
MOD_ID <- "1"
MODEL_DIR <-   fs::path_rel(ABS_MODEL_DIR, getwd()) %>% as.character()

options('bbr.bbi_exe_path' = '/data/apps/bbi')
bbi_init(MODEL_DIR, "/opt/NONMEM", "nm74gf")

# Testing
mod <- read_model(file.path(MODEL_DIR, 1)) # this should be the copy with capped iterations

mods <- test_threads(mod, c(2, 4), .bbi_args = list(nm_version = "nm74gf"))
check_threads(mods)

cleanup_mods(mods)
```

barrettk:
@seth127 The PR is not done yet as I still have to add some tests, but the core functionality is ready for review.

`test_threads` is largely unchanged, but the other two functions have gotten some nice improvements:

## check_threads
- `check_threads` has 3 new arguments, two of which were expected
  - `wait` signifies whether to run `wait_for_nonmem`, which is something we might want to disable if the models have already finished (default is TRUE)
  - `...` additional args passed to `wait_for_nonmem`, which we definitely need to allow, as `time_limit` will need to be adjusted based on how many models they're submitting.
  - `return_times` denotes which times we want to return. The default is only `estimation_time`, but users can specify any subset of `c("estimation_time", "covariance_time", "cpu_time")`, **_or_** 'all', which is the shorthand way of specifying all 3 columns.
- `check_threads` is wrapped in a `tryCatch()`, which will return the warnings of `wait_for_nonmem` (with some additional ones). The error of not finding the lst file is avoided, and will instead print the more informative warning messages
```
> mod <- read_model(file.path(MODEL_DIR, 1)) # this should be the copy with capped iterations

> mods <- test_threads(mod, threads = c(2, 4))
Submitting 2 models with 2 unique configurations.
> check_threads(mods, wait = TRUE, time_limit = 300)
Waiting for 2 model(s) to finish...
Waiting for 2 model(s) to finish...
Waiting for 2 model(s) to finish...
2 model(s) have finished                                                                                                                                                                 
# A tibble: 2 × 2                                                                                                                                                                        
  threads estimation_time
    <dbl>           <dbl>
1       2            4.09
2       4            3.48

```
A user could then go back later and get all times:
```
check_threads(mods, wait = FALSE, return_times = c("estimation_time", "covariance_time"))
# A tibble: 2 × 3                                                                                                                                                                        
  threads estimation_time covariance_time
    <dbl>           <dbl>           <dbl>
1       2            4.04            0.29
2       4            3.36            0.37
```
_**or**_

```
check_threads(mods, wait = FALSE, return_times = "all")
# A tibble: 2 × 4                                                                                                                                                                        
  threads estimation_time covariance_time cpu_time
    <dbl>           <dbl>           <dbl>    <dbl>
1       2            4.04            0.29     9.43
2       4            3.36            0.37    13.8 
```

## cleanup_mods
-  `cleanup_mods` now allows `.tags` to be specified, in the event users have a list of models and want to delete only a subset of them. The main advantage is that the function can now be used outside of `test_threads` benchmarking.

```
> mod_tags <- lapply(mods, function(mod.x){mod.x$tags}) %>% unlist(); print(mod_tags)
[1] "test 2 threads" "test 4 threads"
> cleanup_mods(mods, .tags = mod_tags)
Removed models with the following tags:
- test 2 threads
- test 4 threads

# could also set .tags to NULL to delete all `mods` (default)
> cleanup_mods(mods)
Removed models with the following tags:
- test 2 threads
- test 4 threads
```

Im mostly curious what you think about the `return_times` argument. I like the idea of being able to get that information if it's of interest, but not sure I like specifying 'all', nor the error I get if other strings are accidentally provided (capitalized 'A' in example below):
```
 Error in check_threads(mods, wait = FALSE, return_times = "All") : 
  Assertion on 'all(return_times %in% c("estimation_time", "covariance_time", "cpu_time"))' failed: Must be TRUE. 
```

seth127:
This is all looking good. I'm putting some other comments in, but this is kind of a big thing:

I think we should add an `.maxeval` argument to `test_threads()`. This is going to make it a lot smoother to use. Generally we stay away from programmatically touching NONMEM control streams, but in this case I think it's pretty low risk and will give a lot of benefit. Plus these are "throw-away" models anyway, so it's not like we run the chance of mangling one of someone's real control streams.

```r
#' @param .maxeval Integer, defaulting to 20. Replaces the value of
#' `MAXEVAL` or `NITER` in the NONMEM control stream with this 
#' value. The purpose of this to cap the number iterations so that 
#' testing can go faster. Typically, you will want a number of iterations 
#' that takes 3-5 minutes when `threads = 1`.
```

### Notes

* Make sure we're only modifying the control streams of the newly copied models, not the original model
* Look for `MAXEVAL=[0-9]+` or `NITER=[0-9]+`. Should only find one of them. If it finds neither (or both) should error informatively.
* Can use `get_model_path()` helper to get control stream paths.

R/test-threads.R
```diff
@@ -0,0 +1,141 @@
...
+#' }
+#' @return A tibble with columns `threads` (number of threads) and `time`
+#'   (elapsed estimation time in seconds for test models).
+#'
+#' @importFrom tidyselect all_of
+#'
+#' @export
+check_threads <- function(
+  .mods,
+  return_times = "estimation_time",
```
seth127, line 69: dots in front of arg names. (Also, look through the roxygen docs and make sure this is consistent throughout).

R/test-threads.R
```diff
@@ -0,0 +1,141 @@
...
+  wait = TRUE,
+  ...
+) {
+
+  if("all" %in% return_times){
```
seth127, line 74: > Im mostly curious what you think about the return_times argument.

I think this is the right approach. I see your concern, but I'd just keep it.

R/test-threads.R
```diff
@@ -0,0 +1,141 @@
...
+#'   (elapsed estimation time in seconds for test models).
+#'
+#' @importFrom tidyselect all_of
+#'
+#' @export
+check_threads <- function(
+  .mods,
+  return_times = "estimation_time",
+  wait = TRUE,
+  ...
```
seth127, line 71: > Im also curious if you think the default for wait in check_threads should be TRUE or not

Yes, I think so. We want people to be able to do

```r
mod %>%
  test_threads() %>%
  check_threads()
```
and have it likely work.

R/test-threads.R
```diff
@@ -0,0 +1,141 @@
...
+#' Remove model files associated with the specified tags
+#'
+#' @param .mods a bbi model object or list of model objects.
+#' @param .tags a character vector identifying the tags of the models you want to delete.
+#'              If set to NULL, delete all associated model files.
+#'
+#' @importFrom checkmate check_character
+#'
+#' @export
+cleanup_mods <- function(.mods, .tags = NULL){
```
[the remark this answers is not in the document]
barrettk: Yeah I went through the same thought process in terms of not being able to pipe it anymore. I also couldn't think of an alternative unless `check_threads()` _also_ returned the model objects, which seemed odd to do. 

I also had the same thought process regarding the default. The original default was "test threads", but `grepl` wouldn't work because I made the tags `test 2 threads`, `test 4 threads`, etc (even without `fixed=TRUE` it wouldn't work because of the number). I could get rid of the number, which would make that much easier, but wasn't sure if we wanted to do that. I agree with removing `NULL` as the default, and only made it that so users were very aware that was possible. I also imagined users would likely _not_ pipe in a bunch of models, and only want to delete some, but what do I know about that.

R/test-threads.R
barrettk: Yup had the same thought

R/test-threads.R
barrettk: Oh I didnt realize we were doing that for _all_ arguments (thought it was just some of them, though I lowkey have been going back and forth with no real reason for why one should be a dot), but sounds good. The last PR had some non-dot args, should I fix those as well?

```
wait_for_nonmem <- function(.mod, time_limit = 200, interval = 5) {
  UseMethod("wait_for_nonmem")
}
```

This one definitely should have been `.mod`:
```
check_nonmem_finished <- function(mod) {
```

barrettk:
New `...` arg is passed to `submit_models`, so a dry run and other options are now available:
```
mods <- test_threads(mod, .threads = c(2, 4), .max_eval = 100, .mode = "local", .dry_run = TRUE)
```

barrettk:
`check_threads` now works with a single model object and doesnt necessarily have to correspond to `test_threads` anymore. The function could also be used to grab times from any model run. 

```
> check_threads(mods, .wait = FALSE, .return_times = "all")
# A tibble: 2 × 5                                                                                                                                                                       
  model_name  threads estimation_time covariance_time cpu_time
  <chr>         <dbl>           <dbl>           <dbl>    <dbl>
1 1_2_threads       2            0.23            0.6      1.01
2 1_4_threads       4            0.48            0.39     1.34
```

As such, im wondering if we want to rename the function name to something more general. `check_times` for instance.

barrettk:
- Previously `cleanup_mods` would error out if the dimensions of `.tags` did not match number of models (unless it was only one tag) **OR** if a specified tag wasn't found.
- `cleanup_mods` now works much smoother and will tell you which models were deleted, and which tags were not found (if any):
```
> cleanup_mods(.mods = mods, .tags = c("test threads", "hello"))
The following tags were not found:
- hello 

Removed 2 models with the following tags:
- test threads
```

```
> cleanup_mods(.mods = mods)
Removed 2 models with the following tags:
- test threads
```

```
> cleanup_mods(.mods = mods, .tags = NULL)
Removed 2 models with the following tags:
- test threads
```

barrettk:
## `cleanup_mod` patch
- Previously `cleanup_mods` would error out if the dimensions of `.tags` did not match number of models (unless it was only one tag) **OR** if a specified tag wasn't found.
- `cleanup_mods` now works much smoother and will tell you which models were deleted, and which tags were not found (if any)
- `cleanup_mods` will now prompt you before deleting the model files. You can override this functionality with the new `.force` argument

```
> cleanup_mods(.mods = mods)
Are you sure you want to remove 2 models with the following tags?: `test threads` (Yes/no/cancel) yes
Removed 2 models with the following tags:
- test threads
```

```
> cleanup_mods(.mods = mods, .tags = c("test threads", "hello"), .force = TRUE)
The following tags were not found:
- hello 

Removed 2 models with the following tags:
- test threads
```

```
> cleanup_mods(.mods = mods, .tags = c("test threads", "hello"))
The following tags were not found:
- hello 

Are you sure you want to remove 2 models with the following tags?: `test threads` (Yes/no/cancel) yes
Removed 2 models with the following tags:
- test threads
```

barrettk:
wow - I found so many bugs with `cleanup_mods` after testing with models created outside of `test_threads()`. Glad we're writing tests before handing this out

tests/testthat/test-workflow-bbi.R
```diff
@@ -205,7 +205,47 @@ withr::with_options(list(
...
+    expect_output(str(run_times), " $ cpu_time", fixed = TRUE)
+
+    run_times <- check_run_times(list(mod1, mod2), .wait = FALSE,
+                                 .return_times = c("estimation_time", "covariance_time"))
+    expect_output(str(run_times), "tibble [2 × 4]", fixed = TRUE)
+    expect_output(str(run_times), " $ model_run", fixed = TRUE)
+    expect_output(str(run_times), " $ threads", fixed = TRUE)
+    expect_output(str(run_times), " $ estimation_time", fixed = TRUE)
+    expect_output(str(run_times), " $ covariance_time", fixed = TRUE)
+  })
```
barrettk, line 242: @seth127 let me know what you think about this style of testing. There are perhaps better ways to test this, but this seemed the most comprehensive (though sort of looks dumb? - maybe just me)

seth127: I don't have a strong reason for this, but I would prefer doing something like:

```r
expected_cols <- c("model_run", "threads", "estimation_time", "covariance_time", "cpu_time")

expect_true(all(res_cols %in% names(run_times)))
expect_equal(dim(run_times), c(1, 5))

...
```

Like I said, I don't have a really strong reason to prefer that style, but relying on `str()` printing and doing string matching on it just feels kind of awkward and brittle.

tests/testthat/test-test-threads.R
```diff
@@ -0,0 +1,222 @@
...
+    msg_remove <- paste0(
+      paste("Removed", length(mods), "models with the following tags:\n"),
+      paste("-","test threads", collapse = "\n")
+    )
+    expect_message(
+      cleanup_mods(mods, .force = T),
+      msg_remove
+    )
+
+    for(i in 1:length(mod_ctls)){
```
seth127, line 133: `file_exists()` is vectorized, so you can just do

```r
expect_false(any(fs::file_exists(mod_ctls)))
expect_false(any(fs::file_exists(mod_yamls)))
```

and not do the `for` loop at all.

Same for a few places further down in this file.

tests/testthat/test-test-threads.R
```diff
@@ -0,0 +1,222 @@
+context("test_threads(.dry_run=T)")
+
+###################################
+# testing single model submission
+###################################
+
+
+# define constants
```
seth127, line 8: These tests all look good functionally, but at some point we should go back and refactor them (and probably some other ones, for example in the `test-workflow-bbi.R`) to be more self-contained and use setup helpers like [this one](https://github.com/metrumresearchgroup/bbr/blob/a9c8b90432bbeef88ab3a9f15dc1b6a8e22e626a/tests/testthat/setup-workflow-ref.R#L197-L202) that @rymarinelli is adding in #484

R/test-threads.R
```diff
@@ -0,0 +1,268 @@
...
+    return(NA)
+  }, warning = function(cond){
+    message(cond)
+    message("\nConsider setting/increasing the `time_limit` in `wait_for_nonmem()`. See ?check_run_times for details")
+    return(NA)
+  }
+  )
+}
+
+#' Remove model files associated with the specified tags
```
seth127, line 167: I'm sorry to say this, but I really think we should rename this to `delete_models()`. The "cleanup" name made sense when it was being used for testing, but now that it's exported, I think `delete` is much clearer. Also, all of the other functions spell out "model(s)", so let's keep that consistent.

tests/testthat/test-workflow-bbi.R
```diff
@@ -184,7 +184,7 @@ withr::with_options(list(
     # create model
     mod1 <- read_model(file.path(MODEL_DIR_BBI, "1"))
     submit_model(mod1, .mode = "local", .wait = FALSE)
-    wait_for_nonmem(mod1, 100, interval = 5)
```
seth127, line 187: These tests were failing, but I went ahead and committed the fix in d5f6d9d7af6988c8b172ead4b4db3b3185505afc

We probably missed them because they don't run in CI (because we don't have a NONMEM license on the Drone server).

R/test-threads.R
```diff
@@ -0,0 +1,268 @@
...
+      if (fs::dir_exists(m)) fs::dir_delete(m)
+    }
+    message(msg_remove)
+  }else{
+    msg_prompt <- paste0(
+      paste("Are you sure you want to remove", length(mod_paths), "models with the following tags?: "),
+      paste0("`",mods_removed,"`", collapse = ", ")
+    )
+    delete_prompt <- askYesNo(msg_prompt)
+    if(delete_prompt){
```
seth127, line 255: this needs to `if (isTRUE(delete_prompt))` because otherwise answering "cancel" throws this:

```
Error in if (delete_prompt) { : missing value where TRUE/FALSE needed
```

seth127 (dismissed):
Mostly looking good, but a bunch of small comments.

barrettk:
FYI @seth127 :
```r
  if (inherits(.mods, "bbi_model")) {
    .mods <- list(.mods)
  }
  check_model_object_list(.mods, .mod_types = c(NM_MOD_CLASS, NM_SUM_CLASS))
```

wont work for a list of summary objects. I think we will have to change something here:
```r
> class(sum_mods)
[1] "bbi_summary_list" "list"            
> NM_SUM_CLASS
[1] "bbi_nonmem_summary"
> NM_SUM_CLASS_LIST
[1] "bbi_nonmem_summary" "bbi_model"          "list"  
```

barrettk:
@seth127 Oh sorry I meant that `check_run_times` and `delete_models` were not set up to work with summary objects, which is why I wasnt sure why we were adding it to those functions specifically.

seth127:
@barrettk it's kinda of a judgement call and I always default to thinking about the UX: is this a function that someone is likely only going to use on a model object (like, before they submit/run it) or might they want to use it afterwards when they're looking at model outputs too? If the latter, then it's nice if it works for summary objects too.

`delete_models()` I'm kinda on the fence about. I guess I could see just making it work with model objects, especially because we want people to be careful with it. And because it's related to tags (which the summary object doesn't carry around by default, though it could go look them up if it needed to).

`check_run_times()` might actually be nice if it worked with the summary objects too. This is clearly something people will be using _after_ the model completes. You could also have it be able to take the `bbi_summary_list` object (created [here](https://github.com/metrumresearchgroup/bbr/blob/main/R/classes.R#L90-L112)) that is returned from [`model_summaries()`](https://github.com/metrumresearchgroup/bbr/blob/develop/R/model-summaries.R#L85). 

I don't know... I'll leave either of these up to your judgments. I don't think it's primary functionality for either one, but I do think it could be nice. Mostly just think through things people might plausibly try to do and have them either work or error gracefully.

barrettk:
@seth127 `check_run_times` now works for summary objects as well:

```r
> mods <- test_threads(.mod = MOD1, .threads = c(2,4), .mode = "local")
Submitting 2 models with 2 unique configurations.

> check_run_times(mods, .wait = FALSE, .return_times = "all")
# A tibble: 2 × 5                                                                                                                                                                       
  model_run   threads estimation_time covariance_time cpu_time
  <chr>         <dbl>           <dbl>           <dbl>    <dbl>
1 1_2_threads       2            0.57            0.22     1.02
2 1_4_threads       4            0.34            0.38     1.42

> check_run_times(MOD1, .wait = FALSE, .return_times = "all")
# A tibble: 1 × 5                                                                                                                                                                       
  model_run threads estimation_time covariance_time cpu_time
  <chr>       <dbl>           <dbl>           <dbl>    <dbl>
1 1               4            5.28            0.46     6.82

> model_summary(MOD1) %>% check_run_times(.wait = FALSE, .return_times = "all")
# A tibble: 1 × 5                                                                                                                                                                       
  model_run threads estimation_time covariance_time cpu_time
  <chr>       <dbl>           <dbl>           <dbl>    <dbl>
1 1               4            5.28            0.46     6.82

> model_summaries(mods) %>% check_run_times(.wait = FALSE, .return_times = "all")
# A tibble: 2 × 5                                                                                                                                                                       
  model_run   threads estimation_time covariance_time cpu_time
  <chr>         <dbl>           <dbl>           <dbl>    <dbl>
1 1_2_threads       2            0.57            0.22     1.02
2 1_4_threads       4            0.34            0.38     1.42
```

seth127:
That looks great @barrettk and I agree with you on `delete_models()`. I think these are the next steps here:

* Add tests
* I'll do a final review and then we can merge to `develop`
* Tag as a dev release and find some folks to play with this. If we have significant feedback, we'll open up a new patch branch for changes before we actually release.

seth127 (approved):
Looks great. All tests pass for me. Nice work.

Issue #479, opened 2022-04-18
wait_from_nonmem() helper


seth127:
# Summary

There are situations where we want to monitor whether a NONMEM run has completed. Note that we don't care (in these situations) whether it finished _successfully_ in any sense, just whether it is still running. This is not trivial because these models are often submitted to run on worker nodes in a grid, so we can't just check for the process ID.

# Notes

Initial idea is to check the `.lst` file for "Stop Time". This seems to work for most cases, but there are some things to consider.

* Need to check whether this works if you kill a model
  * submitted locally and killed with process ID `kill -9 <pid>`
  * submitted to SGE and killed with `qdel`
* Need to check if it works in all NONMEM failures
  * if it dies early (i.e. can't find the data set) does it still print "Stop Time"?

# Requirements

* Should be able to take multiple models
* Should be a helper `check_nonmem_finished()` (also exported) to check a single model one-off
* Should take an arg for interval to check
* Should take an arg for total time limit (so this doesn't hang indefinitely)

barrettk:
Stories:
```
MGMT-S019:
  name: wait_for_nonmem
  description: As a user, I want to be able to freeze my R console while a model is executing on the grid or otherwise
  ProductRisk: Medium
  Requirements:
     - MGMT-R010
```

Requirements:
```
MGMT-R010:
  description: wait_for_nonmem() will freeze the console until a stop time is found or the allotted time is reached
  tests: 
     - BBR-UTL-012
     - BBR-UTL-013

```

PR #480, opened 2022-04-18
'wait_for_nonmem()' helper functions

barrettk:
# Summary

There are situations where we want to monitor whether a NONMEM run has completed. Note that we don't care (in these situations) whether it finished successfully in any sense, just whether it is still running. This is not trivial because these models are often submitted to run on worker nodes in a grid, so we can't just check for the process ID.

# Requirements

* Should be able to take multiple models
* Should be a helper `check_nonmem_finished()` (also exported) to check a single model one-off
* Should take an arg for interval to check
* Should take an arg for total time limit (so this doesn't hang indefinitely)

closes https://github.com/metrumresearchgroup/bbr/issues/479
(see issue for more details)


barrettk:
@seth127 let me know what you think of those tests, as well as the new helper functions. The new functions may be overkill and/or weren't necessary, but I can see other possible use cases for them

barrettk:
If we decide against putting those functions in bbr, we could add them to `metrumrg` or other location (I know everyone wants to stay away from that package), and then import them in to use in that test I referenced. 

TBH Im not sure if they belong in bbr, given that they're related to fetching/ending jobs and are not unique to modeling, but I also couldnt think of another location besides `metrumrg`.

R/utils.R
```diff
@@ -504,3 +504,175 @@ download_with_retry <- function(...) {
...
+#' @describeIn wait_for_nonmem takes a `list` of `bbi_nonmem_model` objects.
+#' @export
+wait_for_nonmem.list <- function(mod, time_limit = 200, interval = 5) {
+
+  assert_list(mod)
+
+  expiration <- Sys.time() + time_limit
+  Sys.sleep(1) # wait for lst file to be created
+
+  while ((expiration - Sys.time()) > 0) {
```
seth127, line 598: Let's add a warning when the expiration gets triggered. In other words, if it goes all the way through and it still doesn't look like the model has finished, we should let the user know.

Then we can also check for this warning in your failed run test.

Sidenote: this probably means you'll have to change this from a `while` to a `for`, and put an `if((expiration - Sys.time()) > 0) ...` in there, but whatever works is fine.

R/utils.R
```diff
@@ -504,3 +504,175 @@ download_with_retry <- function(...) {
...
+    if (all(res)) {
+      break
+    }
+    Sys.sleep(interval)
+  }
+}
+
+
+
+#' R wrapper for `qstat`
```
seth127, line 609: As discussed offline, let's not do these two functions (`fetch_model_runs()` and `crash_model_run()`) for now. They are a bit out of scope for current `bbr`, and they will be pretty fragile, in terms of only working in specific situations when using and SGE grid. 

It would be good to drop this code into an issue though. We may want to come back to something like this later.

tests/testthat/test-workflow-bbi.R
```diff
@@ -177,5 +177,39 @@ withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
...
+
+  test_that("wait_for_nonmem() correctly reads in stop time [BBR-UTL-012]", {
+    # create model
+    mod1 <- read_model(file.path(MODEL_DIR_BBI, "1"))
+    submit_model(mod1, .mode = "local", .wait = FALSE)
+    wait_for_nonmem(list(mod1), 100, interval = 5)
+    expect_true(suppressMessages(!rlang::is_empty(read_table(file.path(MODEL_DIR_BBI, 1, "1.tab"), skip = 1))))
+  })
+
+  test_that("wait_for_nonmem() doesn't error out if no stop time found [BBR-UTL-013]", {
```
seth127, line 189: Let's simplify this test because we don't want to use SGE in our test suite, in part because it makes things like CI more of a pain. Plus it's not really within `bbr`'s purview to test whether SGE is hooked up right (only that we pass the right thing to it, and we test that in `bbi`).

Try this instead (after you've added the warning I mention above)

```r
test_that("wait_for_nonmem() runs to time_limit if no stop time found [BBR-UTL-013]", {
    # model setup

    mod_fail <- copy_model_from(
      read_model(file.path(MODEL_DIR_BBI, "1")),
      "failure"
    )
    .p <- submit_model(mod_fail, .mode = "local", .wait = FALSE)
    .p$process$kill()

    #wait_for_nonmem(mod_fail, 3, interval = 1) # hopefully this can go away now?

    expect_warning(
      wait_for_nonmem(mod_fail, 2, interval = 1), # dont need high wait time since we know it failed
      "[whatever 'your model didnt finish' warning we're going to add]"
    )
  })
```

R/utils.R
```diff
@@ -504,3 +504,175 @@ download_with_retry <- function(...) {
...
+#'
+#' @importFrom purrr map_lgl
+#' @importFrom checkmate assert_list
+#'
+#'
+#' @describeIn wait_for_nonmem takes a `list` of `bbi_nonmem_model` objects.
+#' @export
+wait_for_nonmem.list <- function(mod, time_limit = 200, interval = 5) {
+
+  assert_list(mod)
```
seth127, line 593: Probably want to throw in a [`check_model_object_list(mod)`](https://github.com/metrumresearchgroup/bbr/blob/main/R/utils.R#L385-L395) (as used in [`submit_models()`](https://github.com/metrumresearchgroup/bbr/blob/develop/R/submit-models.R#L44) for example) just to be safe.

R/utils.R
```diff
@@ -504,3 +504,175 @@ download_with_retry <- function(...) {
...
+  }
+
+  # pass .lst path to character dispatch
+  build_path_from_model(mod, ".lst") %>%
+    check_nonmem_finished()
+}
+
+#' Wait for NONMEM models to finish
+#' @param mod a `bbi_nonmem_model` object, or list of `bbi_nonmem_model` objects.
+#' @param time_limit integer for maximum number of seconds in total to wait before continuing (exiting function)
```
seth127, line 556: Maybe add something to clarify "even if the run does not appear to have finished"

R/utils.R
```diff
@@ -504,3 +504,175 @@ download_with_retry <- function(...) {
   }
   return(rc)
 }
+
+#' Checks if NONMEM run is done by looking for "Stop Time" in .lst file
+#'
+#' Returns `TRUE` if the model appears to be finished running and `FALSE` otherwise.
+#' @param mod either a bbi_nonmem_model object or a path to an .lst file
+#' @export
+check_nonmem_finished <- function(mod) {
```
seth127, line 513: The more I think about it, I _don't_ think `check_nonmem_finished()` should be S3 dispatch that can optionally take the `.lst` path. I think it should only take the model object. You can just combine these into a single function and call [check_model_object(.mod, NM_MOD_CLASS)](https://github.com/metrumresearchgroup/bbr/blob/main/R/utils.R#L370-L379) at the top.

The reason is that almost all of our other functions insist on taking the model object, which is a decision we made early on to avoid path confusion ("do I pass the path to the YAML, to the .lst, I can't remember?...").

And to be clear, I'm just talking about `check_nonmem_finished()`. I think the S3 implementation of `wait_for_nonmem()` is a good idea.

seth127 (changes requested):
Good stuff. As discussed, we're going to get rid of the SGE functions for now. Some other comments too.

barrettk:
FYI
image

R/utils.R
```diff
@@ -578,101 +562,39 @@ wait_for_nonmem.bbi_nonmem_model <- function(mod, time_limit = 200, interval = 5
...
-#'
-#' @export
-crash_model_run <- function(run_name){
-
-  # warning message occurs if slots are NA (only jobs in the queue)
-  runs <- suppressWarnings(fetch_model_runs())
-
-  if(is_empty(runs)){
-    stop("No jobs running at this time")
+  if(expiration < Sys.time() & !map_lgl(mod, ~check_nonmem_finished(.x))){
```
seth127, line 591: You're gonna want `&&` here. See explanation [here](https://stackoverflow.com/a/6559049/3743213) for details (essentially we only want this to return a single boolean, not a vector of them).

But also, this causes a warning whenever you pass more than one model:

```
Warning message:
In if (expiration < Sys.time() & !map_lgl(mod, ~check_nonmem_finished(.x))) { :
  the condition has length > 1 and only the first element will be used
```

R/utils.R
```diff
@@ -504,3 +504,97 @@ download_with_retry <- function(...) {
...
+
+#' Checks if NONMEM run is done by looking for "Stop Time" in .lst file
+#'
+#' Returns `TRUE` if the model appears to be finished running and `FALSE` otherwise.
+#' @param mod either a bbi_nonmem_model object or a path to an .lst file
+#'
+#' @importFrom readr read_lines
+#'
+#' @export
+check_nonmem_finished <- function(mod) {
```
seth127, line 516: As much as I hate it, you need to add the leading period to this arg (i.e. make it `.mod` instead of `mod`) for consistency's sake. I wish we hadn't committed to this pattern, but at this point there's not really an easy way out of it. And it's better to be annoying and consistent than annoying and inconsistent.

Same is true for all the arguments to `wait_for_nonmem()`.

seth127 (approved):
Two minor comments, but otherwise looks good. I tested the messaging and it looks nice:

```
> submit_model(mod1, .mode = "local", .wait = FALSE)
> Sys.sleep(5)
> submit_models(list(mod2, mod3), .mode = "local", .wait = FALSE)
Submitting 2 models with 2 unique configurations...
> wait_for_nonmem(list(mod1, mod2, mod3), 5, interval = 0.01)
Waiting for 3 model(s) to finish...
Waiting for 2 model(s) to finish...                                                                                                                           
Waiting for 2 model(s) to finish...     
Waiting for 2 model(s) to finish...     
Warning: Expiration was reached, but 2 model(s) haven't finished                                                                                              
> wait_for_nonmem(list(mod1, mod2, mod3), 15, interval = 0.01)
Waiting for 3 model(s) to finish...
Waiting for 2 model(s) to finish...                                                                                                                           
Waiting for 2 model(s) to finish...                                                                                                                           
3 model(s) have finished   
```

R/utils.R
barrettk: Good call here. I've never observed a difference in results between the two (despite knowing the differences), but this is the first example I've seen where its absolutely warranted

seth127 (approved):
nice work

Issue #477, opened 2022-04-13
filter argument to `run_log()` and friends


seth127:
As a user, I would like to be able to pass a vector of run ID's (i.e. the values that show up in the `run` column in the resulting tibble) to any of the `*_log()` functions and have the resulting tibble filtered to only runs in that vector.

Importantly, this filtering should be done _before_ the associated runs are parsed. This won't matter much for `run_log()` but for something like `summary_log()` it would be inefficient to parse all the runs and _then_ filter them.

PR #487, opened 2022-05-10
Creating Additional Column for Starring 

### LINKED ISSUE #474: Starred models
# Summary

As a user, I want to be able to star certain models to indicate special interest level

@riggsmm

rymarinelli:
closes #474


rymarinelli:
@kylebaron @seth127 Do you think this should be ordinal in nature and should be integrated into the filtering from #477 ?

kyleam:
> But I am not sure where to call `reconcile_yaml`

Shouldn't  the `*_star` wrappers behave in the same way as any of the other `modify_model_field`-based wrappers?  Those let `modify_model_field` call `check_yaml_in_sync` and error if the in-memory and disk values aren't in sync, pointing the user to `reconcile_yaml`.

seth127:
> Shouldn't the *_star wrappers behave in the same way as any of the other modify_model_field-based wrappers? 

Short answer: yes

Also, I glanced at this earlier and had this question: what does `replace_star()` do? In the case of `replace_tag()` it's "I want to replace this tag with this new tag" but I think for stars it's either you have it or you don't, so I _think_ `add_star()` and `remove_star()` should cover it.

kyleam:
@seth127 

> Also, I glanced at this earlier and had this question: what does `replace_star()` do? [...]

Ah, I was just wondering the same thing.

R/copy-model-from.R
```diff
@@ -22,6 +22,7 @@
 #'   and ancestry. **Paths must be relative to `.new_model` path.** Note that
 #'   the `.parent_model` will automatically be added to the `based_on` field, so
 #'   no need to include that here.
+#' @param .starred boolean variable to denote significance
```
kyleam, line 25: Might just be me, but I'd suggest avoiding "denote significance", particularly in a model context.  I liked @kylebaron's phrasing in gh-474: "star certain models to indicate special interest level".  Since "star" is in the parameter name, perhaps "mark model to indicate special interest level".

Also, what do you think about`.star` rather than `.starred`?  That'd be shorter and seems to capture the intent well enough ("star this model when you copy it").

R/copy-model-from.R
```diff
@@ -167,6 +173,7 @@ copy_nonmem_model_from <- function(
     .description = .description,
     .based_on = c(.parent_based_on, .based_on_additional),
     .tags = new_tags,
+    .starred = ,
```
kyleam, line 176: unintentionally missing value?

rymarinelli:
@seth127 I used `replace_star` to try and follow the naming conventions. It seemed most of the functions were named `replace_{field_name}` Should not I follow this then?

R/modify-model-field.R
```diff
@@ -31,8 +31,9 @@
 #' @param .remove If `TRUE`, `.value` with be removed from the `.field` instead of added. `FALSE` by default. Cannot have both `.append` and `.remove` be true in the same call.
 #' @param .unique If `TRUE`, the default, de-duplicate `.mod[[.field]]` after adding new values. If `FALSE` duplicate values will be kept.
 #' @param .char_value If `TRUE`, check that `.value` (after unlisting) is a character vector.
+#' @param .bool_value If `TRUE`, check that `.value` (after unlisting) is a boolean vector.
 #' @export
-modify_model_field <- function(.mod, .field, .value, .append = TRUE, .remove = FALSE, .unique = TRUE, .char_value = TRUE) {
+modify_model_field <- function(.mod, .field, .value, .append = TRUE, .remove = FALSE, .unique = TRUE, .char_value = TRUE, .bool_value = FALSE) {
```
kyleam, line 36: Hmm, that's getting a bit unwieldy, and in my view it's not ideal having dependent options like `.bool_value` and `.char_value`, rather than rolling them into one.  On the other hand, this is exported, so we'd have to change this in a backward-compatible way.  That'd be a multi-step process to do that, and that doesn't seem worth it given that this function has more user-friendly wrappers on top.

So I think your choice of adding a new parameter is the way to go.

R/modify-model-field.R
```diff
@@ -46,6 +47,11 @@ modify_model_field <- function(.mod, .field, .value, .append = TRUE, .remove = F
     checkmate::assert_character(.value, null.ok = TRUE)
   }
 
+  if (isTRUE(.bool_value)) {
+    .value <- unlist(.value)
+    checkmate::assertLogical(.value, null.ok = TRUE)
```
kyleam, line 52: is NULL okay?

R/modify-model-field.R
```diff
@@ -164,7 +170,8 @@ replace_all_tags <- function(.mod, .tags) {
     .field = YAML_TAGS,
     .value = .tags,
     .append = FALSE,
-    .char_value = TRUE
+    .char_value = TRUE,
+    .bool_value = FALSE
```
kyleam, line 174: Specifying optional args is nice for guarding against flips in the default, but that is very unlikely to happen in this case, and you don't do it for some other `modify_model_field` calls.  I'd suggest not bothering adding the `.bool_value = FALSE` to any functions.

R/modify-model-field.R
```diff
@@ -178,6 +185,72 @@ remove_tags <- function(.mod, .tags) {
...
+  )
+}
+
+
+
+#' @name add_star
+#' @title Star a model
+#'
+#' @description Adds star  on a model object and corresponding YAML.
+#' Tags can be modified at any time (i.e. before _or_ after a model is submitted).
```
kyleam, line 198: Is this "Tags ..." bit leftover copy-paste cruft?

R/modify-model-field.R
```diff
@@ -178,6 +185,72 @@ remove_tags <- function(.mod, .tags) {
...
+#'
+#' @details
+#' The `starred` fields on a `bbi_{.model_type}_model` object contains a boolean variable if model is starred
+#'
+#'
+#' @return The modified `bbi_{.model_type}_model` object
+#'
+#' @seealso [run_log()] [collapse_to_string()] [modify_notes()] [modify_based_on()] [modify_description()] [modify_bbi_args()]
+#' NULL
```
kyleam, line 207: spurious NULL (this shows up in the docstring)

R/modify-model-field.R
```diff
@@ -178,6 +185,72 @@ remove_tags <- function(.mod, .tags) {
     .append = FALSE,
     .remove = TRUE,
     .char_value = TRUE,
+    .bool_value = FALSE
+  )
+}
+
+
+
+#' @name add_star
```
kyleam, line 194: There are a few ways to share docs for multiple function, but for consistency, take a look at how the other wrapper functions (e.g., `add_notes` and `remote_notes`) in the file do it.  If you do it that way, you want to give this docstring a more generic name and put it above `NULL` and then add docstrings to `add_star` and `remove_star` that use `@describeIn`.

R/modify-model-field.R
```diff
@@ -178,6 +185,72 @@ remove_tags <- function(.mod, .tags) {
...
+#' @title Star a model
+#'
+#' @description Adds star  on a model object and corresponding YAML.
```
kyleam, line 197: extra space between "star" and "on"

R/modify-model-field.R
```diff
@@ -178,6 +185,72 @@ remove_tags <- function(.mod, .tags) {
...
+#'
+#'
+#' @return The modified `bbi_{.model_type}_model` object
+#'
+#' @seealso [run_log()] [collapse_to_string()] [modify_notes()] [modify_based_on()] [modify_description()] [modify_bbi_args()]
+#' NULL
+#' @inheritParams modify_model_field
+#' @param .starred boolean variable to indicate significance
+#' @export
+add_star <- function(.mod, .starred) {
```
kyleam, line 211: What's the point in having the `.starred` parameter?  Is `.starred = TRUE` rolled up in the **add**_star name?  (same for `remove_star` and `.starred = FALSE`)

R/modify-model-field.R
```diff
@@ -178,6 +185,72 @@ remove_tags <- function(.mod, .tags) {
...
+    .append = FALSE,
+    .char_value = FALSE,
+    .bool_value = TRUE
+  )
+}
+
+#' @title replace_star
+#' @param .starred boolean variable to indicate significance
+#' @param .mod list object of model data
+#' @describeIn Add or remove star on a model object and corresponding YAML.
```
kyleam, line 226: I don't think this `@describeIn` is wired up properly.  The first value should point to a name where the actual docs are.  See what `remove_notes` does for an example.

R/modify-model-field.R
```diff
@@ -178,6 +185,72 @@ remove_tags <- function(.mod, .tags) {
...
+#' Tags can be modified at any time (i.e. before _or_ after a model is submitted).
+#' @export
+replace_star <- function(.mod, .starred) {
```
kyleam, line 229: Like Seth, I don't see the purpose of having this function.

R/modify-model-field.R
```diff
@@ -178,6 +185,72 @@ remove_tags <- function(.mod, .tags) {
...
+    .mod = .mod,
+    .field = YAML_STAR,
+    .value = .starred,
+    .append = FALSE,
+    .bool_value = TRUE,
+    .char_value = FALSE
+  )
+}
+
+#' @title remove_star
```
kyleam, line 240: I mostly commented on this above, but `remove_star` should be described in the same place as `add_star`.

inst/model/nonmem/test-workflow-bbi-models/1.yaml
```diff
@@ -0,0 +1,10 @@
+model_type: nonmem
```
kyleam, line 1: This file doesn't looked use by the tests.

tests/testthat/test-modify-model-field.R
```diff
@@ -354,3 +354,50 @@ test_that("remove_tags() works correctly [BBR-MMF-022]", {
   )
   expect_identical(new_mod[[YAML_TAGS]], ref_tags)
 })
+
+test_that("Checking that the star_add function is adding boolean data to YAML[BBR-MMF-023]", {
```
kyleam, line 358: * missing space before test tag
  * `star_add` -> `add_star`
  * You are using the same test tag/id for multiple tests.  This is fine for related tests like these, and means they are rolled up into a single description for the validation docs.  However, be careful because all the non-shared parts of the description are lopped off.  In this case it'd be "Checking that the star_" or Checking that the" (I need to look at the code to see if it's a word-based split).  In either case, that's pretty uninformative for a test description and will look incomplete to anyone reading the validation docs.  For this particular case, I'd say just use different test tags (e.g., "add_star() stars a model [BBR-MMF-023]" and "remove_star() unstars a model [BBR-MMF-024]").

tests/testthat/test-modify-model-field.R
```diff
@@ -354,3 +354,50 @@ test_that("remove_tags() works correctly [BBR-MMF-022]", {
...
+  # make a model object and replace the tags
+  new_mod <- read_model(temp_mod_path)
+  new_mod <- add_star(new_mod, TRUE)
+
+
+  new_mod <- remove_star(new_mod, FALSE) %>% suppressWarnings()
+  expect_true(new_mod$starred)
+
+  new_mod <- remove_star(new_mod, TRUE)
+  expect_failure(expect_true(new_mod$starred))
```
kyleam, line 383: `expect_failure(expect_true(...))` -> `expect_false(...)`?

tests/testthat/test-modify-model-field.R
```diff
@@ -354,3 +354,50 @@ test_that("remove_tags() works correctly [BBR-MMF-022]", {
...
+  expect_true(new_mod$starred)
+
+
+})
+
+test_that("Checking that the star_remove function is removing boolean data from YAML[BBR-MMF-023]", {
+  temp_mod_path <- create_temp_model()
+
+
+  # make a model object and replace the tags
```
kyleam, line 374: Leftover from copy-paste?

tests/testthat/test-modify-model-field.R
```diff
@@ -354,3 +354,50 @@ test_that("remove_tags() works correctly [BBR-MMF-022]", {
...
+  new_mod <- read_model(temp_mod_path)
+  new_mod <- add_star(new_mod, TRUE)
+
+
+  new_mod <- remove_star(new_mod, FALSE) %>% suppressWarnings()
```
kyleam, line 379: What's the warning here?

kyleam (changes requested):

kyleam:
> Also, I wanted your thoughts if I needed additional tests for `copy_model_field` and `modify_model_field`

I think `modify_model_field` is tested well enough by the existing `add_star` and `remove_star` tests.  For `copy_model_field`, either adding a new test to `test-copy-model-from.R` or modifying `copy_from_model options work [BBR-CMF-002]` sounds like a good idea to me.

inst/model/nonmem/test-workflow-bbi-models/1.yaml
rymarinelli: I think I changed this so when it was copied it would have a value under starred.

seth127:
> I used replace_star to try and follow the naming conventions. It seemed most of the functions were named replace_{field_name} Should I not follow this then?

Generally yes, that's the right move. And you do that with `add_star()` and `remove_star()`. I just don't think `replace_star()` is necessary or really makes sense. What would you be replacing it with?

R/modify-model-field.R
rymarinelli: If we wanted to add new data types to the model fields, would we have to do this in the future as well? For instance, would we have to include a `.numeric_value` if we wanted to report additional model coefficients?

R/modify-model-field.R
rymarinelli: Do you have any thoughts what the default should be? Do you think it makes more sense for it to be NULL or FALSE? I was thinking it if was set to NULL, the user would know they had not reviewed the model yet.

R/modify-model-field.R
rymarinelli: I created this function to follow the conventions. I was mostly looking around line 68 at this code chunk. 
` if (!is.null(.description)) .mod <- replace_description(.mod, .description)
  if (!is.null(.tags))        .mod <- replace_all_tags(.mod, .tags)
  if (!is.null(.bbi_args))    .mod <- replace_all_bbi_args(.mod, .bbi_args)
  if (!is.null(.based_on))    .mod <- replace_all_based_on(.mod, .based_on)
  if (!is.null(.starred))     .mod <- replace_star(.mod, .starred)
  return(.mod)`
Would it be alright to use like add_star instead?

kyleam: I see, so here:
https://github.com/metrumresearchgroup/bbr/blob/74baeb15d44eab79fdd110fb81626fbea9212b50/R/new-model.R#L37-L72

> Would it be alright to use like add_star instead?

Hmm, so my understanding is that `.starred = NULL` means "leave as is".  So, yeah, one option would be: if `.starred = TRUE`, call `add_star(.mod)` and if `.starred = FALSE` call `remove_star(.mod)`.  However, my vote would be to call `modify_model_field` directly, basically inlining the current body of `replace_star`. 

I don't think the details matter too much, though, as long as `replace_star` isn't *exported* (because the above is an internal implementation detail, and I can't think of when a user would want to call `replace_start`).

R/modify-model-field.R
kyleam: Right.  I don't know how likely it is that the need for another type would come up, but, if it did, the current approach would mean adding a new argument.  

The main other option is consolidating all the type-related arguments into a single one, but then we have to either 1) worry about adding a temporary compatibility layer, announcing the deprecation, and then removing the old args in a later release or 2) decide we're not worried about compatibility here.

Given those options, I'd say punt on consolidating it into a single argument.

R/modify-model-field.R
kyleam: > Do you have any thoughts what the default should be?

It's not clear to me that there is a scenario where a `NULL` would reach `modify_model_field`.

As you have it now, everything that has a `.starred = NULL` default (`copy_model_from`, `copy_model_from.bbi_nonmem_model`, `copy_nonmem_model_from`) ultimately goes through `new_model`, and that only sends through values that are not `NULL`.

https://github.com/metrumresearchgroup/bbr/blob/74baeb15d44eab79fdd110fb81626fbea9212b50/R/new-model.R#L72

So, I think `NULL` makes sense as a default "leave it be" value for `new_model` and friends, but I don't see why `modify_model_field` needs to be aware of that (or allow it with `null.ok = TRUE`).

This is somewhat related to [discussion elsewhere in this PR](https://github.com/metrumresearchgroup/bbr/pull/487#discussion_r880627005), but here's the (exported) interface that seems most natural to me:

```
copy_model_from(..., .starred = NULL)
copy_model_from.bbi_nonmem_model(..., .starred = NULL)
new_model(..., .starred = NULL)

add_star <- function(.mod)
remove_star <- function(.mod)
```

tests/testthat/test-modify-model-field.R
rymarinelli: `Warning message: In modify_model_field(.mod = .mod, .field = YAML_STAR, .value = .star,  :   `star` does not contain any of the following, so they cannot be removed: FALSE`

tests/testthat/test-modify-model-field.R
rymarinelli: I was getting logical(0). It is not FALSE or TRUE.

tests/testthat/test-modify-model-field.R
kyleam: Thanks.

If you're testing that the expected warning is given, it'd be clearer to use `expect_warning()`.  Though, as I mention elsewhere, I'm confused by the `remove_star(mod, FALSE)` thing and think it should just be `remove_star(mod)`, so this test becomes irrelevant if you go that direction.

R/modify-model-field.R
kyleam: To come back to this, I guess the state I'm describing is that the user can only set the star to `FALSE` or `TRUE`, but not return it to its unset factory default (`NULL`).  That feels okay to me both `mod$starred` of `NULL` or `FALSE` both mean the model is unstarred, so I don't see why the caller would care _how_ we unstar it.

tests/testthat/test-modify-model-field.R
rymarinelli: I think it may be better to consolidate add star and remove star into one function. It might be best to rename replace_start with set_star_status to better reflect the goal of the function.

kyleam: It feels like this comment is in the wrong thread, but I think this is in reply to my suggestion in a few other spots that the `modify_model_field` wrappers for star could be

```
add_star(.mod)
remove_star(.mod)
```

> It might be best to rename replace_start with set_star_status to better reflect the goal of the function

Hmm, I don't know, it seems nice to me to stick with the `remove_` and `add_` used by the other `modify_model_field` wrappers.  But I'm perhaps not the best person to say.  It might be worth seeing what @seth127 thinks.

tests/testthat/test-copy-model-from.R
```diff
@@ -61,6 +61,21 @@ test_that("copy_from_model options work [BBR-CMF-002]", {
...
+
+
+  #Editing YAML to add star status is true from clean model
+  temp_mod_path <- create_temp_model()
+  new_mod <- read_model(temp_mod_path)
+  new_mod <- set_star_status(new_mod, TRUE)
+
+  #Checking that star is written out to the yaml
+  model_yml <- paste0(temp_mod_path, ".yaml") %>% read_yaml() %>% names()
+  expect_equal( model_yml[5],"star")
```
kyleam, line 73: I'm not good about this one myself, but `expect_equal` -> `expect_identical` would be better.

R/modify-model-field.R
kyleam: * nit: document parameters in same order as function signature
 * spurious `describeIn`

R/modify-model-field.R
```diff
@@ -177,7 +183,23 @@ remove_tags <- function(.mod, .tags) {
...
+    .char_value = TRUE
+  )
+}
+
+#' @title  set_star_status
+#' @param .star boolean variable to indicate significance
+#' @param .mod list object of model data
+#' @describeIn Add or remove star on a model object and corresponding YAML.
+#' @export
+set_star_status <- function(.mod, .star) {
```
kyleam, line 195: I see the appeal of having a single function, but given the other `modify_model_field` functions, but I'd vote for `remove_star(.mod)` and `add_star(.mod)`, as I mentioned in a [previous comment](https://github.com/metrumresearchgroup/bbr/pull/487#discussion_r883629404).

Perhaps check with @seth127 to see what his thoughts are on this.

seth127:
I think we should keep it `add_star()` and `remove_star()` for consistency of user interface.

I also think `remove_star()` doesn't need to use `modify_model_field()` internally, because it's the only function we have that actually blows away a model field instead of modifying it. Changing `modify_model_field()` to make it work for `remove_star()` feels unnecessary. I would just implement `remove_star()` like this:

```r
remove_star <- function(.mod) {

  # update .mod with any changes from yaml on disk
  check_yaml_in_sync(.mod)
  
  .mod[[YAML_STAR]] <- NULL
  
  # overwrite the yaml on disk with modified model
  .mod <- save_model_yaml(.mod)
  
  return(.mod)
}
```

Then the tests can just change to be

```r
test_that("Checking that set_star_status is reflecting boolean data correctly YAML[BBR-MMF-023]", {
  temp_mod_path <- create_temp_model()
  new_mod <- read_model(temp_mod_path)
  expect_null(new_mod$star)

  new_mod <- add_star(new_mod, TRUE)
  expect_true(new_mod$star)

  new_mod <- remove_star(new_mod, FALSE)
  expect_null(new_mod$star)


})
```

kyleam:
@seth127:
> I also think `remove_star()` doesn't need to use `modify_model_field()` internally [...]

Ah, interesting point.  

Doesn't that pretty much apply to `add_star()` too?  It is setting a value, but that logic looks much closer to `remove_star()` than `modify_model_field()`.  So I guess I'm suggesting a light internal helper called by `remove_star()` and `add_star` that essentially does what you posted but takes two values.  A nice advantage of that is that we could return `modify_model_field()` back to its state before this PR, and avoid adding the `.bool_value` parameter.

> Then the tests can just change to be [...]

It looks like you forgot to remove the second argument (value).

kyleam:
@rymarinelli:
> Otherwise, the only expected input for the constructor are characters.

Sorry, I'm not following.  Which constructor?

inst/model/nonmem/test-workflow-bbi-models/1.yaml
seth127: This whole folder should be deleted (i.e. not tracked). It is created fresh [here](https://github.com/metrumresearchgroup/bbr/blob/develop/tests/testthat/test-workflow-bbi.R#L18-L26) every time and should be deleted [here](https://github.com/metrumresearchgroup/bbr/blob/develop/tests/testthat/test-workflow-bbi.R#L33-L37) but somehow it seems to have gotten stuck and checked in by accident.

man/Add.Rd
```diff
@@ -0,0 +1,21 @@
+% Generated by roxygen2: do not edit by hand
```
seth127, line 1: I'm confused why this file is called "Add.Rd" but I have a feeling it's left over from running the roxygen before this function was renamed? I don't know. Either way, we probably wanna `git rm` it and run `devtools::document()` again to refresh it. If it comes up weirdly named again, there may be an issue I'm not spotting in the roxygen code.

seth127:
Responses to some comments above

----

> So I guess I'm suggesting a light internal helper called by remove_star() and add_star() that essentially does what you posted but takes the desired state as an argument. 

I could see this approach being good. And I do think it should be internal. Something like this?

```r
#' Adds or removes a boolean attribute
#' @keywords internal
modify_model_field_lgl <- function(.mod, .field, .action = c("add", "remove")) {
  
  # update .mod with any changes from yaml on disk
  check_yaml_in_sync(.mod)
  
  # check action and modify model
  .action <- match.arg(.action)
  if (.action == "add") {
    .mod[[.field]] <- TRUE
  }
  if (.action == "remove") {
    .mod[[.field]] <- NULL
  }
  
  # overwrite the yaml on disk with modified model
  .mod <- save_model_yaml(.mod)
  
  return(.mod)
}
```

And then the two we export would just be:

```r
add_star <- function(.mod) {
  modify_model_field_lgl(.mod, YAML_STAR, "add")  
}

remove_star <- function(.mod) {
  modify_model_field_lgl(.mod, YAML_STAR, "remove")  
}
```

I kinda love that actually. @rymarinelli and I discussed something like this earlier and I said "don't bother" but I think I've changed my tune. I'm not sure if we'll add other boolean attributes or not, but if we do, this sure makes it easy.

---- 

> A nice advantage of that is that we could return modify_model_field() back to its state before this PR, and avoid adding the .bool_value parameter.

This is also a good point, I think I'd support this. Do you follow @rymarinelli ? Basically `modify_model_field()` can revert to [how it currently looks on `develop`](https://github.com/metrumresearchgroup/bbr/blob/develop/R/modify-model-field.R#L1-L73) because we won't be calling it at all with these new functions.

---

> It looks like you forgot to remove the second argument (value). 
_in the test code I wrote above_

This is true. I changed it in the comment above too, but now it should look like

```r
test_that("Checking that set_star_status is reflecting boolean data correctly YAML [BBR-MMF-023]", {
  temp_mod_path <- create_temp_model()
  new_mod <- read_model(temp_mod_path)
  expect_null(new_mod$star)

  new_mod <- add_star(new_mod)
  expect_true(new_mod$star)

  new_mod <- remove_star(new_mod)
  expect_null(new_mod$star)
})
```


> I think the .boolean is needed for new_model. Otherwise, the only expected input for the constructor are characters.

I think you're talking about [here](https://github.com/metrumresearchgroup/bbr/blob/9b86fa32e0ecb32ba4663d068c661de17e5eaa2c/R/new-model.R#L72) but maybe I'm wrong. Either way, we talked earlier about switching that to something like 

```r
if (isTRUE(.star))     .mod <- add_star(.mod)
```

rymarinelli:
> @rymarinelli:
> 
> > Otherwise, the only expected input for the constructor are characters.
> 
> Sorry, I'm not following. Which constructor?

 I mostly put the boolean parameter to ensure that boolean input would be passed into .star. I think the function expects character data otherwise. 

`new_model <- function(
  .path,
  .description = NULL,
  .based_on = NULL,
  .tags = NULL,
  .bbi_args = NULL,
  .overwrite = FALSE,
  .star = NULL,
  .model_type = c("nonmem")
)`

inst/model/nonmem/test-workflow-bbi-models/1.yaml
```diff
@@ -0,0 +1,10 @@
...
+description: original test-workflow-bbi model
+tags:
+- acop tag
+- other tag
+bbi_args:
+  overwrite: yes
+  threads: 4.0
+starred:
```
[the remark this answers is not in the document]
rymarinelli: I think it is save to get rid of then.

R/copy-model-from.R
```diff
@@ -167,6 +172,7 @@ copy_nonmem_model_from <- function(
     .description = .description,
     .based_on = c(.parent_based_on, .based_on_additional),
     .tags = new_tags,
+    .star = ,
```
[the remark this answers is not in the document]
rymarinelli: Thanks, I thought I corrected this earlier.

seth127:
@kyleam I went through this and pair programmed with @rymarinelli to clean a few things up. This should be good to go, but take a quick look to see if any of the changes you requested are still outstanding. Thanks.

tests/testthat/test-copy-model-from.R
```diff
@@ -61,6 +61,21 @@ test_that("copy_from_model options work [BBR-CMF-002]", {
...
+
+
+  #Editing YAML to add star status is true from clean model
+  temp_mod_path <- create_temp_model()
+  new_mod <- read_model(temp_mod_path)
+  new_mod <- add_star(new_mod)
+
+  #Checking that star is written out to the yaml
+  model_yml <- paste0(temp_mod_path, ".yaml") %>% read_yaml() %>% names()
+  expect_identical(model_yml[5],"star")
```
kyleam, line 73: Okay as is, but testing the index 5 seems a bit more brittle than needed.  Wouldn't it be sufficient to just test that star is in name?

(Another nit-picky comment is that you do the `read_yaml` call again below, so you could just do it one time and store it as a variable.)

kyleam (changes requested):
Nice, thank you both.

Left a few minor comments.  Most important one is about the documented type for `.action`.

R/new-model.R
```diff
@@ -68,7 +70,7 @@ new_model <- function(
   if (!is.null(.tags))        .mod <- replace_all_tags(.mod, .tags)
   if (!is.null(.bbi_args))    .mod <- replace_all_bbi_args(.mod, .bbi_args)
   if (!is.null(.based_on))    .mod <- replace_all_based_on(.mod, .based_on)
-
+  if (isTRUE(.star))          .mod <- add_star(.mod)
```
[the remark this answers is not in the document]
seth127: That's right. This code is in `new_model()` so there won't be a star unless someone passes `.star = TRUE`. For the copied model case, the logic is all in [here](https://github.com/metrumresearchgroup/bbr/blob/019171303cbebe3e1e95a4726806a9265447b7ea/R/copy-model-from.R#L127-L182) and you can see that we never inherit a star from a parent, we just pass through the `.star` arg that the user passes.

R/modify-model-field.R
```diff
@@ -473,6 +499,30 @@ safe_based_on <- function(.start, .based_on) {
   return(tools::file_path_sans_ext(.based_on))
 }
 
+#' Adds or removes boolean attribute from model object and YAML
+#' @param .mod list object of model data
+#' @param .field boolean attribute to be set
+#' @param .action list of boolean variables to be set
```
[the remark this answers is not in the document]
seth127: Right, this looks like a copy/paste error. I think it should be something like

```r
#' @param .action character scalar denoting the action to take. Can be either "add" or "remove".
```

R/new-model.R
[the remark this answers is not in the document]
kyleam: > the logic is all in [here](https://github.com/metrumresearchgroup/bbr/blob/019171303cbebe3e1e95a4726806a9265447b7ea/R/copy-model-from.R#L127-L182)

Yes, that's where my understand is coming from :)  It's just that this interaction/logic seems non-obvious enough to be worthy of a "reading aloud" comment.

Thanks for confirming.

tests/testthat/test-modify-model-field.R
```diff
@@ -354,3 +354,30 @@ test_that("remove_tags() works correctly [BBR-MMF-022]", {
...
+
+test_that("Checking that remove_star is removing star from model object [BBR-MMF-024]", {
+  temp_mod_path <- create_temp_model()
+
+  # make a model object and check YAML
+  new_mod <- read_model(temp_mod_path)
+
+  expect_null(new_mod$star)
+  new_mod <- remove_star(new_mod)
+  expect_null(new_mod$star)
```
kyleam, line 380: I can think of two remove_star cases:
* when `remove_star` is called on a model without a star, the end result is a model without a star
* when `remove_star` is called on a model with a star, the end result is a model without a star

This test is only testing the first case.  I'd suggest adding a second block where you add a star, assert `expect_true(new_mod$star)`, remove it, and then assert `expect_null(new_mod$star)`.

tests/testthat/test-copy-model-from.R
```diff
@@ -61,6 +61,23 @@ test_that("copy_from_model options work [BBR-CMF-002]", {
...
+  expect_null(new_mod$star)
+  new_mod <- add_star(new_mod)
+
+  #Checking that star is written out to the yaml
+  model_yml <- paste0(temp_mod_path, ".yaml") %>% read_yaml()
+  yaml_fields <- model_yml %>% names()
+  expect_true("star" %in% yaml_fields)
+
+  #Checking the yaml is correctly specified with true
+  model_yml <- paste0(temp_mod_path, ".yaml") %>% read_yaml()
```
kyleam, line 78: It's okay as is, but as mentioned in [my previous comment](https://github.com/metrumresearchgroup/bbr/pull/487#discussion_r888339707), I don't think you need this second `model_yml` assignment (it repeats the same `read_yaml()` call above).

kyleam (changes requested):
I have one more request for an additional `remove_star` test case.  Otherwise this looks good to me.

tests/testthat/test-modify-model-field.R
rymarinelli: @kyleam just added the extra block

Issue #488, opened 2022-05-24
test_threads() patches


seth127:
#473 created a new function `test_threads()` (among others). It has been merged to `develop` but before release we are doing some user testing. This issue encapsulates changes we plan to make based on feedback from that user testing. Detailed descriptions in comments.

- [x] returns empty for non-parallel model
- [x] suppress incomplete final line warning
- [x] if any of the `model_summaries()` calls fail, `check_run_times()` returns `NA`
- [x] rename `model_run` column to `run` for consistency ([here](https://github.com/metrumresearchgroup/bbr/blob/develop/R/test-threads.R#L179))
- [x] look into pulling threads from somewhere _other than_ `$bbi_args$threads` (potentially `bbi_config.json`)

seth127:
# returns empty for non-parallel model

If `$bbi_args$threads` is empty (i.e. the model was _not_ run in parallel) [here](https://github.com/metrumresearchgroup/bbr/blob/develop/R/test-threads.R#L169) or [here](https://github.com/metrumresearchgroup/bbr/blob/develop/R/test-threads.R#L173) it will cause the tibble returned on [from the map](https://github.com/metrumresearchgroup/bbr/blob/develop/R/test-threads.R#L178) to have 0 rows, effectively skipping that model.

Instead, if `$bbi_args$threads` is empty, set `threads <- 1` to indicate this model was run on a single thread.

seth127:
# incomplete final line

Add a `suppressSpecificWarning(.regexp = "incomplete final line") to [this `readLines()` call](https://github.com/metrumresearchgroup/bbr/blob/develop/R/test-threads.R#L76)

seth127:
# `check_run_times()` returns `NA`

`check_run_times()` should always return a tibble with the columns: `run, threads, all_of(.return_times)` 

Currently it returns `NA` if any of the `model_summary()` calls fail, and it returns `TRUE` if `wait_for_nonmem()` doesn't finish.

## Proposed Fix

Move the `tryCatch()` on line 156 to instead wrap the [interior of the map function](https://github.com/metrumresearchgroup/bbr/blob/develop/R/test-threads.R#L160-L184). If an error is caught, return an one-row tibble with columns: `run, threads, all_of(.return_times)`, with `run` filled in and everything else `NA` 

[...]

PR #489, opened 2022-05-25
check_run_times refactor

barrettk:
closes https://github.com/metrumresearchgroup/bbr/issues/488

### DIFF
--- a/R/test-threads.R
+++ b/R/test-threads.R
@@ -73,7 +73,7 @@ test_threads <- function(
   search_str <- "MAXEVAL|NITER"
   map(.mods, function(.mod){
     mod_path <- get_model_path(.mod)
-    mod_lines <- mod_path %>% readLines()
+    mod_lines <- mod_path %>% readLines() %>% suppressSpecificWarning("incomplete final line found")
     str_line_loc <- which(grepl(search_str, mod_lines))
 
     if(is_empty(str_line_loc)){
@@ -151,14 +151,23 @@ check_run_times <- function(
     check_model_object(.mods, .mod_types = c("bbi_summary_list"))
   }
 
+  if(.wait){
+    tryCatch({
+      wait_for_nonmem(.mods, ...)
+    }, error = function(cond){
+      return(invisible(TRUE))
+    }, warning = function(cond){
+      message(cond)
+      message("\nConsider setting/increasing the `time_limit` in `wait_for_nonmem()`. See ?check_run_times for details")
+      return(invisible(TRUE))
+    })
+  }
 
-
-  tryCatch({
-    if(.wait) wait_for_nonmem(.mods, ...)
-
-    map_dfr(.mods, ~ {
+  map_dfr(.mods, ~ {
+    tryCatch({
       if(inherits(.x, NM_SUM_CLASS) | (inherits(.x, "list") && !inherits(.x, NM_MOD_CLASS))){
         .sum <- .x
+        model_run <- basename(.sum$absolute_model_path)
         run_details <-
           if(inherits(.x, "bbi_nonmem_summary")){
             .sum$run_details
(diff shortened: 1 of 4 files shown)


seth127:
Looking good, but let's change [here](https://github.com/metrumresearchgroup/bbr/blob/5b7c0119fe37311e6e6d67a8860b756d8ebe2ee0/R/test-threads.R#L179) and everywhere else we pull in `bbi.yaml` to pull in `bbi_config.json` instead. The json is written out _after_ the model finishes and reflects the configuration that was used, so that's the safer bet.

R/test-threads.R
```diff
@@ -73,7 +73,7 @@ test_threads <- function(
   search_str <- "MAXEVAL|NITER"
```
seth127, line 73: It won't let me comment on the actual line (because you didn't change it) but we should add `.overwrite = TRUE` to the `copy_model_from()` call.

You already have `overwrite = TRUE` in the `submit_models()` call (which I think is the right move), but if you try to call `test_threads()` and you forgot to cleanup before, it will fail. Calling `delete_mods()` is the obvious move, but this would likely happen in a new session when you don't have your model objects around anymore (and getting them back is actually kind of a pain, because the names are all annoying and constructed).

R/test-threads.R
```diff
@@ -151,46 +151,81 @@ check_run_times <- function(
...
+    })
+  }
 
-
-  tryCatch({
-    if(.wait) wait_for_nonmem(.mods, ...)
-
-    map_dfr(.mods, ~ {
+  map_dfr(.mods, ~ {
+    tryCatch({
```
seth127, line 167: This works, but at this point I think we've built this section up so much that it's gotten a bit unwieldy. I think it can be refactored to simplify quite a bit, which will make maintaining it significantly easier.

Try replacing all of lines 167-213 with this:

```r
    tryCatch({
      # unpack bbi_summary_list element
      if (!is.null(.x$bbi_summary)) .x <- .x$bbi_summary

      # get model id and convert to summary object if necessary
      model_run <- get_model_id(.x)
      if (inherits(.x, NM_MOD_CLASS)) {
        .x <- model_summary(.x, .bbi_args = list(no_grd_file = TRUE, no_ext_file = TRUE, no_shk_file = TRUE))
      }

      # get number of threads from bbi_config.json
      config <- jsonlite::fromJSON(file.path(.x[[ABS_MOD_PATH]], "bbi_config.json"))
      threads <- ifelse(
        config$configuration$parallel,
        config$configuration$threads,
        1
      )

      data <- tibble::tibble(
        run = model_run,
        threads = threads,
        estimation_time = .x$run_details$estimation_time,
        covariance_time = .x$run_details$covariance_time,
        cpu_time = .x$run_details$cpu_time) %>%
        select(run, threads, all_of(.return_times))
      return(data)
```

tests/testthat/test-test-threads.R
```diff
@@ -113,7 +113,14 @@ withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
 
 
   test_that("check_run_times() returns NA for dry runs [BBR-TSTT-005]", {
-    expect_equal(check_run_times(mod1, .wait = F), NA)
```
seth127, line 116: line 9 of this test file should say

```r
MODEL_DIR_BBI <- file.path(dirname(ABS_MODEL_DIR), "test-test-threads-models")
```

The whole point of this pattern to create a temp dir and be able to tell what created it (if it fails to get cleaned up).

seth127 (changes requested):

seth127:
Overall looks great. It passes all the manual tests we defined in #488 

A few changes requested, one a minor refactor for maintainability, and a couple other little fixes I spotted. Nice work.

--- 

**Evidence of manual tests:**

Screen Shot 2022-05-31 at 9 52 12 PM

Screen Shot 2022-05-31 at 10 26 25 PM

Screen Shot 2022-05-31 at 10 32 53 PM

Screen Shot 2022-05-31 at 10 34 37 PM

R/test-threads.R
barrettk: Tested that it worked with this change pretty diligently  and was pleasantly surprised. Cant believe you reduced it by that much, but yeah I agree it looks much nicer and will be much easier to evolve with time.

barrettk:
Made the necessary fixes. Never thought to check it periodically while using qstat to inform expectation, but thats a nice demonstration for expected functionality.

barrettk:
@seth127 I fixed the workflow tests that errored on my end. It was passing when I clicked "run tests" via Rstudio, but failed via `devtools::test()`. Seems I forgot to change `model_run` to `run` in that script. Not sure why the other method didnt pick that up.

seth127 (approved):
Changes look good

PR #493, opened 2022-06-03
Validation: creating bbr-requirements.yaml and modifying bbr-stories.yaml

seth127:
This is an update to our validation process. Where we previously linked user stories directly to tests that confirm the associated functionality, going forward we will link user stories to requirements that need to be met to satisfy the story, and then link those requirements to tests that confirm them.

### DIFF
--- a/inst/validation/bbr-requirements.yaml
+++ b/inst/validation/bbr-requirements.yaml
@@ -1,0 +1,1060 @@
+BBR-R001:
+  description: check_status_code works as expected
+  tests:
+  - BBR-BBR-001
+BBR-R002:
+  description: bbi_dry_run() correctly returns object
+  tests:
+  - BBR-BBR-002
+BBR-R003:
+  description: check_bbi_exe() correctly errors or finds paths
+  tests:
+  - BBR-BBR-003
+BBR-R004:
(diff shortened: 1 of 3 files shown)


seth127:
@kyleam @barrettk @rymarinelli tagging you all for visibility here. I put @barrettk on to review this, but any of you could. Probably pull this and run `build-validation-docs.R` to make sure it works for you and builds docs that look good.

To be clear, this is only to generate the new requirements YAML file. We will be manually adding to that file going forward, _not_ pulling the descriptions from test names as we do here.

barrettk:
@seth127 the validation docs didnt build for me. I didnt see a new version of mrgvalprep to be installed via pkgr. Is there anything else I need to do before generating the docs?
image
image

kyleam:
@barrettk You'll need to install mrgvalprep from its latest develop branch to get the fix from https://github.com/metrumresearchgroup/mrgvalprep/pull/43

barrettk (approved):
Looks good. I like the idea of making all the requirement ID's based on the previous testID, as that makes it easier to assume there are no duplicates. Validation docs built after pulling in the latest mrgvalprep changes

Issue #495, opened 2022-06-13
test_threads fails with no error if there is no bbi.yaml


barrettk:
This is possibly a somewhat larger issue. If `.wait = FALSE`, bbr will not error out saying there is no `bbi.yaml` file. In my opinion we would want this to error out in both circumstances so users aren't waiting for something to happen (see below).

![image](https://user-images.githubusercontent.com/18685116/173414747-4af67ed6-9a24-4652-b99b-262317972c03.png)

![image](https://user-images.githubusercontent.com/18685116/173414802-14734e0c-b21a-4fee-a34d-a682d30fc036.png)

For the purposes of fixing this issue in test_threads, I can simply check for the yaml file before submitting

seth127:
Good catch @barrettk . I see you put in a targeted fix in #496 but I think you're right that this might merit a more general fix.

I didn't look too hard, but I think `submit_model()` checks for this [here](https://github.com/metrumresearchgroup/bbr/blob/develop/R/submit-model.R#L98-L109) before calling out to `bbi`. However, `submit_models()` (_plural_), which is what's called by `test_threads()` looks like it _doesn't_ check for it, unless the user passes a custom config path: look at the code [here](https://github.com/metrumresearchgroup/bbr/blob/develop/R/submit-models.R#L125-L131) and notice that the `else` block is missing.

I would interested in seeing if adding in that `else` block to `submit_models()` fixes the issue (without the fix in #496). Either way, we probably want to add that handling to `submit_models()` I would think.

barrettk:
@seth Interestingly, it doesnt seem to fix the issue. When running `submit_models(.mods, .wait = FALSE)`, we get the appropriate error now:
```r
> submit_models(.mods, .wait = FALSE)
 Error: Looking for default configuration file at /data/Projects/package_dev/bbr/inst/model/nonmem/test-test-threads-models/bbi.yaml: no such file or directory 
```

However, test_threads doesnt seem to pass that error forward:
![image](https://user-images.githubusercontent.com/18685116/174112227-a3af3d81-c573-44a4-a1f4-da102e86d1b5.png)

Im not sure if we necessarily need to keep the test_threads patch I made, but will look into whether it can be removed, as it shouldnt be necessary

PR #496, opened 2022-06-13
submit_models() does not error when wait = FALSE

barrettk:
closes https://github.com/metrumresearchgroup/bbr/issues/495, although not sure if that issue warrants a larger discussion


barrettk:
Do not believe this patch requires any additional requirements

R/submit-models.R
```diff
@@ -128,6 +128,11 @@ submit_nonmem_models <- function(.mods,
         cmd_args,
         sprintf("--config=%s", normalizePath(.config_path))
       )
+    } else {
+      default_config_path <- file.path(model_dir, "bbi.yaml")
+      if (!fs::file_exists(default_config_path) && !isTRUE(.dry_run)) {
+        stop(glue("Looking for default configuration file at {default_config_path}: no such file or directory"), call. = F)
```
kyleam, line 134: Hmm, why does the error message say "or directory"?  Aren't we checking for and expecting a regular file?

R/submit-models.R
```diff
@@ -128,6 +128,11 @@ submit_nonmem_models <- function(.mods,
         cmd_args,
         sprintf("--config=%s", normalizePath(.config_path))
       )
+    } else {
+      default_config_path <- file.path(model_dir, "bbi.yaml")
+      if (!fs::file_exists(default_config_path) && !isTRUE(.dry_run)) {
```
kyleam, line 133: I'd say that the closer a dry-run mode behaves to the regular mode the better, so it'd be good if it would include this "file exists" guard.  Three tests fail with that change, but my initial impression is that that is more of an issue with those tests than with enabling the guard for dry runs.

Perhaps I'm missing something.  Can you explain more why this check should be bypassed when `.dry_run = TRUE`, @barrettk (or @seth127)?  I see `.dry_run` mentioned a bit at gh-495, but nothing that seems to lay that out.

Anyway, here's what that direction could look like, including the necessary test tweaks:

<details>
<summary>diff</summary>

```diff
diff --git a/R/submit-models.R b/R/submit-models.R
index 19c9ca8..8b79729 100644
--- a/R/submit-models.R
+++ b/R/submit-models.R
@@ -84,7 +84,7 @@ submit_models.list <- function(
 #' @importFrom stringr str_detect
 #' @importFrom tools file_path_sans_ext
 #' @importFrom purrr map
-#' @importFrom rlang is_bare_list
+#' @importFrom rlang %||% is_bare_list
 #' @return A list of S3 objects of class `bbi_process`
 #' @keywords internal
 submit_nonmem_models <- function(.mods,
@@ -122,17 +122,14 @@ submit_nonmem_models <- function(.mods,
 
     model_dir <- get_model_working_directory(.run[["models"]][[1L]])
 
+    checkmate::assert_file_exists(
+      .config_path %||% file.path(model_dir, "bbi.yaml"))
+
     if (!is.null(.config_path)) {
-      checkmate::assert_file_exists(.config_path)
       cmd_args <- c(
         cmd_args,
         sprintf("--config=%s", normalizePath(.config_path))
       )
-    } else {
-      default_config_path <- file.path(model_dir, "bbi.yaml")
-      if (!fs::file_exists(default_config_path) && !isTRUE(.dry_run)) {
-        stop(glue("Looking for default configuration file at {default_config_path}: no such file or directory"), call. = F)
-      }
     }
 
     return(list(cmd_args = cmd_args, model_dir = model_dir))
diff --git a/tests/testthat/test-print.R b/tests/testthat/test-print.R
index 8763295..275fb18 100644
--- a/tests/testthat/test-print.R
+++ b/tests/testthat/test-print.R
@@ -37,6 +37,11 @@
     temp_dir <- file.path(get_model_working_directory(MOD1), "print-test")
     fs::dir_create(temp_dir)
     on.exit(fs::dir_delete(temp_dir))
+
+    withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
+      bbi_init(temp_dir, ".", .no_default_version = TRUE)
+    })
+
     mods <- purrr::map(1:50, ~copy_model_from(MOD1, file.path("print-test", .x)))
     proc <- submit_models(mods, .dry_run = TRUE)
 
diff --git a/tests/testthat/test-submit-models.R b/tests/testthat/test-submit-models.R
index b1c135e..02e444b 100644
--- a/tests/testthat/test-submit-models.R
+++ b/tests/testthat/test-submit-models.R
@@ -72,10 +72,14 @@
             })
 
   test_that("submit_models() works for models in different directories [BBR-SBMT-011]", {
-    new_dir <- "level2"
-    fs::dir_create(file.path(MODEL_DIR, new_dir))
+    new_dir <- file.path(ABS_MODEL_DIR, "level2")
+    fs::dir_create(new_dir)
     on.exit(cleanup())
 
+    withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
+      bbi_init(new_dir, ".", .no_default_version = TRUE)
+    })
+
     # TODO: use test helper functions, e.g., create_all_models(), once the
     # model_directory option is deprecated
     mod2 <- copy_model_from(
@@ -160,6 +164,10 @@
     mod <- read_model(temp_mod_path)
     mod <- replace_all_bbi_args(mod, NULL)
 
+    withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
+      bbi_init(dirname(temp_mod_path), ".", .no_default_version = TRUE)
+    })
+
     res <- submit_models(list(mod), .dry_run = TRUE)
 
     expect_identical(
```

</details>

kyleam: @seth127
>  I guess its debatable whether being able to run .dry_run=T without a valid config is actually "desired behavior" (from a UX standpoint) or not

Yes, I think so.  I lean toward a dry run erroring for this (similar to how a dry-run `git push` would error if, say, you didn't have permissions to push to the remote).  But I'm okay with keeping the current `submit_model()` behavior as long as the case is being made that it is better from UX perspective or that it's needed for backward compatibility, as opposed to it being something done for testing convenience.

A related point: It seems inconsistent that a dry run would fail when a config is explicitly given but not for the default.  Perhaps that is done for the sake of the `normalizePath` call, but that doesn't change that it's inconsistent from the UI perspective.

> I will say that I think we should keep submit_model() and submit_models() consistent

Right, sorry, I didn't step back from the changes of this PR when creating that patch, but I very much agree with this.

@barrettk 
> I'll test out Kyle's solution and see If I can tweak it all

Thanks.

barrettk:
@kyleam I made the error handling a bit more informative via some of the `assert_that()` functions. I did something similar in another PR as well. Let me know what you think:

```r
> mods <- test_threads(.mod = MOD1, .threads = c(2,4), .mode = "local", .dry_run = FALSE)
 Error in .f(.x[[i]], ...) : 
  No bbi.yaml file was found in the execution directory. Please run `bbi_init()` in the appropriate directory to continue. 
```

R/submit-model.R
```diff
@@ -95,17 +96,20 @@ submit_nonmem_model <- function(.mod,
...
   model_dir <- get_model_working_directory(.mod)
 
+  coll = makeAssertCollection()
+  assert_file_exists(
+    .config_path %||% file.path(model_dir, "bbi.yaml"),
+    add = coll)
+  if(!coll$isEmpty()){
+    stop(paste("No bbi.yaml file was found in the execution directory.",
+               "Please run `bbi_init()` in the appropriate directory to continue."))
+  }
```
kyleam, line 106: The custom error message is a nice improvement.

What's the advantage of this whole `makeAssertCollection` dance over a plain `file_exists` check?   Something like:

```R
if(!fs::file_exists(.config_path %||% file.path(model_dir, "bbi.yaml"))) {
  stop(paste("No bbi.yaml file was found in the execution directory.",
             "Please run `bbi_init()` in the appropriate directory to continue."))
}
```

Two other minor thoughts (feel free to take or leave):

 * If it's a custom config, it doesn't need to be named "bbi.yaml" so perhaps "No bbi.yaml" -> "No bbi configuration"?
 * "run `bbi_init()` _in_ the appropriate directory" -- very minor, but perhaps _with_ is better, since the caller can pass the directory to that function.

tests/testthat/test-submit-model.R
```diff
@@ -4,16 +4,19 @@ context("submit_model(.dry_run=T)")
...
-readr::write_file("created_by: test-submit-model", "bbi.yaml")
-on.exit({ fs::file_delete("bbi.yaml")})
+
 
 model_dir <- ABS_MODEL_DIR
 mod_ctl_path <- file.path(model_dir, CTL_FILENAME)
 
+# create fake bbi.yaml
+readr::write_file("created_by: test-submit-model", file.path(model_dir, "bbi.yaml"))
+on.exit({ fs::file_delete(file.path(model_dir, "bbi.yaml"))})
```
kyleam, line 14: I like this fake bbi.yaml approach better than the `bbi_init` calls in test-print.R and test-submit-models.R that came from my patch at https://github.com/metrumresearchgroup/bbr/pull/496#discussion_r900086187

For consistency, I think it'd make sense to convert those to using this fake bbi.yaml approach, though no strong objections if you want to leave it as is.

R/submit-model.R
barrettk: I suppose there would be no difference in that case. The purpose was to take advantage of the `assert_file_exists` function, but not have to use its default error message (which isn't that informative). There are a number of cases where we use `assert_that` functions, with no custom error message. Though we could replace them all with if-statements, I like the simplicity of using assert_that functions, but being able to customize the error message. Was hoping this and the other PR where I made that change could be a pattern we start to implement.

tests/testthat/test-submit-model.R
barrettk: Yeah I agree. I noticed that that file didnt have any fake bbi.yaml's in there before, so wasn't sure if models were running in various directories (didnt spend as long looking through it). Will make that change assuming that isnt the case though

R/submit-model.R
kyleam: > The purpose was to take advantage of the assert_file_exists function

But what value is the `assert_file_exists` function adding?  In my opinion, it is more complex and less readable, so that pattern should be buying us something if we decide to use it.

kyleam: > Well its minor, but in the case of that function, it also checks accessing rights of those directories.

Hmm, you're talking about the `access` argument?  You're calling that with the default value, so it's not checking the access:

```
> checkmate::assert_file_exists("/etc/sudoers")
> checkmate::assert_file_exists("/etc/sudoers", access = "r")
Error: Assertion on '"/etc/sudoers"' failed: '/etc/sudoers' not readable.
```

> I might be a little biased, but the more i've used checkmate functions, i've really taken a liking to them, partly because I find them less prone to coding errors, but mainly because I like the pattern of them (i.e. always add an assert_blah call for every function argument).

I'm not arguing against checkmate functions for simple argument checking.  I'm objecting to this pattern unless it brings some concrete value that we decide it is worth the added complexity:


```
  coll = makeAssertCollection()
  assert_*(..., add = coll)
  if(!coll$isEmpty()){
    stop(...)
  }
```

To understand that, a reader has to stop and try to figure out `makeAssertCollection` is, what the `add` argument is about, what `$isEmpty()` is referring to.  They'd need to spend much less time parsing this:

```
if(!fs::file_exists(...))) {
  stop(...)
}
```

And if we do decide there is a `checkmate` function that we want to use, why not use the `check_*` variants:

```
> isTRUE(checkmate::check_file_exists("/etc/sudoers"))
[1] TRUE
> isTRUE(checkmate::check_file_exists("/etc/sudoers", access = "r"))
[1] FALSE
```

barrettk: Oop you're right about the access thing. I personally dont think inspecting functions like that is that big of a deal, and comes with the territory, but I realize it looks a little more complicated than it needs to. But again, I know I am biased and like that method, mainly because you can do other things like `coll$push("Custom error message")`, `coll$getMessages()`, etc., and thought that might be a nice pattern to reinforce in the future. 

Hadn't looked into the difference in `test_file_exists` before, so perhaps thats the best path forward for the time being

barrettk: Changed it to:
```r
  .path_exists <- test_file_exists(
    .config_path %||% file.path(model_dir, "bbi.yaml"),
    access = "rw")
  if(!.path_exists){
    stop(paste("No bbi configuration was found in the execution directory.",
               "Please run `bbi_init()` with the appropriate directory to continue."))
  }
```

barrettk:
@kyleam btw I added a small fix for test_threads, that will delete the models if the submit_model call fails (which was related to the original intention of this PR). I think it's relevant enough to include in this PR, and didnt think it was worth making a separate one for

R/test-threads.R
```diff
@@ -97,7 +97,15 @@ test_threads <- function(
...
 
-  submit_models(.mods, .wait = FALSE, ...)
+
+  tryCatch({
+    submit_models(.mods, .wait = FALSE, ...)
+  },
+  error = function(cond){
+    delete_models(.mods = .mods, .force = TRUE) %>% suppressMessages()
+    message(cond)
+    return(invisible(TRUE))
```
kyleam, line 107: Why return `TRUE`?  I expected this to clean up the models and then re-signal the error.  If there's some reason it shouldn't do that, then I'd say `NULL` would be the more typical return value.

tests/testthat/test-print.R
```diff
@@ -37,6 +37,11 @@ withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
     temp_dir <- file.path(get_model_working_directory(MOD1), "print-test")
     fs::dir_create(temp_dir)
     on.exit(fs::dir_delete(temp_dir))
+
+    withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
+      bbi_init(temp_dir, ".", .no_default_version = TRUE)
+    })
```
kyleam, line 43: You switched the other spots over to using the dummy file.  Is there are reason this one doesn't use that approach?

R/submit-model.R
kyleam: I think checking for access is an overkill and, if the message isn't going to say anything about, it's inviting confusion.  (And is write access really needed?)  I'd just leave it at `file_exists()`, something that is done a good amount in this code base and is widespread throughout code bases in general (nothing specific to R here).

But I'm not going to block on this.  The change to using `test_file_exists` is good enough for me.  Thanks.

R/test-threads.R
barrettk: It does re-signal the error via `message(cond)`, but the important change was deleting the newly created model files _if_ submission fails. Im fine with changing it to `NULL` though.

kyleam: > It does re-signal the error via `message(cond)`

That doesn't re-signal it; it gives a message about the error.  Re-signaling it would mean that execution was halted unless outer code caught it.

This...

```R
f1 <- function() {
  stop("boom")
}

f2 <- function() {
  tryCatch(f1(),
           error = function(e) {
             message(e, " is the message")
           })
}

f2()
message("on I go")
```

versus this:

```diff
diff --git a/scratch.R b/scratch-rethrow.R
index 2952a0c..e0a1533 100644
--- a/scratch.R
+++ b/scratch-rethrow.R
@@ -6,7 +6,7 @@
 f2 <- function() {
   tryCatch(f1(),
            error = function(e) {
-             message(e, " is the message")
+             stop(e)
            })
 }
```

Perhaps demoting the error to a message is the right thing to do here, but it's different (and, as I said, not what I expected).  Is there a reason you prefer to give the message rather than propagate the error?

barrettk: Ahh I take your point. And no to be honest most of the trycatch examples I had looked at in the past used message, so thats probably where I got it from. Can make the change

R/submit-models.R
```diff
@@ -122,8 +123,15 @@ submit_nonmem_models <- function(.mods,
 
     model_dir <- get_model_working_directory(.run[["models"]][[1L]])
 
+    .path_exists <- test_file_exists(
```
kyleam, line 126: One last comment (follow-up on https://github.com/metrumresearchgroup/bbr/pull/496/files#r900510130): I see you adjusted the other site to use `file_exists` (thanks for that), but it looks like you forgot to update this site.

Aside from that inconsistency, this looks ready to go to me.

barrettk: Ah yes, got distracted by the drone errors and forgot to make that change. Newest commit reflects that

R/submit-models.R
```diff
@@ -84,7 +84,8 @@ submit_models.list <- function(
 #' @importFrom stringr str_detect
 #' @importFrom tools file_path_sans_ext
 #' @importFrom purrr map
-#' @importFrom rlang is_bare_list
+#' @importFrom rlang %||% is_bare_list
+#' @importFrom checkmate test_file_exists
```
kyleam, line 88: Please drop this (now unused) and regenerate NAMESPACE.

kyleam (approved):

Issue #404, opened 2021-09-24
format_tags_diff()


seth127:
Makes a nicely formatted column for HTML `log_df` tables that shows the tags added and removed

# Discussion Points

* this is fairly specific to HTML formatting. We should test with pdf and other rendering stuff and see if we need to change anything, or maybe make an argument like `format = c("html", "pdf")` or something.
* We should also test how this interacts with `knitr::kable()`

# Proposed Code

```r
format_tags_diff <- function(.log_df, .keep = FALSE) {
  if (
    !any(str_detect(names(.log_df), "tags_added")) ||
    !any(str_detect(names(.log_df), "tags_removed"))
  ) {
    stop("Must have both `tags_added` and `tags_removed` columns to call `format_tags_diff()`. Use `add_tags_diff()` to create them")
  }
  .log_df <- bbr::suppressSpecificWarning(
    collapse_to_string(.log_df, tags_added, tags_removed),
    .regexpr = "collapse_to_string\\(\\) only works on list columns"
  ) %>%
    mutate(
      tags_diff = paste0(tags_added, "; ~~", tags_removed, "~~") %>% 
        str_replace("; ~~~~","") %>% 
        str_replace("^; ", "")
    )
  if (isFALSE(.keep)) {
    .log_df <- select(.log_df, -tags_added, -tags_removed)
  }
  return(.log_df)
}
```

barrettk:
@seth im not sure this function is right unless im missing something (I know its WIP). Im finding that `tags_added` and `tags_removed` are not necessarily names of `tags_diff(log_df)`, but rather one level down:
```r
> tags_diff(log_df)
$`1`
$`1`$tags_added
[1] "acop tag"  "other tag"

$`1`$tags_removed
[1] ""


$`2`
$`2`$tags_added
[1] "new tag 1" "new tag 2"

$`2`$tags_removed
[1] "acop tag"  "other tag"


$`3`
$`3`$tags_added
[1] "the newest"

$`3`$tags_removed
[1] "other tag" "new tag 2"
```

Im going to write the stories/requirements assuming my assumption is correct. Wanted to give it a run before writing them to have a better understanding of functionality, but I can definitely put something together for whats here right now

barrettk:
Stories:
```
MGMT-S013 (this already exists):
  name: format tags difference
  description: As a user, I want to be able to see the difference in the tags element
    of different models. Be able to either compare two models directly, or compare
[...]

PR #498, opened 2022-06-16
Optionally return new tags_diff column formatted for html rendering

barrettk:
closes https://github.com/metrumresearchgroup/bbr/issues/404


seth127:
@kyleam @barrettk you can pause on this for a moment. I'll take a look next week. This is gonna be a pretty specialized function and I need to think through the use cases a little more. For example: 

* Do we only intend it to be used in Rmd? 
* Does it _need_ to be used with `knitr`?
* Is there a difference based on whether we have `results = 'asis'` in the chunk options

Issue #494, opened 2022-06-10
Requirement is missing information


barrettk:
I believe the below requirement is missing word(s) at the end of the description

see `inst/validation/bbr-requirements.yaml` line 625:
```yml
PEST-R003:
  description: param_estimates correctly errors on
  tests:
  - BBR-PEST-003
```

barrettk:
There seem to be two duplicate test ids:
![image](https://user-images.githubusercontent.com/18685116/173145042-0c55eb0b-5318-4564-b7ed-bd7971c9f580.png)

kyleam:
I believe the repeated test ID is intentional.  If you repeat the same test ID, mrgvalprep can collapse the separate tests into one, setting the new description to the common prefix ("param_estimates correctly errors on" here).

https://github.com/metrumresearchgroup/mrgvalprep/blob/5af255d299eef868441b32a244730cb62cc08ccf/R/parse-test-output.R#L1-L10

But that takes a bit of care to get a result that looks good, and I chose a bad stem here.  

To make it a bit easier to come up with a better stem while keeping the full description readable, mrgvalprep strips a trailing colon:

https://github.com/metrumresearchgroup/mrgvalprep/commit/b7bb5106a929d8c52ab47f36ff8d6791b5cffe1e

So something like this 

```
param_estimates correctly errors: Bayesian model with multiple estimation methods
param_estimates correctly errors: ONLYSIM models
```

would give a better description: "param_estimates correctly errors"

PR #500, opened 2022-06-16
fix requirement ids for parameter_estimates

barrettk:
closes https://github.com/metrumresearchgroup/bbr/issues/494

### DIFF
--- a/inst/validation/bbr-requirements.yaml
+++ b/inst/validation/bbr-requirements.yaml
@@ -619,13 +619,12 @@ PEST-R001:
   tests:
   - BBR-PEST-001
 PEST-R002:
-  description: param_estimates correctly errors on Bayesian model
+  description: param_estimates correctly errors on:
+     - Bayesian model
+     - Bayesian model with multiple estimation methods
+     - ONLYSIM models
   tests:
   - BBR-PEST-002
-PEST-R003:
-  description: param_estimates correctly errors on
-  tests:
-  - BBR-PEST-003
 PEST-R004:
   description: param_estimates correctly warns on mixture model
   tests:
--- a/inst/validation/bbr-requirements.yaml
+++ b/inst/validation/bbr-requirements.yaml
@@ -619,7 +619,7 @@ PEST-R001:
   tests:
   - BBR-PEST-001
 PEST-R002:
-  description: param_estimates correctly errors on: Bayesian model, Bayesian model with multiple estimation methods, ONLYSIM models
+  description: param_estimates correctly errors on unsupported models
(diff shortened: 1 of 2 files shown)


seth127 (dismissed):
We can't use bullets within a requirement:

```
  description: param_estimates correctly errors on:
     - Bayesian model
     - Bayesian model with multiple estimation methods
     - ONLYSIM models
```

inst/validation/bbr-requirements.yaml
```diff
@@ -619,13 +619,9 @@ PEST-R001:
   tests:
   - BBR-PEST-001
 PEST-R002:
-  description: param_estimates correctly errors on Bayesian model
+  description: param_estimates correctly errors on: Bayesian model, Bayesian model with multiple estimation methods, ONLYSIM models
```
kyleam, line 622: I guess the common factor here is just there they're some sort of unsupported model, so what do you think about "param_estimates errors on unsupported models"?

tests/testthat/test-param-estimates.R
```diff
@@ -38,7 +38,7 @@ withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
     )
   })
 
-  test_that("param_estimates correctly errors on ONLYSIM models [BBR-PEST-003]", {
+  test_that("param_estimates correctly errors on ONLYSIM models [BBR-PEST-002]", {
```
kyleam, line 41: I'm to blame for this, but while you're fixing up these IDs, do you mind also dealing with the weird "errors on" ending that you mentioned in gh-494?  Using the colon approach mentioned in gh-494, that'd be something like

```
param_estimates correctly errors: Bayesian model
param_estimates correctly errors: Bayesian model with multiple estimation methods
param_estimates correctly errors: ONLYSIM models
```

inst/validation/bbr-requirements.yaml
barrettk: Yeah I thought about that. Part of me likes the idea of referencing the specific tests they correspond to, mainly in the event we added another unsupported model that deserved its own requirement for some reason. Dont have a strong opinion though and am ok with that change

kyleam: Hmm, yeah, either way is fine with me.  I find the current form a bit wordy, and it seems written in a way that's more likely to need to be updated for new tests, but that's not a big concern.  Whatever you want to go with woks for me.

kyleam (approved):
LGTM.  Thanks for cleaning up my poor wording for the test descriptions.

Issue #491, opened 2022-05-27
Update parallel tips and tricks vignette


seth127:
Now that `test_threads()` will be an exported function, we need to update the [Running NONMEM in Parallel: bbr Tips and Tricks](https://metrumresearchgroup.github.io/bbr/articles/nonmem-parallel.html) vignette to include it.

barrettk:
Note: this issue doesnt require any additional stories/requirements

PR #501, opened 2022-06-17
Patch/bbi tips tricks

barrettk:
closes https://github.com/metrumresearchgroup/bbr/issues/491

> Now that test_threads() will be an exported function, we need to update the [Running NONMEM in Parallel: bbr Tips and Tricks](https://metrumresearchgroup.github.io/bbr/articles/nonmem-parallel.html) vignette to include it.


barrettk:
@seth127 
Assuming there is more going on here than im aware of, as im not sure how to actually run this RMD. Specifically, there is no `pk-parallel` folder in:
```r
MODEL_DIR <- system.file("model", "nonmem" , "pk-parallel", package = "bbr")
```
which prevents all blocks from executing, starting with the section below:
image

I added some `test_threads`/`check_run_times`/`delete_models` code that utilizes the referenced models, but just wanted to confirm there was another procedure that generates this folder & model 200 that im not aware of

seth127:
Ah good point. I don't think we actually include those models in the repo. Most of this vignette doesn't actually run the code (because we obviously don't want to try to submit models to the grid every time we build the vignettes).

Think about if there's an easy way to "fake it" like we're already doing. If not, lemme know and we can talk next week about how to best handle this. To be clear though, I don't think we actually want to submit any models in the code (for reason mentioned above).

PR #504, opened 2022-06-22
README: update for change in default branch

kyleam:
This repo's default branch is now `main` rather than `develop`.

The first commit regenerates `README.md` so that the output matches what the latest knitr would produce (and so that those changes don't get mingled with the branch name change).

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,16 +58,16 @@ functionality is rolled out. A complete list can be found
 
 ### Featured Vignettes
 
-  - [Getting Started with
+-   [Getting Started with
     bbr](https://metrumresearchgroup.github.io/bbr/articles/getting-started.html)
     – Some basic scenarios for modeling with NONMEM using `bbr`,
     introducing you to its standard workflow and functionality.
-  - [Using the based\_on
+-   [Using the based_on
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -58,16 +58,16 @@ functionality is rolled out. A complete list can be found
 
 ### Featured Vignettes
 
-  - [Getting Started with
+-   [Getting Started with
     bbr](https://metrumresearchgroup.github.io/bbr/articles/getting-started.html)
     – Some basic scenarios for modeling with NONMEM using `bbr`,
     introducing you to its standard workflow and functionality.
-  - [Using the based\_on
+-   [Using the based_on
```
kyleam, line 65: > I've confirmed that the link still works in the rendered html.

And now I've just confirmed it works in GitHub's rendering of the README.

README.md
```diff
@@ -83,8 +83,8 @@ provide isolation. To replicate this environment,
 2.  install pkgr
 
 3.  open package in an R session and run `renv::init()`
```
seth127, line 85: this is unrelated to this branch change, but I have been meaning to change this for awhile... this line should be `renv::init(bare = TRUE)`

PR #506, opened 2022-06-27
model summary: detect NaN objective function values from bbi

kyleam:
If the lst files contain values like

    OBJECTIVE FUNCTION VALUE WITHOUT CONSTANT:                        NaN
    OBJECTIVE FUNCTION VALUE WITH CONSTANT:                           NaN

then summarizing the model fails with

    > model_summary(mod)
    Error: parse error: premature EOF

                         (right here) ------^

The main fix for this is on the bbi side [*].  With that change, NaN
values now come through as DefaultFloat64 (aka BBI_NULL_NUM).  Teach
create_summary_object() to map BBI_NULL_NUM to NA.

Making bbr flag these values is left for later (gh-505).

[*] https://github.com/metrumresearchgroup/bbi/pull/268

---

Tests pass for me locally with a bbi built from 5561ec0 (nmparser: use strToFloat to handle NaN objective function values, 2022-06-23):

```
── Skipped tests  ──────────────────────────────────────────────────────────────
• getOption("bbr.DEV_no_min_version") is TRUE (1)

[ FAIL 0 | WARN 0 | SKIP 1 | PASS 984 ]
```

There are other incoming bbi changes, so I think it'd make sense to hold off tagging a bbi dev release and testing on bbr's Drone until more of those changes have landed.

PR #507, opened 2022-06-27
Fix some confusing requirement and test descriptions 

kyleam:
This series is a follow-up to gh-493.  That PR used test descriptions to fill in requirement descriptions when doing the one-time generation of `inst/validation/bbr-requirements.yaml`.  gh-500 fixed one case where this led to a confusing/incomplete requirement description.  When I was working on gh-506, I spotted some more issues and decided to do a wider sweep.  While this probably doesn't catch all the problems, it will hopefully mean that we don't discover a new issue each time we touch `bbr-requirements.yaml`.

Most of these commits follow the same pattern: do a commit (or more) that cleans up a related set of test descriptions followed by a commit that cleans up the `bbr-requirement.yaml` entries.  Some notable things that fall outside this pattern:

```
requirements: collapse param_estimates_compare() requirement
test-param-estimates-batch.R: split up a test

  These two commits rewire the requirement and test IDs of
  `param_estimates_compare` a bit.

test-param-labels.R: actually test all block() cases

  This one fixes a test logic bug that led to a confusing requirement
  description.

requirements: append "works" to bare function name descriptions

  This one adjusts the requirements to avoid descriptions that consist
  of only the function name, but it leaves the corresponding test IDs as is.
```

### DIFF
--- a/tests/testthat/test-param-estimates-batch.R
+++ b/tests/testthat/test-param-estimates-batch.R
@@ -124,7 +124,7 @@ withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
   # compare tests
   ###################
 
-  test_that("param_estimates_compare() works on same model [BBR-PEST-009]", {
+  test_that("param_estimates_compare() works: same model [BBR-PEST-009]", {
 
     on.exit(cleanup())
 
@@ -166,7 +166,7 @@ withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
   })
 
 
-  test_that("param_estimates_compare() works on lots of params [BBR-PEST-009]", {
+  test_that("param_estimates_compare() works: lots of params [BBR-PEST-009]", {
 
     on.exit(cleanup())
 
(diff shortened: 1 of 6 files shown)


tests/testthat/test-param-estimates-batch.R
```diff
@@ -214,15 +214,15 @@ withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
...
+  test_that("param_estimates_compare() works: probs argument [BBR-PEST-011]", {
     res <- param_estimates_compare(
       param_estimates_batch(MODEL_DIR),
       probs = c(.3, .4, .6)
     )
     expect_true(all(c("30%", "40%", "60%") %in% names(res)))
   })
 
-  test_that("param_estimates_compare() works na.rm argument [BBR-PEST-011]", {
+  test_that("param_estimates_compare() works: na.rm argument [BBR-PEST-013]", {
```
kyleam, line 225: There's already a BBR-PEST-013 in another file.  Will bump to 014.

PR #510, opened 2022-06-30
Add extra logic to filter .ctl and .mod file extensions

### LINKED ISSUE #486: ignore file extension in new_model
## Problem

If I pass a model name that includes a valid file extension, e.g. `new_model(file.path(MODEL_DIR, "001.ctl"))`, I get an error. However, I would like `bbr` to just ignore the extension in this case.

rymarinelli:
Closes #486


rymarinelli:
I tried to add some additional handling in new_model for arbitrary file extensions. But `find_nonmem_model_file_path` already throws an error. Including similar logic made tests concerning fail paths fail when using periods. For instance,  bar.foo.ctl would wrongly fail with this logic. I decided to exclude it, since there was already handling 
```
#If there is a file extension that is not .ctl or .mod
  if((stringr::str_detect(.path_extension, "\\.\\w+") == TRUE) && (stringr::str_detect(.path, "\\.ctl|\\.mod") == FALSE))
  {
    #stop('File extension is not supported')
  }
```

rymarinelli:
```
Stories:
  MGMT-S017:
  name: Passing model file extensions
  description: As a user, I want to be able to pass '.ctl' and '.mod' file extensions when making a new model.
  ProductRisk: Low
  requirements:
  - CLM-R005
  - CLM-R006
 ```
```
Requirements:
  CLM-R005:
   description: accept '.ctl' file extension when calling new_model
   tests:
  - BBR-CLM-005
 CLM-R006:
   description: accept '.mod' file extension when calling new_model
   tests:
   - BBR-CLM-006
 ```

kyleam (changes requested):
I'll focus on higher-level comments for now rather than making comments on specific parts of the code:

 * is there any advantage of giving special treatment to `.ctl` or `.mod` rather than just removing the extension, if any?  For example, you could check whether the path is a file with `fs::is_file()` and then if so remove any extension with `fs::path_ext_remove()` or `tools::file_path_sans_ext()`.  (The `fs::is_file()` would avoid doing the wrong thing if a directory path happens to have a period somewhere.)

   At that point, I think it'd be sufficient to just have one test calling `new_model()` using a path with an extension.

 * rather than define your own extension regexs, I suggest use the extension helpers in `tools::` or `fs::`.  That avoids worrying about some problems that this PR hits into, like not anchoring the regex or incorrect escape sequences.

 * in your `new_model()` change, you remove the extension _after_ the `check_for_existing_model()` call, but I think you need to do it before so that the `check_for_existing_model()` call is accurate.

 * the linked issue is about `new_model()`, and so are the added tests, stories, and requirements.  However, this PR is also touching `get-path-from-object.R` functions.

   Perhaps you went through the code base and decided a similar change was called for in other spots.  If that's the case, though, you should provide a summary of your analysis, but I don't spot anything in the commit messages or this PR or the linked issue.  In addition, these changes should also have tests.

 * related to the previous point: this PR defines `sanitize_file_extension()` twice, once in `new-model.R` and once in `get-path-from-object.R`, with two slightly different definitions.

inst/validation/bbr-requirements.yaml
```diff
@@ -106,6 +106,14 @@ CLM-R004:
...
   tests:
   - BBR-CLM-004
+CLM-R005:
+  description: accept '.ctl' file extension when calling new_model
+  tests:
+  - BBR-CLM-005
+CLM-R006:
+  description: accept '.mod' file extension when calling new_model
+  tests:
+  - BBR-CLM-006
```
kyleam, line 116: `CLM`?  Shouldn't the reqs and tests use `NWMD`?

inst/validation/bbr-stories.yaml
```diff
@@ -220,6 +220,13 @@ MGMT-S016:
   - CLM-R002
   - CLM-R003
   - CLM-R004
+MGMT-S017:
+  name: Passing model file extensions
+  description: As a user, I want to be able to pass '.ctl' and '.mod' file extensions when making a new model.
```
kyleam, line 225: This seems to be tightening up the "ignore extensions" request from gh-486.  Given that request and that that is what the code does, is there a reason to make this story more specific than "ignores extensions"?

inst/validation/bbr-requirements.yaml
```diff
@@ -106,6 +106,14 @@ CLM-R004:
   description: delete_models() with .tags=NULL
   tests:
   - BBR-CLM-004
+CLM-R005:
+  description: accept '.ctl' file extension when calling new_model
+  tests:
+  - BBR-CLM-005
+CLM-R006:
+  description: accept '.mod' file extension when calling new_model
```
kyleam, line 114: Even if you keep multiple test IDs (as I suggest elsewhere, you can roll them up), this can be a single requirement ID with multiple test IDs.

tests/testthat/test-new-model.R
```diff
@@ -146,3 +146,41 @@ test_that("new_model() supports `.path` containing a period [BBR-NWMD-010]", {
   mod <- new_model(fs::path_ext_remove(temp_ctl), "path with period")
   expect_true(fs::file_exists(temp_yaml))
 })
+
+test_that("new_model accepts .ctl file extension when passed [BBR-CLM-005]", {
```
kyleam, line 150: As a suggest elsewhere, I think it's fine to just test one extension, but if you keep testing both:

 * these test IDs can be rolled up into each other (e.g., "new_model ignores file extension: .ctl [BBR-CLM-005]" and "new_model ignores file extension: .mod [BBR-CLM-005]")
 * these are sharing a lot of code and it looks like all the assertions, so it'd be nice to avoid the duplication.  (See `BBR-SUM-003` for an example.)

tests/testthat/test-new-model.R
```diff
@@ -146,3 +146,41 @@ test_that("new_model() supports `.path` containing a period [BBR-NWMD-010]", {
...
+  expect_identical(mod_ctl$yaml_md5,"6ccf206e167485b5adf29bc135197929")
+})
+
+test_that("new_model accepts .mod file extension when passed [BBR-CLM-006]", {
+  clean_test_enviroment(create_rlg_models)
+
+    #Make a copy of the .ctl to test for .mod and remove .ctl file
+    file.copy(file.path(system.file("model", package = "bbr", mustWork = TRUE),"nonmem", "basic","1.ctl"),
+    file.path(system.file("model", package = "bbr", mustWork = TRUE), "nonmem","basic", "1.mod" ))
+    file.remove(file.path(system.file("model", package = "bbr", mustWork = TRUE),"nonmem","basic","1.ctl"))
```
kyleam, line 168: * indentation is off
* use `file.rename` or `fs::file_move` to handle this in one step?
* I'd suggest setting up an `on.exit` handler for restoring the state rather than doing it at the end (line 181-84).  Even better would be to work within a temporary directory rather than `inst/`, but I know a lot of bbr's tests already do that, so that's fine.

kyleam (changes requested):

R/new-model.R
```diff
@@ -50,6 +50,7 @@ new_model <- function(
 
   # check if file already exists and decide whether to overwrite if it does
   check_for_existing_model(.path, .overwrite)
+  .path <- sanitize_file_extension(.path)
```
kyleam, line 53: As I [mentioned before](https://github.com/metrumresearchgroup/bbr/pull/510#pullrequestreview-1029193785), it seems to me that this `sanitize_file_extension` call must happen _before_ the `check_for_existing_model` in order for that call to be valid.

R/new-model.R
```diff
@@ -90,7 +91,7 @@ new_model <- function(
 #' @seealso [copy_model_from()], [new_model()]
 #' @export
 read_model <- function(.path) {
-  yaml_path <- paste0(.path, ".yaml")
+  yaml_path <- paste0(.path %>% sanitize_file_extension(), ".yaml")
```
kyleam, line 94: Everything in here (tests, reqs, issue that prompted this) is about `new_model()`.  Is this `read_model()` change intentional?

inst/validation/bbr-requirements.yaml
```diff
@@ -614,6 +614,10 @@ NWMD-R010:
   description: new_model() supports `.path` containing a period
   tests:
   - BBR-NWMD-010
+NWMD-R011:
+    description: valid model extensions may be passed into new_model
```
kyleam, line 618: This is minor, but this reads to me like there is some extent to validate the extensions.  What about "new_model() ignores files extensions"?  (I suppose a similar comment can be made for the story description.)

inst/validation/bbr-stories.yaml
```diff
@@ -220,6 +220,13 @@ MGMT-S016:
   - CLM-R002
   - CLM-R003
   - CLM-R004
+MGMT-S017:
+  name: Passing model file extensions
+  description: As a user, I want to be able to pass valid file extensions for new models.
+  ProductRisk: Low
+  requirements:
+  - CLM-R005
+  - CLM-R006
```
kyleam, line 229: Looks like you forgot to update this to `NWMD-R011`.

tests/testthat/test-new-model.R
```diff
@@ -146,3 +146,29 @@ test_that("new_model() supports `.path` containing a period [BBR-NWMD-010]", {
...
+  mod <- new_model(file.path(system.file("model", package = "bbr", mustWork = TRUE), "nonmem", "basic", "1.mod"),
+                   .description = "original acop model",
+                   .tags = ORIG_TAGS,
+                   .bbi_args = list(overwrite = TRUE, threads = 4)
+  )
+  expect_identical(mod$absolute_model_path, file.path(system.file("model", package = "bbr", mustWork = TRUE), "nonmem", "basic", "1"))
+  expect_identical(mod$model_type,"nonmem")
+  expect_identical(mod$yaml_md5,"6ccf206e167485b5adf29bc135197929")
+  on.exit({fs::file_move(file.path(system.file("model", package = "bbr", mustWork = TRUE),"nonmem","basic","1.mod"),
+                         file.path(system.file("model", package = "bbr", mustWork = TRUE),"nonmem","basic","1.ctl"))})
```
kyleam, line 173: The `on.exit` should be placed as close as possible to the thing it is cleaning up so that, if any downstream code croaks, the cleanup is still triggered.  Placing it at the end of the function doesn't have the desired effect because the function is about to exit anyway.

Also, I'd suggest assigning these file names to variables since they're repeated across the two `file_move` calls.

And a nit-pick: This

```
file.path(system.file("model", package = "bbr", mustWork = TRUE),
          "nonmem", "basic", "1.ctl")
```

can be simplified to

```
system.file("model", "nonmem", "basic", "1.ctl",
            package = "bbr", mustWork = TRUE)
```

(ditto for several other spots above)

kyleam (changes requested):

inst/validation/bbr-requirements.yaml
kyleam: > hint that only .mod and .ctl would be accepted into new models.

I don't understand.  The current code just ignores all extensions.  What do we gain by not making this description match the behavior?

kyleam (approved):
Thanks for the updates.  LGTM

PR #511, opened 2022-07-06
Fix check_run_times and model_summary tests

barrettk:
This PR addresses bbr changes corresponding to this bbi PR: https://github.com/metrumresearchgroup/bbi/pull/270#event-6942798794

closes https://github.com/metrumresearchgroup/bbr/issues/509

PR #512, opened 2022-07-06
Check new bbi version on drone

barrettk:
DO NOT MERGE

This PR is just to inspect how drone works with the new version of bbi (3.2.0-alpha.1)


kyleam:
Tested (and incorporated into gh-511), so closing.  (Note that you can also trigger drone builds just by pushing a scratch/throwaway branch, no need to open the PR.)

PR #514, opened 2022-07-07
Setting Run Parrallel as Default

### LINKED ISSUE #385: parallel vs threads warning
There is an annoying situation where a user specifies more threads (e.g. through `.bbi_args(list(threads = 6))`) but forgets to add `parallel = TRUE` and therefore only one thread is used. This is especially common because `parallel` defaults to `false` in the `bbi.yaml` file.

I think we should either give a warning when `threads > 1 && parallel == FALSE`, or set `parallel: true` as the default going forward, or maybe both.

rymarinelli:
Closes #385


seth127:
@barrettk you can reference this group consensus that we documented in https://github.com/metrumresearchgroup/bbr/issues/385#issuecomment-1182171875

inst/validation/bbr-requirements.yaml
```diff
@@ -1034,6 +1034,10 @@ UTL-R013:
   description: wait_for_nonmem() doesn't error out if no stop time found
   tests:
   - BBR-UTL-013
+UTL-R014:
+  description: check_bbi_args() provides warning when `threads > 1` and parallel is `FALSE`
+  tests:
+  - BBR-UTL-014
```
barrettk, line 1040: I know all the tests basically now confirm that parallel is TRUE (in submit_model and the workflow tests, but I think we need an additional recursive test that specifically confirms:

- if threads = 1 and parallel not set: parallel is untouched (test with parallel = TRUE and FALSE)

So far we have:
- if threads >1 and parallel set to FALSE: parallel is untouched and warning is rendered
- if threads > 1 and parallel not set: parallel is set to TRUE (see comment below for this one)

tests/testthat/test-utils.R
```diff
@@ -122,9 +121,51 @@ test_that("strict_mode_error() works correctly [BBR-UTL-010]", {
...
+      #Testing that check_bbi_args is appending --parallel when not passed
+      expect_identical("--parallel", str_subset(res,"--parallel") %>% str_extract("--parallel"))
+      fs::file_delete(file.path(tempdir(), "test_path", "1.yaml"))
+
+
+      mod1 <-  new_model(file.path(tempdir(), "test_path", "1"), .description = "original test-workflow-bbi model",
+                         .tags = ORIG_TAGS,.bbi_args = list(overwrite = TRUE, threads = 2, parallel = FALSE))
+
+      #When `threads` > 1 and parallel is False raise a warning
+      expect_warning(submit_model(mod1, .dry_run = TRUE))
```
barrettk, line 165: When using `expect_warning()`, we want to have a specific warning we're testing for. If the warning does not match, we want it to be spit back to us.

Something like:
```r
expect_warning(submit_model(mod1, .dry_run = TRUE), "`threads > 1` but model will not run in parallel because `parallel = FALSE`")
```

Though note: I sometimes have to play around with some regex or `expect_warning()` arguments to get them to say they actually match (when they do)

tests/testthat/test-utils.R
```diff
@@ -122,9 +121,51 @@ test_that("strict_mode_error() works correctly [BBR-UTL-010]", {
...
+                        .tags = ORIG_TAGS,.bbi_args = list(overwrite = TRUE, threads = 2))
+
+      readr::write_file("created_by: test-utils", file.path(tempdir(), "test_path", "bbi.yaml"))
+
+
+      res <- capture.output(submit_model(mod1, .dry_run = TRUE))
+
+      #Testing that check_bbi_args is appending --parallel when not passed
+      expect_identical("--parallel", str_subset(res,"--parallel") %>% str_extract("--parallel"))
+      fs::file_delete(file.path(tempdir(), "test_path", "1.yaml"))
```
barrettk, line 158: I think this test right here addressed one of my previous concerns (if threads > 1 and parallel not set: parallel is set to TRUE). 

I would change the test description, as well as the description in the requirements yaml to be more broad, since we're technically testing for more than just the warning here. I think its fine to add all new tests (warning, setting parallel to TRUE, and leaving parallel alone) all to the same requirement you made, as long as the description encompasses those things. 

Could be something like:

```yaml
UTL-R014:
  description: when `threads > 1`, check_bbi_args() warns if parallel is set to `FALSE`, and automatically sets parallel to `TRUE` if not specified.
  tests:
  - BBR-UTL-014 (includes two tests)
  - BBR-UTL-015 (new regression test - see above comment)
```

barrettk (changes requested):
Looking good so far. Just some minor comments, and then I think we're ready to merge. Main thing is I think we need another regression test for the following case: `if threads = 1 and parallel _not_ set`
 - you could add another few lines to your existing test that handles this case, or make a new one. If you make a new one, I would separate the existing test where you wrote: 
   - > Testing that check_bbi_args is appending --parallel when not passed
 - For this test, I would test two cases: 1) when parallel is set to TRUE, and 2) when parallel is set to FALSE
    - Could be similar to what you have here:
      ```r
       res <- capture.output(submit_model(mod1, .dry_run = TRUE))

      #Testing that check_bbi_args is appending --parallel when not passed
      expect_identical("--parallel", str_subset(res,"--parallel") %>% str_extract("--parallel"))
      ```
         - where you just test that parallel isnt found (when set to FALSE), and is found (when set to TRUE). Should be a simple addition

tests/testthat/test-utils.R
rymarinelli: I was trying to keep this within one test because of all the set-up required. Is it alright to duplicate the code?

barrettk: Yeah I take your point. One option is to copy all those files over as part of the setup (see the first 34 lines of `tests/testthat/test-test-threads.R`), but you could include the additional test as part of that (up to you). 

Either way I would update the **requirement** **description** to be more general (since its not just about the warning). If you do choose to include all 3 tests under the same **test** **description**, I would make that one more general too

barrettk (changes requested):
Looking good. Only one minor comment (need someone to hop in to confirm the stories though - otherwise it might make sense to create a separate PR to add those other requirements to a story)


Also, could you merge in the changes from the main branch to get all the roxygen changes into this PR:
```r
git merge origin/main
# (then type ':wq' then 'return' to exit VIM or whatever that is)
git push --force-with-lease origin HEAD
```

run `pkgr install --update` after you do that too. I get the following on my end because I have the latest package versions without the docs merge:
image

Should be ready to merge

inst/validation/bbr-requirements.yaml
kyleam: > I noticed that none of the UTL-RXXX requirements have been added to any story. @kyleam do you know if thats by design?

Good question.  I don't spot these in the previous validation docs, which I think explains why they didn't get wired up by the mass import in https://github.com/metrumresearchgroup/bbr/pull/493

I suspect the answer is that, while these things are good things to test, they're not really conceptually tied to any stories.  I'm curious what @seth127's opinion is on this, but my view is that 1) we should test what we feel appropriate to test without worrying about whether it ties into story and 2) we shouldn't force a story just because there is a test.  As a consequence with both of those, I'm okay having tests that aren't linked, though in that case it would probably make sense to 1) drop the req and 2) drop the test ID.

barrettk: @kyleam I talked to Seth offline and he said that it was fine that the previous requirements weren't added to stories, but any _future_ requirements should be. I agree in terms of linking **tests** to stories, but I was under the impression that all **requirements** should be linked to a story (felt like that was important when rendering the docs - like it would fail when trying to map a requirement to a non-existent story). 


Based on my discussion with Seth though, it made sense to add this new requirement to a `submit_model` story, and leave the previous `UTL-RXXX` requirements alone. @rymarinelli After you make the requirement more general, can you add it to a story?

kyleam: > @kyleam I talked to Seth offline and he said that it was fine that the previous requirements weren't added to stories, but any _future_ requirements should be.

The fact that there are requirements for these at all is just an artifact of import process in the pr I linked above (gh-493).  I don't see a reason to keep them around.

> I agree in terms of linking **tests** to stories, but I was under the impression that all **requirements** should be linked to a story

In my reply, I wasn't suggesting to add requirements that aren't linked to a story (as I said, "it would probably make sense to 1) drop the req and 2) drop the test ID.").  So, I guess we're in agreement, but I'm not sure I understand what you're trying to say.

> (felt like that was important when rendering the docs - like it would fail when trying to map a requirement to a non-existent story. Assuming thats not the case though).

Again, I'm not saying we should add orphan reqs.  But if there are, no, things shouldn't fail altogether.  You can find the missing links with these helpers: https://metrumresearchgroup.github.io/mrgvalidate/reference/find_missing.html

barrettk: Oh no I wasnt suggesting you were saying anything lol, was just relaying my previous assumptions/concerns, mainly about issues with doc generation via mrgvalidate. Good to know about those helpers though, that definitely removes any concerns I had. 

But yeah im all for removing the other requirements in this PR. I guess it would make sense to remove the corresponding test ids too, but im also ok to leave them (can see reasons to remove and keep), but ill defer to you for that

kyleam: > But yeah im all for removing the other requirements in this PR.

Yeah, it is conceptually unrelated, so could be done elsewhere, but the main argument I see for doing it in this PR is that, assuming you're adding a requirement with the same stem, it avoids the val docs starting off at (... goes and greps ...) 015 while 001-014 are missing from the docs.

rymarinelli:
- [ ] Delete all UTL-RXXX requirements from bbr-requirements.yaml except yours (keep UTL-R014)
- [ ] Add UTL-R014 to a story surrounding submit_model. I think you can do CFG-S007

```
CFG-S007:
  name: Set execution mode globally
  description: As a user, I want parallel to be set to TRUE if threads are specified to be greater than 1.
  ProductRisk: Low
  requirements:
  - UTL-R014
 ```

barrettk (approved):
Looks good! Just run `devtools::document()` (I think I forgot to update the docs for `set_bbi_null` in my latest PR) and commit that before you merge

Issue #442, opened 2021-11-22
return omega/sigma matrix from model_summary object


callistosp:
## Problem:
As a user, I want to easily access the OMEGA matrix for a given model run. Currently the OMEGA half-matrix stored in the `model_summary` object is accessible through `modsum$parameters_data[[1]]$estimates$omega`. It would be more useful if we had a helper function that could easily return the full matrix. This would entail filling the top half the half-matrix currently in the vector and returning a `matrix` object.

seth127:
Should we have a similar helper for the SIGMA matrix?

Also, I assume you can get whatever you need related to THETA from either `param_estimates()` or `param_estimates_batch()`, but let us know if that's _not_ correct.

kylebaron:
THETA is not simple to get from `param_estimates()` (it's possible but not simple). If you have additional helpers to go get everything, it would be 5 functions that need to get called in order to extract this basic information. 

How about a single function call to get the estimates back from a given estimation method:

```r
> x$parameters_data[[1]]$estimates
$theta
[1]  0.4431030  4.1189900  1.1719500  4.2094100  1.2845500  0.4847440 -0.0377712  0.4189300

$omega
[1] 0.2187490 0.0668106 0.0823777 0.1214010 0.0703631 0.1144050

$sigma
[1] 0.0399375
```

`theta` has names and `omega` and `sigma` are matrices; matrices could have names but IMO less likely to need them.

callistosp:
I like the idea of getting theta, omega, and sigma all as a list from one function call, but I think a function that just returns the omega or sigma matrix would be nice too.

callistosp:
Thanks for the work on this guys! 

My initial thought is that the separate `get_*()` implementation should work fine. If someone wants a function which returns everything they could build a helper similar to what you describe, but I think the base case will be that people will just want to pull in one or two which will then be used separately (@kylebaron would love your opinion on this as well).

The reason why I have wanted to use something like this is to pull in the final estimates for some simulations. In this case I would also want to just pull in the estimates from the final estimation method step, so I agree that it would make sense to mirror the `param_estimates()` assumption of pulling in final estimates only.

barrettk:
Awesome thanks for the feedback so quickly! I will move forward with that assumption for now, but yes @kylebaron if you have any additional thoughts feel free to chime in

callistosp:
I think these look great, exactly the structure that I would hope to see for the outputs. I like the implementation for multiple models as well, that will be useful.

PR #515, opened 2022-07-11
get_* helpers for returning omega/sigma matrices and labeled theta values

barrettk:
# Proposed functionality:

### Single model summary
```r
sum <- MOD1 %>% model_summary()
```
```r
> sum %>% get_theta()
     THETA1      THETA2      THETA3      THETA4      THETA5 
  2.3412300  58.7268000 529.6200000  -0.0807956   4.1189400 


> sum %>% get_omega()
           OMEGA(1,1) OMEGA(1,2)
OMEGA(1,1)   0.104694   0.000000
OMEGA(2,1)   0.000000   0.159368


> sum %>% get_sigma()
           SIGMA(1,1)
SIGMA(1,1)          1
```

### With multiple model summaries
```r
mods <- list(MOD1, mod)
sums <- mods %>% model_summaries()
```
```r
> sums %>% get_theta()
[[1]]
     THETA1      THETA2      THETA3      THETA4      THETA5 
  2.3171600  54.6151000 462.5140000  -0.0820082   4.1795900 

[[2]]
     THETA1      THETA2      THETA3      THETA4      THETA5 
  2.3412300  58.7268000 529.6200000  -0.0807956   4.1189400 


> sums %>% get_omega()
[[1]]
           OMEGA(1,1) OMEGA(1,2)
OMEGA(1,1)  0.0985328   0.000000
OMEGA(2,1)  0.0000000   0.156825

[[2]]
           OMEGA(1,1) OMEGA(1,2)
OMEGA(1,1)   0.104694   0.000000
OMEGA(2,1)   0.000000   0.159368


> sums %>% get_sigma()
[[1]]
           SIGMA(1,1)
SIGMA(1,1)          1

[[2]]
           SIGMA(1,1)
SIGMA(1,1)          1
```

closes https://github.com/metrumresearchgroup/bbr/issues/442


seth127:
I like this approach. 

Do we have any models (maybe in `complex`?) that have off-diagonals for either OMEGA or SIGMA. I just want to make sure we can test against that to make sure we're parsing the matrices correctly.

kylebaron:
I think the names on the matrices should be refactored; it doesn't really make sense to have both the row and column index in the row name   ... the 2nd row is all OMEGA 2 and the 3rd column is all OMEGA 3.  

Something like

```r
[[1]]
           OMEGA(1)   OMEGA(2)
OMEGA(1)  0.0985328   0.000000
OMEGA(2)  0.0000000   0.156825
```

or make it easier to work with in R

```r
[[1]]
           OMEGA_1   OMEGA_2
OMEGA_1  0.0985328   0.000000
OMEGA_2  0.0000000   0.156825
```

barrettk:
> I like this approach.
> 
> Do we have any models (maybe in `complex`?) that have off-diagonals for either OMEGA or SIGMA. I just want to make sure we can test against that to make sure we're parsing the matrices correctly.

It seems `example2_saemimp` is the only one we can rely on. `acop-fake-bayes` and `acop-iov` do have the necessary info, but do not have any covariance data. Its possible that `1001` and `acop-onlysim` would based on the ctl files, but they are missing relevant data in the summary calls and cant be used.

```r
> sums %>% get_omega()
[[1]]
           OMEGA(1,1) OMEGA(1,2)
OMEGA(1,1)  0.0985328   0.000000
OMEGA(2,1)  0.0000000   0.156825

[[2]]
             OMEGA(1,1)  OMEGA(1,2)  OMEGA(1,3)   OMEGA(1,4)
OMEGA(1,1)  0.010410500 0.000319542 0.001977200 -0.000189864
OMEGA(2,1)  0.000319542 0.008278020 0.000872168  0.001005720
OMEGA(3,1)  0.001977200 0.000872168 0.013139700  0.004040370
OMEGA(4,1) -0.000189864 0.001005720 0.004040370  0.010987300
```

tests/testthat/test-get-param.R
```diff
@@ -0,0 +1,126 @@
...
+    values <- sum$parameters_data[[2]]$estimates$omega
+    labels <- sum$parameter_names$omega
+    matrix <- format_matrix(values, labels, .type = "OMEGA")
+
+    expect_true(isSymmetric(matrix)) # only confirmed in this test
+
+    par_df <- mod_complex %>% model_summary() %>% param_estimates() %>% filter(grepl("OMEGA", parameter_names))
+
+    check_matrix(matrix, par_df, .type = "OMEGA")
+  })
```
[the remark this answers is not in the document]
barrettk: For reference:
```r
# matches values reported in parameter estimates to location in matrix
# does not check symmetry
check_matrix <- function(matrix, par_df, .type = "OMEGA"){
  for(name.i in unique(par_df$parameter_names)){
    # message("spot checking ", name.i)
    spot_chk <- par_df$estimate[par_df$parameter_names == name.i]
    # Get positions from labels
    matrix_pos <- gsub(.type, "",name.i)
    matrix_pos <- gsub("[()]", "", matrix_pos)
    matrix_pos <- strsplit(matrix_pos, ",")
    row.i <- matrix_pos[[1]][1] %>% as.numeric()
    col.i <- matrix_pos[[1]][2] %>% as.numeric()

    expect_identical(matrix[row.i,col.i], spot_chk)
  }
}
```

R/get-param.R
```diff
@@ -0,0 +1,128 @@
+#' Return theta values as a named vector
+#'
+#' @param .summary  A `bbi_nonmem_summary` or `bbi_summary_list` object.
+#' @export
+get_theta <- function(.summary){
+
+  if (inherits(.summary, "bbi_nonmem_summary")) {
```
kyleam, line 7: Seems a bit odd to use "bbi_nonmem_summary" directly here and then the `NM_SUM_CLASS` indirection in the `check_model_object_list` right below.

R/get-param.R
```diff
@@ -0,0 +1,128 @@
...
+    .summary <- list(.summary)
+    check_model_object_list(.summary, .mod_types = c(NM_SUM_CLASS))
+  }else if(inherits(.summary, "bbi_summary_list")){
+    check_model_object(.summary, .mod_types = c("bbi_summary_list"))
```
kyleam, line 11: Given 1) `NM_SUM_CLASS` and 2) a `SL_CLASS <- "bbi_summary_list"` mapping exists, should `SL_CLASS` be used here?

R/get-param.R
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  omegas <- map(.summary, ~ {
+    # unpack bbi_summary_list element
+    if (!is.null(.x$bbi_summary)) .x <- .x$bbi_summary
+
+    param_names <- .x[[SUMMARY_PARAM_NAMES]]
+    omega <- with(
```
kyleam, line 48: The `with` documentation recommends against using in the programming context:

```
Note:

     For _interactive_ use this is very effective and nice to read.
     For _programming_ however, i.e., in one's functions, more care is
     needed, and typically one should refrain from using ‘with()’, as,
     e.g., variables in ‘data’ may accidentally override local
     variables, see the reference.
```

(Same comment for `get_sigma`.)

R/get-param.R
```diff
@@ -0,0 +1,128 @@
...
+  # Get positions from labels
+  matrix_pos <- gsub(.type, "",.labels)
+  matrix_pos <- gsub("[()]", "", matrix_pos)
+  matrix_pos <- strsplit(matrix_pos, ",")
+  .rows <- map(matrix_pos, function(.x){ .x[1] }) %>% unlist() %>% as.numeric()
+  .cols <- map(matrix_pos, function(.x){ .x[2] }) %>% unlist() %>% as.numeric()
+
+  # Assign values to corresponding location
+  .mat <- matrix(rep(0,max(.rows, .cols)*2), nrow = max(.rows), ncol = max(.cols))
+  for(i in 1:length(.rows)){
```
kyleam, line 115: `seq_along` would be better for guarding the 0 case, which is a good to do even if it should be unreachable.

R/get-param.R
```diff
@@ -0,0 +1,128 @@
...
+    .mat[.rows[i],.cols[i]] <- .values[i]
+    .mat[.cols[i],.rows[i]] <- .values[i]
+  }
+
+  # Assign row and column names
+  dimnames(.mat) <- list(
+    sprintf(paste0("%s_",seq(1,max(.cols)),""), .type),
+    sprintf(paste0("%s_",seq(1,max(.rows)),""), .type)
```
kyleam, line 123: Okay as is, but could be simplified (in my eyes) by dropping `paste0` and expanding format string to include `seq(...)` result.  Or, given this code base uses `glue` a lot, could even be `glue::glue("{.type}_{seq(...)}")`.

R/get-param.R
```diff
@@ -0,0 +1,128 @@
...
+format_matrix <- function(.values, .labels, .type = c("OMEGA", "SIGMA")){
+
+  .type <- match.arg(.type)
+
+  # Get positions from labels
+  matrix_pos <- gsub(.type, "",.labels)
+  matrix_pos <- gsub("[()]", "", matrix_pos)
+  matrix_pos <- strsplit(matrix_pos, ",")
+  .rows <- map(matrix_pos, function(.x){ .x[1] }) %>% unlist() %>% as.numeric()
+  .cols <- map(matrix_pos, function(.x){ .x[2] }) %>% unlist() %>% as.numeric()
```
kyleam, line 111: Okay as is, but you can simplify the index access: `map(matrix_pos, 1)`.

kyleam (commented):
Looks like a good start.  Here are comments from an initial read-through.  I plan to have at least one more round where I look more closely at the implementation and test the functions a bit.

R/get-param.R
barrettk: We have similar code [here](https://github.com/metrumresearchgroup/bbr/blob/main/R/test-threads.R#L180-L184), but I take your point. Its interesting we dont seem to have one for `"bbi_summary_list"`
```r
> NM_SUM_CLASS_LIST
[1] "bbi_nonmem_summary" "bbi_model"          "list"    
```

R/get-param.R
barrettk: Oh I take the above comment back, will make the change to `SL_CLASS`

R/get-param.R
```diff
@@ -0,0 +1,128 @@
...
+#' @param .summary  A `bbi_nonmem_summary` or `bbi_summary_list` object.
+#' @export
+get_theta <- function(.summary){
+
+  if (inherits(.summary, "bbi_nonmem_summary")) {
+    .summary <- list(.summary)
+    check_model_object_list(.summary, .mod_types = c(NM_SUM_CLASS))
+  }else if(inherits(.summary, "bbi_summary_list")){
+    check_model_object(.summary, .mod_types = c("bbi_summary_list"))
+  }
```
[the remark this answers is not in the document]
barrettk: I suppose we should have an else statement that stops the function if any other method is piped in. I like the handling we did [here](https://github.com/metrumresearchgroup/bbr/blob/main/R/test-threads.R#L180-L185), so i'll probably just improve upon that.

R/get-param.R
```diff
@@ -0,0 +1,128 @@
...
+      .x[[SUMMARY_PARAM_DATA]][[length(.x[[SUMMARY_PARAM_DATA]])]],
+      format_matrix(estimates$omega, param_names$omega, .type = "OMEGA")
+    )
+    omega
+  })
+
+
+  if(length(omegas) == 1) omegas <- omegas[[1]]
+
+  return(omegas)
```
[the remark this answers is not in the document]
barrettk: @kyleam Im now wondering if we want to export this other function, rather than making it internal. I currently have something like this:

image

KyleB was [suggesting](https://github.com/metrumresearchgroup/bbr/issues/442#issuecomment-976746226) something like this (though return everything)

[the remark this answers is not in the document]
kyleam: Hmm, but given it doesn't return everything, I'm not sure what functionality it adds to the interface, so my initial reaction is that it'd be better to leave it internal.  (And, when in doubt, I think internal is a good default because it's straightforward to expose something in a later release, but it's breaking to go in the other direction.)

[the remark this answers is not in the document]
barrettk: @kyleam I actually just refactored it again to be more in line with what Kyle was suggesting, so it now has some value, and returns an object similar to how `model_summaries()` does:

### Single summary
```r
> SUM1 %>% get_param()
$omega
          OMEGA_1  OMEGA_2
OMEGA_1 0.0985328 0.000000
OMEGA_2 0.0000000 0.156825

$sigma
        SIGMA_1
SIGMA_1       1

$theta
     THETA1      THETA2      THETA3      THETA4      THETA5 
  2.3171600  54.6151000 462.5140000  -0.0820082   4.1795900
```

### summary list
```r
> sums %>% get_param()
$omega
$omega[[1]]
          OMEGA_1  OMEGA_2
OMEGA_1 0.0985328 0.000000
OMEGA_2 0.0000000 0.156825

$omega[[2]]
             OMEGA_1     OMEGA_2     OMEGA_3      OMEGA_4
OMEGA_1  0.010410500 0.000319542 0.001977200 -0.000189864
OMEGA_2  0.000319542 0.008278020 0.000872168  0.001005720
OMEGA_3  0.001977200 0.000872168 0.013139700  0.004040370
OMEGA_4 -0.000189864 0.001005720 0.004040370  0.010987300


$sigma
$sigma[[1]]
        SIGMA_1
SIGMA_1       1

$sigma[[2]]
        SIGMA_1
SIGMA_1       1


$theta
$theta[[1]]
     THETA1      THETA2      THETA3      THETA4      THETA5 
  2.3171600  54.6151000 462.5140000  -0.0820082   4.1795900 

$theta[[2]]
    THETA1     THETA2     THETA3     THETA4     THETA5     THETA6     THETA7     THETA8     THETA9    THETA10    THETA11 
 3.3005000  3.2519700 -0.6122190 -0.2080120  0.7372940  1.1359900  0.3346290  0.1923520  0.6916970  2.2993700  0.0991684 
```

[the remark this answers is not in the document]
kyleam: > so it now has some value

That's true, though I'm not really a fan.  I think we should choose one approach, not expose a function that returns all of the parameters _and_ accessor functions for the individual parameters.  It complicates the interface and code.  

That's of course just one opinion.  Perhaps others disagree with me here.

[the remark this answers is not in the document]
barrettk: The main thing I like is being _**able to**_ retrieve all parameters, but not necessarily forced to. Perhaps I could make `"omega"` the default instead of all 3. If we chose to keep the 3 specific functions (and hide `get_param()`, I think I would still leave the ability to map across all 3 parameters just in case we want that functionality in the future.

@callistosp @kylebaron what do you guys think of the above change (see code snippets)? Does it make sense to have 1 general function, a 3 specific functions in terms of how you would/could use them?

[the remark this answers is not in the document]
seth127: I think I agree with @kyleam to err on the side of keeping it simple here (i.e. _don't_ add the extra helper). Also, I think Sam already weighed in on this [on the issue](https://github.com/metrumresearchgroup/bbr/issues/442#issuecomment-1180619733) and seemed to agree that the helper wouldn't likely be necessary in most cases.

barrettk:
@seth127 wondering if I could do this:

```r
#' @section Examples:
#'
#' ```{r, include = FALSE}
#' knitr::opts_chunk$set(collapse = TRUE, comment = "#>")
#' options('bbr.bbi_exe_path' = read_bbi_path())
#' MODEL_DIR <- system.file("model", "nonmem", "complex",   package = "bbr")
#' mod <- read_model(file.path(MODEL_DIR, "example2_saemimp"))
#' ```
#'
#' ```{r, comment = "#>", collapse = TRUE}
#' sum <- mod %>% model_summary()
#'
#' sum %>% get_omega()
#'
#'
#' sum %>% get_sigma()
#'
#'
#' sum %>% get_theta()
#' ```
```

I think this will appease drone, but im worried about how this helper doc would render after someone installed the package on their end (without running `use_bbi()`). I would add hidden conditional logic that would use that function if not found at that path, but I have a feeling there are reasons we wouldnt want to do that. A lot of this package doesn't have examples (which I think would be nice), but I can see the difficulty in making them for most functions without having bbi installed. Have thought about this for a while, but wondering what your thoughts are (in this case, and having examples relying on bbi in general)

seth127:
Good thoughts there @barrettk . At this point, I think our best move is to just do examples and make them `\dontrun{`. The main downside of that is that CRAN doesn't like it, but getting `bbr` on CRAN really isn't a priority because pretty much all likely users are gonna get it from MPN anyway. We already use this pattern in other places and you're right that we should probably add more examples for other functions.

> maybe having some version of bbi installed in inst or something (Im sure thats a terrible idea)

Definitely _not_ a terrible idea. I've thought about this a few times, not `inst` specifically, but somehow having a version of `bbi` included in `bbr`. Every time I run through it in my head, I end up deciding it's not a good idea, but I do keep coming back to it. There are a couple big advantages, a couple off the top:

* streamline getting started for new users
* `bbi` version pinned (to something that's compatible) by default and users only change it if they have a reason to.
* would make it easier for regulators or whoever else that wants to re-run code without having to work hard to recreate the environment.
* the test suite and example stuff that you pointed out

Anyway, something to consider at some point, though I doubt we'll bite it off anytime soon. Lemme know if you want to talk further at some point.

R/get-param.R
```diff
@@ -0,0 +1,155 @@
...
+#'
+#'
+#' sum %>% get_sigma()
+#'          SIGMA_1
+#'  SIGMA_1       1
+#'
+#'
+#' sum %>% get_theta()
+#'     THETA1     THETA2     THETA3     THETA4     THETA5     THETA6     THETA7     THETA8     THETA9    THETA10    THETA11
+#'  3.3005000  3.2519700 -0.6122190 -0.2080120  0.7372940  1.1359900  0.3346290  0.1923520  0.6916970  2.2993700  0.0991684
```
kyleam, line 86: This comes triggers a note:

```
❯ checking Rd line widths ... NOTE
  Rd file 'get_omega.Rd':
    \examples lines wider than 100 characters:
           THETA1     THETA2     THETA3     THETA4     THETA5     THETA6     THETA7     THETA8     THETA9    THETA10    THETA11
        3.3005000  3.2519700 -0.6122190 -0.2080120  0.7372940  1.1359900  0.3346290  0.1923520  0.6916970  2.2993700  0.0991684
```

That's not really a problem, though it is nice to keep the check output as quiet as possible.  Maybe it'd be enough to just show, say, the `get_omega()` calls.  An example is nice, but once you've seen one, I'm not sure seeing each other one adds much value.  Up to you, just a thought.

R/get-param.R
```diff
@@ -0,0 +1,155 @@
...
+  }else{
+    stop("Please pass in a `", NM_SUM_CLASS, "` or `", SL_CLASS, "` type object.")
+  }
+
+  assert_true(all(.param %in% c("omega", "sigma", "theta")))
+
+  param_obj <- map(.param, function(param.i){
+
+    if(param.i %in% c("omega", "sigma")){
+      param_obj_i <- map(.summary, ~ {
```
kyleam, line 25: Missing `map` `@importFrom` for this function.

tests/testthat/test-get-param.R
```diff
@@ -0,0 +1,113 @@
+context("Test get_omega, get_sigma, and get_theta functions")
+
+skip_if_not_drone_or_metworx("test-get-param")
+
+mod_complex <- read_model(file.path(MODEL_DIR_X, "example2_saemimp"))
+
+
+# matches values reported in parameter estimates to location in matrix
+# does not check symmetry
```
kyleam, line 9: Stale comment as of 620b1064?

tests/testthat/test-get-param.R
```diff
@@ -0,0 +1,113 @@
...
+    expect_equal(par_df2$estimate, unname(thetas[[2]]))
+  })
+
+
+  test_that("get_omega() returns correct values and labels: model_summary [BBR-PEST-017]", {
+    clean_test_enviroment()
+    omegas <- MOD1 %>% model_summary() %>% get_omega()
+    par_df <- MOD1 %>% model_summary() %>% param_estimates() %>% filter(grepl("OMEGA", parameter_names))
+
+    # order is preserved, and duplicate coviarances are dropped
```
kyleam, line 70: Typo:  "coviarances" (here and in other spots)

R/get-param.R
```diff
@@ -0,0 +1,155 @@
...
+#' @param .summary a `bbi_nonmem_summary` or `bbi_summary_list` object.
+#' @param .param the parameter(s) to retrieve and format. Any subset of `c("omega", "sigma", "theta")`
+#'
+#' @importFrom stats setNames
+#' @keywords internal
+get_param <- function(.summary, .param = c("omega", "sigma", "theta")){
+
+  if (inherits(.summary, NM_SUM_CLASS)) {
+    .summary <- list(.summary)
+    check_model_object_list(.summary, .mod_types = c(NM_SUM_CLASS))
```
kyleam, line 13: I'm confused by `check_model_object_list` is achieving here.  In the condition right above, you just did the same `inherits` check on `NM_SUM_CLASS` that `check_model_object_list` did underneath.  So isn't the failure unreachable?

https://github.com/metrumresearchgroup/bbr/blob/033899436217eb6a24083f97ae2063933b75a828/R/utils.R#L385-L386

Similar comment for the `check_model_object` call for the next condition.

kyleam (commented):
Thanks for the update.  A few more minor comments, but looking almost ready to me.

R/get-param.R
barrettk: Yes I had this thought after merging the test_threads PR a while back (used in `check_run_times()`). I forget how I had it originally, but Seth and I came up with that together. I dont think either of us fully inspected what the `check_` functions were doing at that point (we probably assumed it did more than confirm the class). In the case of `check_run_times()` we wanted to ensure a model object, model list, model summary, or model summary list object was passed in, and in this case we only want to allow a model summary or model summary list. Because of classes we weren't able to use solely `check_model_object` or `check_model_object_list`.



Ill rework this to make it actually meaningful. Will likely ditch the check_ functions altogether

kyleam: Okay, to be clear, not a big deal, I was just scratching my head.

barrettk: No worries. Made a change that I think improves the logic. Will hold off on touching the other function.

kyleam (approved):
Update looks good to me.  Thanks.

PR #519, opened 2022-07-14
max_eval overwrites estimation methods for complex models

barrettk:
Previously, test_threads had not been tested with complex models (models with multiple estimation methods). Updating `MAXEVAL/NITER`, would cause the second estimation method to inherit the first. Additionally, in cases where the first estimation method used `MAXEVAL`, and the second one used `NITER`, both would have `MAXEVAL` (not illustrated)

The below example illustrates the behavior, where the second `$EST` line became SAEM instead of IMP

Original ctl file:
```
$EST METHOD=SAEM INT EONLY=0 NBURN=3000 NITER=250  PRINT=5  GRD=DDDDDDDDDDDD CTYPE=3 SEED=2021  RANMETHOD=P
CALPHA=0.05 ISAMPLE=2
$EST METHOD=IMP INT EONLY=1 NITER=4 ISAMPLE=3000 PRINT=1  GRD=DDDDDDDDDDDD RANMETHOD=P  SEED=2021 CTYPE=3  FILE=28h.EXT
CALPHA=0.05 MSFO=28h.MSF
$COV UNCONDITIONAL MATRIX=R PRINT=E
```

The ctl file generated by the test_threads function:
```
$EST METHOD=SAEM INT EONLY=0 NBURN=3000  PRINT=5  GRD=DDDDDDDDDDDD CTYPE=3 SEED=2021  RANMETHOD=P NITER=10
CALPHA=0.05 ISAMPLE=2
$EST METHOD=SAEM INT EONLY=0 NBURN=3000  PRINT=5  GRD=DDDDDDDDDDDD CTYPE=3 SEED=2021  RANMETHOD=P NITER=10
CALPHA=0.05 MSFO=28h.MSF
$COV UNCONDITIONAL MATRIX=R PRINT=E
```


barrettk:
Not sure why drone was failing here in the previous commit. `typed.missing = TRUE` is a valid argument, and prevents users from passing `NA` and stuff like that (`asser_numeric` will allow that by default). Its not necessary, so I just removed it to appease drone.

image

barrettk:
@seth127 I requested Kyle to review this PR, but tagging you in case you'd like to take a look/be involved

kyleam:
> Not sure why drone was failing here in the previous commit.

The [`mpn:oldest` build](https://github-drone.metrumrg.com/metrumresearchgroup/bbr/2884/2/1) was failing.  That uses the 2020-06-08, which has checkmate 2.0.0:

```
$ curl -fSsL https://mpn.metworx.com/snapshots/stable/2020-06-08/src/contrib/PACKAGES | grep -A1 'Package: checkmate'
Package: checkmate
Version: 2.0.0
```

The `typed.missing` argument wasn't added to checkmate until 434a2cf (Flag to control handling of missing values, 2020-10-14), which was a part of the v2.1.0 release.

So, dropping the argument (as you've already done) is the way to go, unless we bump the minimum MPN that we test against.

seth127:
Thanks @barrettk . From a quick glance, this all looks good to me, but I'll let @kyleam do a real review.

tests/testthat/test-test-threads.R
```diff
@@ -67,6 +74,61 @@ withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
...
+
+    search_str <- "MAXEVAL|NITER"
+    max_evals <- map(mods_complex, function(.mod){
+      mod_path <- get_model_path(.mod)
+      mod_lines <- mod_path %>% readLines()
+      str_line_loc <- which(grepl(search_str, mod_lines))
+      str_values <- str_split(mod_lines[str_line_loc], " ")
+
+      values <- map(1:length(str_values), ~ {
+        est_loc <- grepl("METHOD", str_values[[.x]])
```
kyleam, line 89: I have two minor comments on the new tests, but I think they're fine as is.

  * `expect_identical` would be more appropriate for these assertions than `expect_equal` unless you need to tolerate some small numeric difference.  (I'm not very consistent about this, either, though.)

  * the logic in these tests feel a bit too close to the logic in the code.  Rather than repeating the logic and risking some of the same bugs in both places, I'd prefer dumber test code that's more straightforward (even if repetitive).  For example, instead of looping and doing complicated parsing, read in the model path for each model separate, assert that a couple of expected lines are there and maybe verify that the expect number of lines match "MAXEVAL|NITER".

Just my two cents.  Again, the tests look fine to me overall.

kyleam (dismissed):
The PR description does a good job of providing the background/context, thanks.  This looks almost ready to go.

----

Docs need to be regenerated:

```
$ git diff --name-only
$ Rscript -e 'devtools::document()'
ℹ Updating bbr documentation
ℹ Loading bbr
Writing test_threads.Rd

$ git diff --name-only
man/test_threads.Rd
```

R/test-threads.R
```diff
@@ -72,30 +73,38 @@ test_threads <- function(
...
-    mod_path <- get_model_path(.mod)
-    mod_lines <- mod_path %>% readLines() %>% suppressSpecificWarning("incomplete final line found")
-    str_line_loc <- which(grepl(search_str, mod_lines))
-
-    if(is_empty(str_line_loc)){
-      delete_models(.mods = .mods, .force = TRUE) %>% suppressMessages()
-      stop("Neither MAXEVAL or NITER were found in the ctl file. Please ensure one is provided.")
-    }
+  if(!is.null(.max_eval)){
+    assert_numeric(.max_eval, lower = 1, upper = 9999)
```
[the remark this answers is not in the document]
barrettk: I must have not prefaced with `checkmate::`, because I was looking for that function. Yes we definitely want integer

tests/testthat/test-test-threads.R
barrettk: Made the change to `expect_identical`. I take your point on feeling similar to the function code. Thought about that as well, but feel the test expectations (names and values) cover any issues that could come up (in terms of test quality). I mainly like that you get a nice neat object before the tests, that you can inspect if any issues were to arise from the function misbehaving.

Has to be:
```r
> max_evals
[[1]]
  METHOD=1, MAXEVAL METHOD=BAYES, NITER 
                100                 100 

[[2]]
  METHOD=1, MAXEVAL METHOD=BAYES, NITER 
                100                 100 
```
which presents & confirms A) the method was preserved, B) `MAXEVAL`/`NITER` was preserved, and C) the values were correctly changed. I could remove the looping part, but Id still have to parse the files to find those lines and `gsub` the number from them, so I feel the majority of the code would still look similar to what it is now, just without the loop.

kyleam (approved):
LGTM (and drone is now happy, and all tests pass for me on metworx too).  Thanks, @barrettk.

PR #522, opened 2022-07-20
model-diff: hide content when models are identical

### LINKED ISSUE #516: change model_diff() output when files are identical
Currently, when the files `model_diff()` is comparing are identical, it prints out a message saying there are identical and then prints out the entire contents of the files. This is because this is the behavior of the `diffobj::diffFile()` function that we use under the hood, but I find it to be confusing and annoying. 

Consider changing this to catch these cases and _not_ print out the entire file contents (instead printing only the message).

kyleam:
When passed files with identical contents, diffobj::diffFile() has a
message that says "No visible differences between objects" but then
confusingly prints a diff with the file content to the console.  Catch
this case, give a message, and return early.

Closes #516.

### DIFF
--- a/R/model-diff.R
+++ b/R/model-diff.R
@@ -117,7 +117,7 @@ model_diff_impl <- function(model_A, model_B, .viewer) {
   }
 
   if (tools::md5sum(model_A) == tools::md5sum(model_B)) {
-    message("Models are identical")
+    message("Relevant model files are identical")
     return(invisible(NULL))
   }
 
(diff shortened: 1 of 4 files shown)


kyleam:
> diffobj::diffFile() has a
> message that says "No visible differences between objects" but then
> confusingly prints a diff with the file content to the console

Here's an example:

seth127 (approved):
This looks great and clean, though let's change the message to something like "Relevant model files are identical". I'm mainly thinking of Stan models where you could be comparing different files and "Models are identical" could be misleading if, for example, the .stan files are identical but other files are not.

kyleam:
> let's change the message to something like "Relevant model files are identical"

Done.  Thanks, @seth127.

Issue #520, opened 2022-07-15
Replace all BBI_NULL_NUM with NA


seth127:
`bbi`  passes back `-999999999` (defined [here](https://github.com/metrumresearchgroup/bbi/blob/main/parsers/nmparser/structs.go#L186-L199)) as a "default" in many cases. Presumably this was done based on some NONMEM convention, but in most (all?) cases it seems to correspond what would more intuitively be `NA` in R.

The proposal here would be to have `model_summary()` (and any other relevant functions that parse `bbi` output) to replace [`BBI_NULL_NUM` and `BBI_NULL_STR`](https://github.com/metrumresearchgroup/bbr/blob/main/R/aaa.R#L171-L172) with `NA_real_` and `NA_character_` respectively, at parse time.

Investigations, we should do before making this change:

* Look through `bbi` to see if we can find cases where these values are returned that _don't_ make sense to map to `NA` in R.
* Consult some users on whether `-999999999` actually _would_ be an expected/desirable value to get in R. In my experience talking with users, they either are confused when they see `-999999999`, or they are used to seeing it and assume it means `NA`.

seth127:
You can see three examples in our `model_summary()` test reference object: [here](https://github.com/metrumresearchgroup/bbr/blob/main/inst/test-refs/1_summary_obj.R#L4), [here](https://github.com/metrumresearchgroup/bbr/blob/main/inst/test-refs/1_summary_obj.R#L10), and [here](https://github.com/metrumresearchgroup/bbr/blob/main/inst/test-refs/1_summary_obj.R#L60).

Probably unrelated: the first two of those are `run_start` and `mod_file` and it's not clear to me why `bbi` can't parse these things in this case. That may be worth looking into, but again I think that's a separate issue.

kyleam:
Related to https://github.com/metrumresearchgroup/bbr/issues/518

seth127:
@callistosp @timwaterhouse @kylebaron @janelleh any thoughts on this? Have you noticed the `-999999999` coming through before? If so, is that desirable in any cases you can think of, or would you rather have `NA`?

janelleh:
[...]

Issue #518, opened 2022-07-14
check_run_times: handle BBI_NULL_NUM special case?


kyleam:
It's possible for the value of the `postprocess_time` or `cpu_time` field to come through as `BBI_NULL_NUM` if no values are parsed from the model output.  Should `check_run_times` catch this special case and convert it into something a little more obvious to the user (perhaps `NA_real_`)?

https://github.com/metrumresearchgroup/bbr/blob/f31d080e8e5fba97956fe0b1d4b8a69f1f8ec22d/R/test-threads.R#L212-L213

cc: @barrettk

barrettk:
tbh im not sure what's desired. I would vote to get a scientists opinion on this. 

@kylebaron, @callistosp do you guys have any thoughts?

For reference:
```r
> BBI_NULL_NUM
[1] -1e+09
```

barrettk:
@kyleam btw im not able to load that link you attached:
![image](https://user-images.githubusercontent.com/18685116/179032103-c49d4c65-0023-4563-a047-a11b0adb6784.png)

kyleam:
> @kyleam btw im not able to load that link you attached:

Ah, thanks. That was derived from an unpushed commit.  Updated to be based off `main`.

kyleam:
https://github.com/metrumresearchgroup/bbr/issues/520 proposes doing this replacement tree-wide

PR #524, opened 2022-07-21
Replace all BBI_NULL_NUM with NA

barrettk:
`bbi`  passes back `-999999999` (defined [here](https://github.com/metrumresearchgroup/bbi/blob/main/parsers/nmparser/structs.go#L186-L199)) in `bbi` as a "default" in many cases. Variables are set [here](https://github.com/metrumresearchgroup/bbr/blob/main/R/aaa.R#L171-L172) in `bbr`.

## To do:
 - modify `model_summary()` and other functions to rely on new internal functions for the replacement
    -  update `check_run_times` (though this might fix itself if the change happens in `model_summary()`)
 - modify test reference objects
 - modify tests as needed
 - add a new recursive test (we were already replacing `BBI_NULL_NUM` with `NA_real_` for objective functions and checking for that)

closes https://github.com/metrumresearchgroup/bbr/issues/520
closes https://github.com/metrumresearchgroup/bbr/issues/518


kyleam:
Hmm, @barrettk and @seth127 my gut feeling is that we should hold off on this for this release (just saw that that the issue got tagged for 1.4.0).   One issue is whether we should audit all the places where this comes out of bbi, but in general I think it wouldn't necessarily be a bad idea to let the idea sit.

barrettk:
> Hmm, @barrettk and @seth127 my gut feeling is that we should hold off on this for this release (just saw that that the issue got tagged for 1.4.0). One consideration is whether we should audit all the places where this comes out of bbi and be more targeted with the replacement, but in general I think it wouldn't necessarily be a bad idea to let the idea sit.


I'll let Seth answer this, as Im only tackling it because we talked about it during our Monday meeting. I will say that the functions I implemented do a great job at fixing this, and so far requires pretty little intervention other than changing a few tests/test files and adding a call to `model_summary()`:

```r
  res_list <- combine_list_objects(res_list, bbi_list)

  res_list <- map_list_recursive(res_list, set_bbi_null) # new line

  res_list <- create_summary_object(res_list)
```

It might make more sense to do in bbi, but its not as invasive as I thought it would be if we do it in bbr

kyleam:
> It might make more sense to do in bbi [...]

I'm not really proposing that.  I'm suggesting that we might want to be more focused and selective in what we map, unlike the approach here.  Or not. And we might want to consider performance hit here when there are lots of values for `estimates`.  Again, I don't know, but this is not a new problem, and my preference would be to hold off on it for 1.4.0.

barrettk:
Ahh, I just inferred that from

> One consideration is whether we should audit all the places where this comes out of bbi

But yeah I take your point. In terms of waiting or not I have no comment, but performance wise Im not _too_ concerned. Not really sure how long a list of estimates could become, but I saw no difference when working with the example models (I know they're simpler). If warranted I could benchmark the time difference (before and after this implementation), but it would probably make more sense to do that with a bootstrap run or something? 

I'll hold off on writing any tests until Seth responds, but at least the core change (or a version of it) is done (existing tests seem to be passing now).

barrettk:
@kyleam I've been struggling to write a test for this. Tried a number of things to try to filter the list to only the values that get overwritten. At this point I have little to show for, and to avoid going further down the rabbit hole im wondering if you think updating the summary objects was good enough, or if we still want some type of additional recursive test

barrettk:
@seth127 @kyleam 

Feel like I sunk too much time into this for not a lot of value, but at least we have a working test now. Main downside is that I "had to" (couldnt figure out another way) add another arg to `model_summary` to be able to test this. After putting so much time into trying to evaluate a sensible comparison (the tests), I at least wanted to document the option in case it was of value. 

### Pros/cons
On one hand, this new arg would at least allow users to **_not_** overwrite these values. If they had a feeling something was wrong, they could inspect it more easily and make sure nothing wonky was going on with NA replacement. God forbid anything weird did happen, we could have a much easier way to figure out what's going on, and the scientists may feel a little better about the change (not that they were worried).

<br>

On the other hand, im not sure we want to draw attention to anything bbi related (like BBI_NULL_NUM), and this arg may make some scientists (mostly external in my mind) raise an eyebrow.

### The options are:
1. this test and new arg are ok (assuming not)
2. we dont want this new `model_summary` arg, but we still want a recursive test (might need some guidance if so)
3. the model_summary reference objects are enough, and we can just delete all the changes I made in the latest commit

Let me know what you guys think (or whoever responds first)

seth127:
Talked about this with @barrettk offline. I think that test (and the associated arg/helpers) are overkill. Going with this:

> 3. the model_summary reference objects are enough, and we can just delete all the changes I made in the latest commit

inst/test-refs/1_summary_obj.R
```diff
@@ -1,13 +1,14 @@
 structure(list(
   absolute_model_path = "/data/home/sethg/bbr/inst/model/nonmem/basic/1",
   run_details = list(
-    version = "7.4.4", run_start = "-999999999",
+    version = "7.4.4", run_start = NA_character_, #"-999999999",
```
seth127, line 4: I'm not sure where these commented `999...` are coming from in this file, but I think you just want to regenerate this fresh with `source("data-raw/build-model-summary-refs.R")`

R/utils.R
```diff
@@ -598,3 +598,49 @@ wait_for_nonmem.list <- function(.mod, .time_limit = 300, .interval = 5) {
...
+#'
+#' @param .vals a numeric or character vector
+#'
+#' @importFrom purrr map
+#' @keywords internal
+set_bbi_null <- function(.vals){
+  map(.vals, function(.val){
+    if(inherits(.val, "character") && .val == BBI_NULL_STR && !is.na(.val)){
+      return(NA_character_)
+    }else if(inherits(.val, "numeric") && .val == BBI_NULL_NUM && !is.na(.val)){
```
seth127, line 612: I don't think we need one of these code paths. It looks like the `-999999999` comes in through from the json as an integer (based on the [change to the ofv test](https://github.com/metrumresearchgroup/bbr/pull/524/files#diff-ed82dffdf3c54e67fe827da182bf90b01bf71102844ef0323833bf3f572bcff7R214)), so this block doesn't ever happen. And even if it did sometimes, I don't think we could predict when it would happen in a way that makes this `NA_real_` vs `NA_integer_` distinction meaningful.

My gut is to go with `NA_real_` because it's more general. @kyleam do you have any opinion on that?

tests/testthat/test-model-summary.R
```diff
@@ -211,10 +211,80 @@ withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
 
       ofvs <- model_summary(mod)[[OFV_COL]]
       ofv <- ofvs[[length(ofvs)]]
-      expect_identical(ofv[["ofv_no_constant"]], NA_real_)
-      expect_identical(ofv[["ofv_with_constant"]], NA_real_)
+      expect_identical(ofv[["ofv_no_constant"]], NA_integer_)
```
seth127, line 214: This is discussed in another comment above, but I think we want to change this back.

seth127 (changes requested):
As noted [above](https://github.com/metrumresearchgroup/bbr/pull/524#issuecomment-1201351012) I think we should revert this branch back to a4a0141f274a856975992eafcfe06350e1e96182

A few other small comments, but I think this will be pretty much there, once we revert.

R/utils.R
kyleam: I agree, I don't think we gain anything with the `NA_integer_` vs `NA_real_` distinction, and think it makes sense to go with `NA_real_`.

R/model-summary.R
```diff
@@ -134,12 +134,14 @@
 #' See [print_bbi_args()] for full list of options.
 #' @param ... args passed through to [bbi_exec()]
 #' @param .dry_run show what the command would be without actually running it
+#' @param .overwrite whether or not to overwrite values of error/missing values (defaults to -1e+09) with `NA`
```
kyleam, line 137: I suggest writing out the nines to avoid any confusion about whether the value is `-999999999` or `-1000000000`

```
> -999999999
[1] -1e+09
> -999999999L
[1] -999999999
> -1e9
[1] -1e+09
> -1e9L
[1] -1000000000
> identical(-1000000000, -1e9)
[1] TRUE
> identical(-1000000000L, -1e9L)
[1] TRUE
```

R/utils.R
barrettk: @kyleam @seth127 I do have to have the integer and numeric for `inherits()` (will combine integer and numeric), but I can change the return value to be `NA_real_`. Will change the corresponding tests too

inst/test-refs/1_summary_obj.R
barrettk: ooh good catch jeez. Yeah I modified this directly and forgot it was generated

R/model-summary.R
barrettk: Interesting, didnt realize that. I reverted back a commit, so that new argument wont be there, but sounds good in general. Will look for cases of that

seth127: 🤯

seth127 (approved):
Looks good. Thanks.

R/model-summary.R
```diff
@@ -230,6 +230,8 @@ nonmem_summary <- function(
 
   res_list <- combine_list_objects(res_list, bbi_list)
 
+  res_list <- map_list_recursive(res_list, set_bbi_null)
+
```
kyleam, line 234: What do you think about moving this to `create_summary_object`?  That way gh-527 won't need to be adjusted to make a similar call.

R/summary-log.R
```diff
@@ -218,7 +218,7 @@ extract_condition_number <- function(.s) {
   })
   .out <- map_dbl(.ofv, function(.x) {
     .cn <- .x[[SUMMARY_COND_NUM]]
-    if (is.null(.cn) || .cn == BBI_NULL_NUM) {
+    if (is.null(.cn) || .cn == BBI_NULL_NUM || is.na(.cn)) {
```
kyleam, line 221: Should `BBI_NULL_NUM` ever get this this point (i.e. is it still needed in this condition)?  Without looking closely at this, it seems to me that this could be reduced to an `is.null` check.

R/utils.R
```diff
@@ -598,3 +598,46 @@ wait_for_nonmem.list <- function(.mod, .time_limit = 300, .interval = 5) {
...
+
+#' Replace BBI_NULL_NUM and BBI_NULL_STR with NA_real_ or NA_character_
+#'
+#' @param .vals a numeric or character vector
+#'
+#' @importFrom purrr map
+#' @keywords internal
+set_bbi_null <- function(.vals){
+  map(.vals, function(.val){
+    if(inherits(.val, c("integer", "numeric", "character")) && (.val == BBI_NULL_NUM)){
```
kyleam, line 610: When I read https://github.com/metrumresearchgroup/bbr/pull/524#discussion_r935081383, I thought is was just talking about the `_real` and `_integer` distinction.  Should "character" be included here?

R/utils.R
```diff
@@ -598,3 +598,46 @@ wait_for_nonmem.list <- function(.mod, .time_limit = 300, .interval = 5) {
...
+      return(NA_real_)
+    }else{
+      return(.val)
+    }
+  }) %>% unlist()
```
kyleam, line 615: Seems worth considering whether we're happy to just go with `NA` and just let type coercion do it's thing.  In that case, this whole function could be reduced to two lines:

```
  is.na(.vals[.vals == BBI_NULL_NUM]) <- TRUE
  .vals
```

Depending on the other values in the vector, that could lead to an `NA_integer_` when a `NA_real_` would be more appropriate, but I don't think that's a problem and the simplicity is nice.  (You could of course also keep the vector subsetting approach and add more conditions to handle the different types.)

R/summary-log.R
barrettk: Id still need the new NA call (evaluated to `logical(0)` without it), but I could likely get rid of the `BBI_NULL_NUM` part

R/model-summary.R
barrettk: I looked into this earlier, but `model_summaries()` shouldn't need to be adjusted since it maps over `model_summary` (see below). Without doing too much of an investigation, it still seems like you're calling out to that in your PR. Am I missing something?
image

kyleam: `model_summaries.list` branches based on the bbi version:

https://github.com/metrumresearchgroup/bbr/blob/ce785a40b9b5c5bc0d00a4fccd75dc227104f1ba/R/model-summaries.R#L178-L182

`model_summaries_concurrent` doesn't go through `model_summary`.

barrettk: Ahh, I didnt catch that. Ok yeah Ill nest that function inside `combine_list_objects` then. @seth127 pinging you so you're aware

kyleam: >  Ill nest that function inside combine_list_objects then

Why `combine_list_objects` (which is a general utility) rather than `create_summary_object`?

R/summary-log.R
barrettk: It had something to do with x being `NA`. If I type in `is.null(NA_real_)` I get `FALSE`, but I know I had to add that change because that function was evaluating to that after `BBI_NULL_NUM` was changed. Ill see if I can reproduce that case

R/utils.R
barrettk: Im going to keep the functions the way they are for now, as I think the logic is easier to follow in the first function. Definitely biased, but I also like having the new custom function, and think it will be useful in the future based on how its set up (main reason I chose to do it that way - though could have made it a map call instead of a for-loop)

R/model-summary.R
barrettk: Made this change in latest commit

R/summary-log.R
barrettk: Made this change in latest commit (wasnt able to reproduce the issue, so not 100% sure why it works now)

R/utils.R
barrettk: Yes, some occurrences of `BBI_NULL_NUM` are characters beforehand (see [here](https://github.com/metrumresearchgroup/bbr/blob/main/inst/test-refs/1_summary_obj.R#L10)). Updated the docs

kyleam: > Yes, some occurrences of BBI_NULL_NUM are characters beforehand (see [here](https://github.com/metrumresearchgroup/bbr/blob/main/inst/test-refs/1_summary_obj.R#L10)).

Er, right.  My question is whether that's the intended behavior.  As I said, that wasn't how I read the initial comment I linked to.  It was just about whether to distinguish integer vs real.

As I mentioned elsewhere, my preference would be to just assign NA to it and keep whatever the current vector type is without any of these `inherit` checks.  If we do keep the `inherit` checks, then I think it's valuable to keep the character distinction.

barrettk:
Latest commit was a minor refactor so I think its still safe to merge post PR review. Just as a sanity check, here are all the tests passing after the latest commit:

```r
> devtools::test()
ℹ Testing bbr
✔ | F W S  OK | Context
✔ |        21 | bbr exec functions [0.6s]                                                                                                                         
✔ |        13 | checking if models are up to date [0.6s]                                                                                                          
✔ |         9 | Collapse columns to string representation [0.7s]                                                                                                  
✔ |        73 | Constructing config log from bbi_config.json [2.6s]                                                                                               
✔ |        40 | Copying model objects [0.5s]                                                                                                                      
✔ |         6 | copy-model-helpers [0.3s]                                                                                                                         
✔ |        34 | cov-cor [0.7s]                                                                                                                                    
✔ |        35 | Extract model paths from based_on fields [1.5s]                                                                                                   
✔ |        34 | Test get_omega, get_sigma, and get_theta functions [1.6s]                                                                                         
✔ |        49 | Build paths from model object [0.4s]                                                                                                              
✔ |         7 | model_diff() comparing models [0.4s]                                                                                                              
✔ |        19 | Test bbi summary on multiple models [1.0s]                                                                                                        
✔ |       105 | Test bbi summary functions [1.6s]                                                                                                                 
✔ |        73 | Modify attributes of model object [0.4s]                                                                                                          
✔ |        25 | Testing function to create or read in model object [0.1s]                                                                                         
✔ |        28 | nm-file [0.2s]                                                                                                                                    
✔ |        29 | nm-join [1.5s]                                                                                                                                    
✔ |        20 | nm-tables [0.2s]                                                                                                                                  
✔ |        30 | Test bbi batch parameter estimate functions [7.0s]                                                                                                
✔ |        11 | Test param_estimates functions [0.7s]                                                                                                             
✔ |        43 | test parsing labels for parameter table [1.9s]                                                                                                    
✔ |        33 | testing print methods for bbi objects [5.1s]                                                                                                      
✔ |         5 | read_bbi_path() helper function                                                                                                                   
✔ |        50 | Reading NONMEM output files into R [0.9s]                                                                                                         
✔ |        41 | Constructing run log from model yaml [1.5s]                                                                                                       
✔ |        10 | submit_model(.dry_run=T)                                                                                                                          
✔ |        14 | submit_models(.dry_run=T) [0.4s]                                                                                                                  
✔ |        75 | Test creating summary logs [16.4s]                                                                                                                
✔ |         9 | Comparing tags between models [0.4s]                                                                                                              
✔ |        31 | test_threads(.dry_run=T) [4.9s]                                                                                                                   
✔ |         7 | test-use-bbi [4.3s]                                                                                                                               
✔ |        29 | Utility functions for building args, etc. [0.1s]                                                                                                  
✔ |        39 | testing a composable workflow and running bbi [95.8s]                                                                                             

══ Results ═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Duration: 154.6 s

[ FAIL 0 | WARN 0 | SKIP 0 | PASS 1047 ]
```

R/model-summary.R
barrettk: > > Ill nest that function inside combine_list_objects then
> 
> Why `combine_list_objects` (which is a general utility) rather than `create_summary_object`?

That was a typo my bad lol. Edited it but it likely showed the first version in the github email. Made the change to the correct function (`create_summary_object`)

R/utils.R
kyleam: > Im going to keep the functions the way they are for now,

Sure, Seth already approved it, so I think he agrees these are okay.  I'm just giving my opinion.

> as I think the logic is easier to follow in the first function.

I think using `map` on a vector is convoluted and you should just use more standard operations.  And the condition branching is unnecessary.  So, in my eyes, definitely not easier to follow.

I'd suggest at least move the conditions outside the map, because that's not changing within the loop.

barrettk: I had a feeling I should have waited to merge. Was in the mindset of 'checking off boxes' and it seemed good to go after the latest commit. I can make a quick new PR for that change if you'd like

kyleam: > I had a feeling I should have waited to merge

You merging is fine from my perspective.  I'm just trying to clarify my input/opinion, but the main reason I jumped in was to ask for [relocating the map_list_recursive](https://github.com/metrumresearchgroup/bbr/pull/524#discussion_r936137729) for compatibility with gh-527, and you addressed that (thanks).

PR #525, opened 2022-07-25
docs: regenerate with latest roxygen2

kyleam:
This repo's pkgr.yml pulls from the latest CRAN, so the generated docs
should generally be built with the latest roxygen2 available.
roxygen2 v7.2.1 came out 2022-07-18 and produces a lot off Rd changes,
most of which are probably related the describedIn changes [*].

Regenerate the documentation in a dedicated PR so that we can quickly
merge these changes and allow in-flight PRs to build on top of
them (avoiding conflicts and duplicate commits for the same Rd
changes).

[*] https://github.com/r-lib/roxygen2/blob/main/NEWS.md#roxygen2-721

---

Spotted these when working on another series and am trying to get ahead of conflicts and unnecessary noise.  Requesting a review from @barrettk, though @seth127 or @rymarinelli please feel free to jump and review instead.  After this lands, will ping internally saying to update roxygen2.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -79,5 +79,5 @@ VignetteBuilder:
 Encoding: UTF-8
 LazyData: true
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.2.0
+RoxygenNote: 7.2.1
 SystemRequirements: bbi (>= 3.0.2)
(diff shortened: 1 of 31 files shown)


barrettk (approved):
Ran all the tests and went through some new docs. LGTM

PR #527, opened 2022-07-27
model_summaries: use concurrency of `bbi nonmem summary` if available

### LINKED ISSUE #395: Refactor model_summaries() to use bbi/Go parallelism
When `model_summaries()` was implemented (#53 and #117) we decided just to repeatedly call `bbi nonmem summary` under the hood for each model. This was relatively fast and much easier to implement.

However, `bbi` has the capability to [accept multiple models](https://github.com/metrumresearchgroup/bbi/blob/develop/cmd/summary.go#L74-L160) to the `bbi nonmem summary` call, which would then use Go routines to run the summaries in parallel. This has been in `bbi` [since June 2020](https://github.com/metrumresearchgroup/bbi/pull/171).

The speed of `model_summaries()` is not an issue when only summarizing a couple models, but it has been noted several times that it is prohibitive for large numbers of models (e.g. a bootstrap). This was a big part of the motivation for the batch parameter estimates function (#147). However, there are times when the user might want other information (for example heuristics) from a large number of models.

kyleam:
This series addresses gh-395 and builds on top of changes on the bbi side (<https://github.com/metrumresearchgroup/bbi/pull/272>) that are available under the `v3.2.0-alpha.3` dev release.

Commit overview:

```
* check_bbi_version_constraint: improve an error message
* check_bbi_version_constraint: rename a variable
* check_bbi_version_constraint: return bbi exe path
* check_bbi_version_constraint: rename to assert_bbi_version
* add test_* variant of assert_bbi_version
* skip_if_old_bbi: reduce number of spaces used for indentation
* skip_if_old_bbi: take minimum requirement not last incompatible
* check_run_times: use test_bbi_version
* cov_cor.bbi_nonmem_model: drop bbi version check

  This first set reworks check_bbi_version_constraint(), renaming it
  to assert_bbi_version() and adding a new test_bbi_version() helper.
  test_bbi_version() is needed by the new model_summaries()
  implementation.

  I can split this off into its own series if desired.

* test-refs/1_summary_obj.R: run through RStudio's "Reformat Code"
* test-refs/1_summary_obj.R: regenerate with development bbi
* setup-workflow-ref.R: add "success" to list of names

  These are adjusting the tests for the new output format of
  `bbi nonmem summary --json`.

* bbi_exec: let caller bypass non-zero exit error signaling
* model_summaries.list: drop unnecessary re-assignment
* model_summaries.list: extract model_summary loop to its own function
* test-summary-log.R: set bbi_exe_path via file-wide
* model_summaries.list: pass multiple models to bbi if supported
* ci: test with bbi v3.2.0-alpha.3

  The fifth commit is the main commit of interest.  It reimplements
  model_summaries() using `bbi nonmem summary --json MOD1 MOD2 ...`.
```

Closes #395.


kyleam:
Marked as draft until I figure out the CI failures (`devtools::check()` passed for me on metworx with alpha.3).

seth127 (changes requested):
Some proposed changes in #529, notably 7f51a88072dd92d9b5742e66b7ac97d6b00cb26b which I'm pretty sure is the right move.

seth127 (approved):
I just merged in my [`remove_check_lst_file`](https://github.com/metrumresearchgroup/bbr/pull/529) branch and then did some ad-hoc testing with the [expo repo](https://github.com/metrumresearchgroup/expo1-nonmem-foce/). Testing (and results, in comments) are below. I think this is good to go though.

<details>
<summary>ad-hoc testing:</summary>
(You'll need to run `script/boot-collect.R` to generate bootstrap output in order to actually run this code)

```r
library(bbr)
library(dplyr)
boot_dir <- here::here("model", "pk", "boot")

options("bbr.bbi_exe_path" = "/data/apps/bbi")
bbr::use_bbi(.version = "v3.2.0-alpha.4")

# break some stuff to test error handling
fs::file_move(
  file.path(boot_dir, "003/003.lst"),
  file.path(boot_dir, "003/003.lost")
)
fs::file_move(
  file.path(boot_dir, "005/005.lst"),
  file.path(boot_dir, "005/005.lost")
)
writeLines("this is gonna break", file.path(boot_dir, "005/005.lst"))
fs::file_move(
  file.path(boot_dir, "007/007.ext"),
  file.path(boot_dir, "007/007.eoxt")
)

### using the installed version of `bbr`, no parallelization, 
### takes ~30 seconds to run 1000 model summary calls
bbi_version()
# [1] "3.2.0-alpha.4"
packageVersion("bbr")
# [1] ‘1.3.1’
system.time(smry1 <- summary_log(boot_dir, .fail_flags = list(no_ext_file = T)))
# user  system elapsed                                                                                                                  
# 17.866  12.766  28.829 
err1 <- filter(smry1, !is.na(error_msg))
print(nrow(err1))
# [1] 1
nrow(filter(smry1, needed_fail_flags))
# [1] 2

### loading `bbr` from the repo, this branch,
### takes ~5 seconds to run 1000 model summary calls (with 8 cores)
devtools::load_all("~/metrumresearchgroup/bbr")
bbi_version()
# [1] "3.2.0-alpha.4"
packageVersion("bbr")
# [1] ‘1.3.1.9001’

system.time(smry2 <- summary_log(boot_dir, .fail_flags = list(no_ext_file = T)))
# user  system elapsed                                                                                                                                                                 
# 5.784   0.452   4.878 
err2 <- filter(smry2, !is.na(error_msg))
print(nrow(err2))
# [1] 1
nrow(filter(smry2, needed_fail_flags))
# [1] 3
### slightly weird result here, see below...

# unbreak stuff
fs::file_move(
  file.path(boot_dir, "003/003.lost"),
  file.path(boot_dir, "003/003.lst")
)
fs::file_move(
  file.path(boot_dir, "005/005.lost"),
  file.path(boot_dir, "005/005.lst")
)
fs::file_move(
  file.path(boot_dir, "007/007.eoxt"),
  file.path(boot_dir, "007/007.ext")
)
```
</details>

<details>
<summary>slightly weird result, mentioned above...</summary>
I don't think this necessarily needs to be fixed, but it is a little weird, so I though I might as well document it.

With the changes in this branch, and `bbi v3.2.0-alpha.4`, I noticed two unexpected things:

1. In `summary_log()` the `needed_fail_flags` column gets set to `TRUE` for all three models that I "break". I can follow the code path and see why this happens and I guess it's fine. But in my thinking (and the previous behavior) I'm pretty sure this was only `TRUE` when the model initially errored, but then _didn't_ error with the `.fail_flags`.

2. This is even weirder: the `005` model, where I overwrote the `.lst` with only the text `"this will break"`, actually parses when `no_ext_file = T` is passed. I noticed this in `summary_log()`, but I can reproduce it with a single `model_summary()` call too. Not a big deal, but this should... probably fail.

```r
# using the bbr test model
devtools::load_all()
setwd("tests/testthat")
source("setup-workflow-ref.R")

# break it
fs::file_move(
  build_path_from_model(MOD1, ".lst"),
  build_path_from_model(MOD1, ".lost")
)
writeLines("this is gonna break", build_path_from_model(MOD1, ".lst"))

model_summary(MOD1)
# Error: model_summary(1) failed with the following error. This may be because the modeling run has not finished successfully.
# 
# ERROR: 
#   lengths of parameter data (1) and estimation methods (0) differ. Incomplete run? 

model_summary(MOD1, .bbi_args = list(no_ext_file = T))
# Dataset: -999999999                                                                                                                                      
# 
# Records: -999999999	 Observations: -999999999	 Subjects: -999999999
# 
# Objective Function Value (final est. method): NA
# 
# Estimation Method(s):
#   
#   –  
# 
# No Heuristic Problems Detected

# unbreak it
fs::file_move(
  build_path_from_model(MOD1, ".lost"),
  build_path_from_model(MOD1, ".lst")
)
```

kyleam:
> script/boot-collect.R

Thanks for testing this with a larger case.

> [weird result 1]
> I can follow the code path and see why this happens and I guess it's fine. But in my thinking (and the previous behavior) I'm pretty sure this was only TRUE when the model initially errored, but then didn't error with the .fail_flags.

Thanks.  I missed that detail in my conversion.  I think it'd be good for both code paths to be consistent (even if it's unlikely to matter in practice).

> [weird result 2]
> the 005 model, where I overwrote the .lst with only the text "this will break", actually parses when no_ext_file = T is passed. I noticed this in summary_log(), but I can reproduce it with a single model_summary() call too. Not a big deal, but this should... probably fail.

Hmm, yeah.  I'll take a closer look.

kyleam:
> [weird result 1]

This one should be resolved with 06562409 (model_summaries_concurrent: fix .fail_flags discrepancy, 2022-08-02).

While working on that, I realized that gh-529 is too aggressive because it leads to some JSON parsing failures with older versions of bbi.  I've added back `check_lst_file` (220a931b), calling it only for bbi versions before 3.2.0.  Let me know if you have any objections to that.

remaining todos:

- [ ] look into weird result 2
- [ ] wait for gh-524 to land because that will have a text conflict in `inst/test-refs/1_summary_obj.R` and (at least at the moment) conceptual conflicts

kyleam:
@seth127: there have been some substantial changes since you last took a look (c36155b6d), especially reintroducing `check_lst_file`, so I've re-requested your review.

```
$ git log --oneline --first-parent --no-decorate c36155b6d..
8a44798e Merge branch 'main' into summary-multi
ce785a40 model_summaries_concurrent: replace a subsetted map with modify_at
06562409 model_summaries_concurrent: fix .fail_flags discrepancy
bc74814d model_summaries: handle NULL .fail_flags
220a931b model-summary: restore check_lst_file() check for older bbi versions
2073187c test-refs/1_summary_obj.R: regenerate with build-model-summary-refs.R
```

seth127 (approved):
This all works for me. I just tried it on all the same manual tests I did last time and it's working as advertised. 

It's a bit unfortunate that we need that convoluted code in there to support the pre-3.2.0 behavior, but oh well. I think this is the right approach.

kyleam:
@seth127 Thanks for taking another look.

> It's a bit unfortunate that we need that convoluted code in there to support the pre-3.2.0 behavior, but oh well

Agreed.  I looked forward to whenever we bump the min bbi version and can strip out that and similar compatibility kludges.

PR #528, opened 2022-07-28
updated pkgr to latest MPN snapshot (for mrgvalprep and mrgvalidate)

barrettk:
Update pkgr to have latest MPN snapshot. mrgvalprep had a bug in it that has since been addressed

### DIFF
--- a/pkgr.yml
+++ b/pkgr.yml
@@ -9,7 +9,7 @@ Packages:
 
 Repos:
   - CRAN: https://cran.rstudio.com
-  - MPN: https://mpn.metworx.com/snapshots/stable/2022-02-11 # used for mrgval
+  - MPN: https://mpn.metworx.com/snapshots/stable/2022-06-15 # used for mrgval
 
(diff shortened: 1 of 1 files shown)


inst/validation/bbr-requirements.yaml
```diff
@@ -619,7 +619,7 @@ NWMD-R010:
   tests:
   - BBR-NWMD-010
 NWMD-R011:
-    description: new_model() ignores files' extensions
+  description: new_model() ignores files' extensions
```
kyleam, line 622: Could you drop this change?  Let's keep this PR restricted to the simple MPN bump.

barrettk: Is there an easy way to remove that commit?

kyleam: pinging you for call

kyleam (approved):
Thanks, good to go once drone completes.

PR #529, opened 2022-07-29
Remove `check_lst_file()`

seth127:
This branch began as a proposed one-liner fix to the changes made in #527, but it expanded a bit in scope as I poked around in the code.

### Original purpose

That proposed fix is in the first commit (7f51a88072dd92d9b5742e66b7ac97d6b00cb26b) and I still think we should, at the very least, bring that in as part of #527. More detail is in that commit message.

### Refactoring to remove `check_lst_file()`

The rest of the commits are related to removing `check_lst_file()` altogether. The basic principle here is that I think we should move towards passing the user request through to `bbi` and passing back the (ideally reasonable) error that `bbi` would possibly return. This is a bit wrapped up in [another comment I made on the `bbi` repo](https://github.com/metrumresearchgroup/bbi/pull/272#issuecomment-1198836335). I'm not sure we _need_ to address those together, but I would kind of prefer it.

A bit more on the `check_lst_file()` function and why I think it's time has come to bid adieau:

* It is [quite old](https://github.com/metrumresearchgroup/bbr/commit/9d3aeb158548d6c063aee4dac8a5bae8dd987874) and uses some unfortunate patterns (discussed below).
* The [original comment](https://github.com/metrumresearchgroup/bbr/blame/main/R/model-summary.R#L193-L196) above its (only) use in the code says `"the output always should be runname.lst"`, which is true (since `bbi` will be naming the `.lst` file) and also proves the point that the [glob syntax](https://github.com/metrumresearchgroup/bbr/blame/main/R/model-summary.R#L243) was never needed in the first place.
* In fact, the aforementioned glob syntax actually created [some issues](https://github.com/metrumresearchgroup/bbr/commit/2ea83b1c4d0d334620a7ea289253e2c0351ed69d) that we had to deal with, and with the changes here we can remove a test that was necessary from a code-path point of view, but testing a situation that would probably never happen in the wild anyway. 
  * Although... that [commit message](https://github.com/metrumresearchgroup/bbr/commit/2ea83b1c4d0d334620a7ea289253e2c0351ed69d) _does_ say this was discovered "in the wild", which I'm confused by. Either way, the new behavior would be to always only look for the `.lst` file at `build_path_from_model(.mod, ".lst")` (i.e. `<run>/<run>.lst`), which I think is always the correct thing to do, even in the weird case where there are multiple `.lst` files in a single output dir.
* As a final note, the [commit](https://github.com/metrumresearchgroup/bbr/commit/9d3aeb158548d6c063aee4dac8a5bae8dd987874) that `check_lst_file()` originally came from is called `wip: dump for seth` which is... funny. I mean, maybe it's late in the evening and that's a bit crude. But it is a _little_ funny, and probably (jokes aside) a good indication that its a relic from a previous era of this codebase.


kyleam:
> that [commit message](https://github.com/metrumresearchgroup/bbr/commit/2ea83b1c4d0d334620a7ea289253e2c0351ed69d) does say this was discovered "in the wild", which I'm confused by.

Skimming over this for the time being (since changes on the bbi side need to land first), but, just for reference, I'll send you the internal thread offline.

seth127:
thanks for the context. For posterity: it was basically that someone had manually put a second `.lst` file into their output dir, which caused some weird behavior.

I think I still stand by my basic point that:
* bbi names that file so, assuming the model was run by bbr/bbi, we know what it'll be
* if someone puts in a second .lst, I think ignoring it is probably the best move
* if someone manually renames it... I'm fine with dying and telling them "we can't find the lst file at the path we expect"
* even if it wasn't run with bbr/bbi:
  * `<run>/<run>.lst` seems to be a pretty strong naming convention anyway
  * we don't make any promises that `model_summary()` can parse a non-bbr run

kyleam (approved):
@seth127 Thanks.  The motivation you laid out all makes sense to me, as do the code changes.

> Note that BBR-SUM-004 still fails because of bbi panic that is no longer swallowed by model_summary(). I would prefer to fix that in bbi, rather than make this test expect to see the ugly output that currently comes back (a JSON parsing error).

Resolved with the latest push to that bbi pr (7c0caa7).  I'm holding off on tagging `alpha.4` for bbi to avoid proliferating tags if more changes are needed to that bbi pr.  I'm okay with merging this to the other pr with the test failure, and then cutting `alpha.4` and confirming a green build before the merge of gh-527.  But let me know if you want me to tag `alpha.4` for this pr (or of course feel free to do so yourself).

<details>
<summary>before and after test runs</summary>

On e1e1bda (`alpha.3`):

```
ip-10-128-0-236:bbr$ Rscript -e 'devtools::test(filter = "model-summary")'
ℹ Testing bbr
setting `bbr.DEV_no_min_version` to TRUE so no validation of bbi version will occur
✔ | F W S  OK | Context
✖ | 1     107 | Test bbi summary functions [1.2s]
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Failure (test-model-summary.R:150:5): model_summary() fails predictably if it can't find some parts (i.e. model isn't finished) [BBR-SUM-004]
`model_summary(mod2)` threw an error with unexpected message.
Expected match: "model_summary.*modeling run has not finished"
Actual message: "lexical error: invalid char in json text.\n                                       panic: runtime error: index out\n                     (right here) ------^\n"
Backtrace:
  1. testthat::expect_error(model_summary(mod2), regexp = NOT_FINISHED_ERR_MSG)
       at test-model-summary.R:150:4
 10. jsonlite::fromJSON(., simplifyDataFrame = FALSE)
       at magrittr/R/pipe.R:136:2
 11. jsonlite:::parse_and_simplify(...)
       at jsonlite/R/fromJSON.R:101:2
 12. jsonlite:::parseJSON(txt, bigint_as_char)
       at jsonlite/R/fromJSON.R:117:2
 13. jsonlite:::parse_string(txt, bigint_as_char)
       at jsonlite/R/parseJSON.R:5:4
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

══ Results ═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Duration: 1.2 s

[ FAIL 1 | WARN 0 | SKIP 0 | PASS 107 ]
```

On 7c0caa7 (current tip of bbi's `summary-json-errors` branch):

```
ip-10-128-0-236:bbr$ Rscript -e 'devtools::test(filter = "model-summary")'
ℹ Testing bbr
setting `bbr.DEV_no_min_version` to TRUE so no validation of bbi version will occur
✔ | F W S  OK | Context
✔ |       108 | Test bbi summary functions [1.1s]

══ Results ═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Duration: 1.1 s

[ FAIL 0 | WARN 0 | SKIP 0 | PASS 108 ]
```

</details>

seth127:
I got a clean bill of health with `alpha.4` (mentioned [here](https://github.com/metrumresearchgroup/bbi/pull/272#issuecomment-1201917624)) so I'm gonna go ahead and merge this.

PR #530, opened 2022-07-29
Patch For Repair Names

### LINKED ISSUE #456: nm_join: repair names when coercing to tibble
# Summary
As a user, I want `nm_join()` to repair duplicate names when coercing to `tibble()`.

rymarinelli:
closes #456 

```
NMF-R008:
  description: nm_file() repairs duplicate column names that are present in input file
  tests: ...
```


rymarinelli:
I added a test case based on Kyle's error. It would be useful to get feedback if this test is too complex and should be simplified or if it adds confidence.

R/nm-file.R
```diff
@@ -96,7 +96,9 @@ nm_data <- function(.mod) {
   check_model_object(.mod, c(NM_MOD_CLASS, NM_SUM_CLASS))
   .path <- get_data_path(.mod)
   verbose_msg(glue("Reading data file: {basename(.path)}"))
-  .d <- as_tibble(fread(.path, na.strings = ".", verbose = FALSE))
+  .d <- fread(.path, na.strings = ".", verbose = FALSE)
+  colnames(.d) <- make.unique(.d %>% colnames(), sep="_")
+  .d <- as_tibble(.d)
```
barrettk, line 101: I recall talking about this. I think we said we might as well put something in this function as well just in case? If so im wondering if we want the same error handling as we have in `nm_file_impl`, in which case we'd want another (similar) test

barrettk (changes requested):
Mainly one concern (`nm_data()` error handling). If we're going to have the make.unique() parts in that function, I think we'd want it to be similar to `nm_file_impl`, in that an identical warning message is returned. To avoid duplicate code it might make sense to pull out the code below, and turn it into a helper function thats references in both functions:

```r
    if(any(duplicated(names(data)))){
      dup_cols <- names(data)[duplicated(names(data))]
      dup_cols_str <- paste(dup_cols, collapse = ", ")
      dup_id <- c(which(duplicated(names(data))) -1, which(duplicated(names(data)))) %>% sort()
      warning(glue("{basename(.path)} had the following duplicated columns: {dup_cols_str}\n  Duplicate names will be repaired with `make.unique()`")) # something like this, and then say if they're equal
      colnames(data) <- make.unique(colnames(data) )
    }
```

maybe something like (with better function name):
```r
address_duplicates <- function(data){
    if(any(duplicated(names(data)))){
      dup_cols <- names(data)[duplicated(names(data))]
      dup_cols_str <- paste(dup_cols, collapse = ", ")
      dup_id <- c(which(duplicated(names(data))) -1, which(duplicated(names(data)))) %>% sort()
      warning(glue("{basename(.path)} had the following duplicated columns: {dup_cols_str}\n  Duplicate names will be repaired with `make.unique()`")) # something like this, and then say if they're equal
      colnames(data) <- make.unique(colnames(data) )
    }
return(data)
}
```

Then you could just add `data <- address_duplicates(data)` to `nm_file_impl` and `nm_data`

R/utils.R
```diff
@@ -650,7 +650,7 @@ remove_dup_cols <- function(data){
     dup_cols <- names(data)[duplicated(names(data))]
     dup_cols_str <- paste(dup_cols, collapse = ", ")
     dup_id <- c(which(duplicated(names(data))) -1, which(duplicated(names(data)))) %>% sort()
-    warning(glue("{basename(.path)} had the following duplicated columns: {dup_cols_str}\n  Duplicate names will be repaired with `make.unique()`")) # something like this, and then say if they're equal
+    warning(glue("Following duplicated columns: {dup_cols_str}\n  Duplicate names will be repaired with `make.unique()`")) # something like this, and then say if they're equal
     colnames(data) <- make.unique(colnames(data) )
```
barrettk, line 654: @rymarinelli can you remove this comment before merging?

barrettk (approved):
Looks good! Thanks

R/utils.R
barrettk: In case its not clear, just this comment: `# something like this, and then say if they're equal`

rymarinelli: I added an extra comment at the end. Is it what you had in mind?

barrettk: I just meant remove the previous one (see above), but the comment you added is fine

PR #534, opened 2022-08-02
test-bbr: don't assume bbi_binary is in bbi.yaml

kyleam:
A bbi_init() test checks that bbi_binary is in the generated yaml, but
that's no longer the case as of 807cbf8 (sge: use process bbi
executable in scripts unless overridden, 2022-07-30).

The particular key isn't important to this test, so check a different
key.


seth127 (approved):
good stuff

PR #539, opened 2022-08-08
Shorten output on .join_col warning

### LINKED ISSUE #536: Printing of duplicate rows in nm_join is excessive
After merging https://github.com/metrumresearchgroup/bbr/pull/533, @kylebaron pointed out that printing the duplicate rows could be a bit excessive in the case of 1000's of duplicates:
image

Additionally, it would make sense to update the docs and inform users of how duplicate rows are actually handled:
image

Small adjustment, but while we're at it we might as well make this code more legible:
image

rymarinelli:
Closes #536


R/nm-join.R
```diff
@@ -109,10 +115,10 @@ nm_join <- function(
     stop(glue("couldn't find `.join_col` {.join_col} in data with cols: {paste(names(.d), collapse = ', ')}"))
   }
 
-  if(.d[.join_col] %>% anyDuplicated() != 0)
+  if(anyDuplicated(.d[.join_col]) != 0)
   {
    dup_row <- .d[.join_col][duplicated( .d[.join_col]) %>% which(),]
-   stop(glue("Duplicate rows in data in {.join_col}: {dup_row}"))
+   stop(glue("Duplicate rows in data in {.join_col}: {dup_row %>% head()}"))
   }
```
barrettk, line 122: If we `head()` the duplicate rows, we might want to write `...truncated` if there are more duplicate values beyond the default of 6. My vote would be to remove the printing of the duplicate rows entirely, and just say:
```
stop(glue("Duplicate rows were found in {.join_col}. Please see `?nm_join` for more details"))
```

barrettk (commented):
Seems like a step in the right direction. @kylebaron do you have any thoughts about the documentation or using `head()` for the duplicate rows?

barrettk (approved):
tests pass on my end after merging in the changes/updating the bbi version. Would like to get someone's opinion on the new details for the `nm_join()`, but otherwise looks good to me

PR #541, opened 2022-08-10
test-print.R: remove stray print call

kyleam:
Clean up a recently introduced print call that adds noise and breaks
what should be a single line of output:

    ⠦ |         7 | testing print methods for bbi objects
    [1] "  …/bin/bbi nonmem run sge ... [truncated] --overwrite --parallel --threads=4"
    ✔ |        33 | testing print methods for bbi objects [3.9s]


barrettk (approved):
simple fix, looks good to me

PR #545, opened 2022-08-16
ci: test against minimum required bbi version

kyleam:
The first commit of this PR update the `mpn:oldest` build to use bbi version 3.0.2, the value of `bbr.bbi_min_version`.  The goal is to keep bbr compatible (and the tests passing) with the minimum version that's specified there.

The other two commits adds some skipping/conditional logic to the tests to avoid some failures when running with the minimum required bbi.

I talked with @seth127, and he is on board with the general idea, but of course please chime in with any comments on the specifics.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -69,7 +69,7 @@ trigger:
 ---
 kind: pipeline
 type: docker
-name: mpn:oldest
+name: oldest
 
 platform:
   os: linux
@@ -90,7 +90,7 @@ steps:
   image: 906087756158.dkr.ecr.us-east-1.amazonaws.com/r-dev-ci-mpn-4.0:2020-06-08
   commands:
   - chmod +x install_bbi.sh
-  - ./install_bbi.sh -p /ephemeral/bbi -v 'v3.2.0-rc.1'
+  - ./install_bbi.sh -p /ephemeral/bbi -v 'v3.0.2'
   volumes:
   - name: cache
     path: /ephemeral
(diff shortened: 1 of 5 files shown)


barrettk:
@kyleam do we want a requirement/story for this, or do you think its unnecessary?

seth127:
> do we want a requirement/story for this

No, this is more just "good housekeeping"

kyleam:
> @kyleam do we want a requirement/story for this, or do you think its unnecessary?

Hmm, I was considering all of these changes developer-facing meta stuff related to the CI that don't need to have requirements/stories.  But...

> Not a whole lot changed, but I feel like the compatibility with older versions of bbi might be something to keep track of (perhaps the release notes are sufficient though)

... I see the point that compatibility with older versions fits a user story 

But, in terms of _this_ PR, I think my view is that `bbr.bbi_min_version` isn't changing, and we're just cleaning up some tests that were a bit too forward-thinking in their expectations.  So, while we may want a story at some point, this PR doesn't need to be responsible for taking that on.

Does that make sense?  (Of course, if someone has a story right at the tip of their tongue and has a strong opinion that we need it, it's very little work to include that here and would be fine by me.)

> (though I'm off the top of my head, I'm not sure what exact wording we'd want)

Us adding a build to the CI that tests with the minimum version and cleaning up some tests doesn't feel NEWS-worthy to me.  However, if we _were_ changing `bbr.bbi_min_version`, that's definitely something that should be highlighted.

seth127:
This looks good to me. Thanks y'all.

barrettk (approved):
Looks good to me (tests ran)

PR #358, opened 2021-03-09
DO NOT MERGE: bbr Stan modeling ALPHA release

seth127:
# bbr_stan_alpha
The intention of this branch is to begin implementing support for Stan modeling with `bbr`. While this code should all be tested and working, **the function names and API are considered experimental and may change with new releases.** The plan is to merge this functionality into `bbr` proper once this has stabilized.

All releases tagged `1.1.0.9xxx` refer to "alpha" releases cut from the `bbr_stan_alpha` branch.

Issue #492, opened 2022-05-31
Consider changing all _log() functions to default to .recurse = FALSE


seth127:
Functions `run_log()`, `summary_log()`, and `config_log()` all default to searching for models recursively in sub directories. However, user feedback tells us:

* Typically users organize their models by directory and intend to create separate run logs for each group of models (i.e. each directory)
* This causes problems when something like a bootstrap is in a subdirectory (common practice) because then the log attempts to read in all of the bootstrap models as well, making the log huge and unreadable

We should consider changing the this default so that users would have to specify `.recurse = TRUE` to search sub directories.

barrettk:
Stories:
```
LOG-S002 (this already exists):
  name: Working directory and paths in run log tibble are correct
  description: As a user, I want to be able to use the paths included in the run_log()
    output to unambiguously find model input and output files. A user should be able
    to do this, even when the models were run on a different machine or in a different
    directory than the one in which they are currently being read (by run_log()).
  ProductRisk: Low
  Requirements:
     - (requirements of previous tests)
     - LOG-R011
     - LOG-R012
     - LOG-R013
```

_New_ Requirements:
```
LOG-R011:
  description: run_log() can optionally look in subdirectories via the `.recurse` argument
  tests: 
LOG-R012:
  description: summary_log() can optionally look in subdirectories via the `.recurse` argument
  tests: 
LOG-R013:
  description: config_log() can optionally look in subdirectories via the `.recurse` argument
  tests: 
```

seth127:
I looks like we're gonna move ahead with a [`1.5.0` release](https://github.com/metrumresearchgroup/bbr/milestone/27) (as opposed to `1.4.1`, see here) and I'm considering if we should tack this on as well. It's just another little thing that has been sitting around but never seems important enough to do a release for. 

Any thoughts @kyleam? Any reason you can think of that we wouldn't want to include this in here?

kyleam:
> Any reason you can think of that we wouldn't want to include this in here?

In terms of including this in a minor release, I suppose the main thing is that the default flip is a backward incompatible change.  But these functions are all very interactive in nature, so I'm not sure there's really much breakage to be concerned about in practice.  The worst scenario that I can think of is that a user doesn't realize that some models are being dropped because we're no longer recursing and assumes they are working the complete set.

Do you think there's anything worth worrying about there?

seth127:
[...]

PR #547, opened 2022-08-17
Add .overwrite argument to submit_model and submit_model

### LINKED ISSUE #544: Add submit_model(..., .overwrite) argument
The proposed change is to add an `.overwrite` argument to `submit_model()` (similar to how `new_model()` and `copy_model_from()` have one) so that the user doesn't need to pass it through `.bbi_args`.

When we make this change, we should also update [this section](https://metrumresearchgroup.github.io/bbr/articles/getting-started.html#passing-arguments-to-bbi) of the "Getting Started" vignette.

kyleam:
This extracts two commits from @seth127's [bbr_alpha_stan branch](https://github.com/metrumresearchgroup/bbr/pull/358) that address #544.

At this point, `bbr_alpha_stan` serves as the base for the bbr.bayes split (#543), so I'd prefer we not actually remove the corresponding commits (902a13d8, 1c44d1c3) from that branch.  However, it would be fine to merge this PR rather than wait on #543 to be merged.  (That'd just lead to a minor conflict and two sets of commits making these changes.)

---

I'm marking this as a draft because the plan is to merge this after the upcoming release.

Leftover bits:
- [x] add test ID and adjust stories/reqs
- [x] any other spots in the documentation that should switch to using this argument rather than `.bbi_args`?
   Edit: There are a few instances in `vignettes/nonmem-parallel.Rmd`, but I think those are okay to leave as is (given we have no plan to deprecate passing overwrite via `.bbi_args`).  I'm not spotting anything else, at least with simple grepping.

Closes #544.

### DIFF
--- a/R/submit-model.R
+++ b/R/submit-model.R
@@ -16,6 +16,10 @@
 #'   `"local"` for local execution. This can be passed directly to this argument
 #'   or set globally with `options("bbr.bbi_exe_mode")`.
 #' @param ... args passed through to `bbi_exec()`
+#' @param .overwrite Logical to specify whether or not to overwrite existing
+#'   model output from a previous run. If `NULL`, the default, will defer to
+#'   setting in `.bbi_args` or `bbi.yaml`. If _not_ `NULL` will override any
+#'   settings in `.bbi_args` or `bbi.yaml`.
 #' @param .config_path Path to a bbi configuration file. If `NULL`, the
 #'   default, will attempt to use a `bbi.yaml` in the same directory as the
 #'   model.
@@ -31,6 +35,7 @@ submit_model <- function(
   .bbi_args = NULL,
   .mode = getOption("bbr.bbi_exe_mode"),
   ...,
+  .overwrite = NULL,
   .config_path = NULL,
   .wait = TRUE,
   .dry_run=FALSE
@@ -45,6 +50,7 @@ submit_model.bbi_nonmem_model <- function(
   .bbi_args = NULL,
   .mode = getOption("bbr.bbi_exe_mode"),
   ...,
+  .overwrite = NULL,
   .config_path = NULL,
   .wait = TRUE,
   .dry_run=FALSE
@@ -54,6 +60,7 @@ submit_model.bbi_nonmem_model <- function(
                              .bbi_args = .bbi_args,
                              .mode = .mode,
                              ...,
+                             .overwrite = .overwrite,
                              .config_path = .config_path,
                              .wait = .wait,
                              .dry_run = .dry_run)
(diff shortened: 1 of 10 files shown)


seth127:
Looking at https://github.com/metrumresearchgroup/bbr.bayes/pull/30 and the associated changes in https://github.com/metrumresearchgroup/bbr/tree/nmbayes and #543 ... should we prioritize moving this PR along? 

I don't fully have my head around all the implications, but:

* I think merging this sooner than later will minimize how tangled things get
* I'm not aware of any reason _not_ to merge this before the next `bbr` release (aside from the two"leftover bits" mentioned above)

Any thoughts on that @kyleam ? If we're on the same page, I'd vote to move this to https://github.com/metrumresearchgroup/bbr/milestone/27 (which may end of being `1.5.0`, not sure yet...)

kyleam:
> * I think merging this sooner than later will minimize how tangled things get

It definitely doesn't hurt to get this settled.  Conceptually I would want it to land before bbr-bayes-split so that these changes come in through a dedicated PR rather than being brought in as part of the mega bbr-bayes-split changeset.

> * I'm not aware of any reason _not_ to merge this before the next `bbr` release (aside from the two"leftover bits" mentioned above)

I think the main reason is...

> Any thoughts on that @kyleam ? If we're on the same page, I'd vote to move this to https://github.com/metrumresearchgroup/bbr/milestone/27 (which may end of being `1.5.0`, not sure yet...)

... this last bit.  If we wanted to have a patch-bump bugfix release, then this doesn't belong in it.  If we decide we'll go straight for 1.5.0, then this works [*]

Regardless, I'll work on getting this ready to merge so that we can pull the trigger when we decide it's time.

[*] Of course if we keep on hitting into these sorts of issues, we can have two branches so that we can cut patch releases from the bugfix branch, but as we've discussed before, that's probably a bit too complex for the workflow that's actually been used for bbr.  (Even back when there were multiple long-lived branches, they weren't used in practice to isolate changes this way. )

seth127 (approved):
All looks good and `devtools::check()` passes for me on Metworx. Thanks for cleaning this up.

seth127:
@kyleam are we good to merge this and close https://github.com/metrumresearchgroup/bbr/issues/544 ?

kyleam:
> @kyleam are we good to merge this and close #544 ?

I was going to double check with you tomorrow that you're okay with committing to 1.5.0 for the next release (though I assumed that was the case based on you approving this after I mentioned it and what you said [here](https://github.com/metrumresearchgroup/bbr/issues/492#issuecomment-1324186161)).  So, yes, it's good to merge.

Issue #513, opened 2022-07-07
ci: update after bbi release


kyleam:
With the merge of gh-511, `.drone.yml` installs v3.2.0-alpha.1.  Before the next bbr release, we need to 

 - [x] release a new bbi version
 - [ ] update `.drone.yml` to use to that version

PR #549, opened 2022-08-19
Prepare for 1.4.0 release

kyleam:
* update drone to use latest bbi release (gh-513)
 * release notes
 * pkgdown updates
 * `DESCRIPTION` updates

changeset: https://github.com/metrumresearchgroup/bbr/compare/1.3.1...release/1.4.0


_pkgdown.yml
```diff
@@ -100,3 +102,8 @@ reference:
   - reconcile_yaml
   - check_yaml_in_sync
   - suppressSpecificWarning
+  - test_threads
+  - check_run_times
+  - delete_models
+  - check_nonmem_finished
```
barrettk, line 108: Did you decide you wanted to include `delete_models()`? Im fine with either, just wanted to make sure it was intentional

barrettk (approved):
Looks good. Left one question, but fine with whichever direction you wanted to go

_pkgdown.yml
kyleam: Yeah, I flipped!  In the end, I think, given we export it, we should have it in the index in case someone goes looking for it.  I dunno, but the biggest decision is to export it, and that's been made.  We can always remove it from the pkgdown index later if we want.

PR #535, opened 2022-08-04
Add missed requirements and stories for early 1.4.0 issues

barrettk:
Some of the early PR's were missing requirements/stories, as we weren't in the habit of adding them per PR early on for this milestone. In addition, some tests/requirements were renamed/moved

closes https://github.com/metrumresearchgroup/bbr/issues/502

PR #551, opened 2022-08-24
Match order of stories to previous release

barrettk:
After sending the `bbr 1.4.0` validation docs to Aing, we found that the previous order of stories (i.e. story ID's) must be consistent across previous releases as well.

In this PR:
 - move `MGMT-S010` to `MGMGT-S016`
 - reformat other story ID's to be consistent with the above change^

The validation docs have already been regenerated and forwarded to Aing based on this change

### DIFF
--- a/inst/validation/bbr-stories.yaml
+++ b/inst/validation/bbr-stories.yaml
@@ -184,15 +184,6 @@ MGMT-S009:
   - MMF-R021
   - CTS-R001
 MGMT-S010:
-  name: Add and replace description
-  description: As a user, I want to be able to add, delete, and replace descriptions on a
-    model object. Descriptions should be persisted in the model YAML file.
-  ProductRisk: Low
-  requirements:
-  - MMF-R008
-  - MMF-R009
-  - MMF-R010
-MGMT-S011:
   name: Print method for model object
   description: As a user, I want to be able to print the `bbi_nonmem_model` object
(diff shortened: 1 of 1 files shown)


kyleam:
Ah, so this came from you changing the story ID in eace5331 (additional changes to stories ..., 2022-08-16)?   I should have caught that when reviewing gh-535.  Makes sense, we talked about changing *requirement* IDs (because those were new to 1.4.0), but a story ID shouldn't have been changed.

kyleam (approved):

seth127:
Just for posterity's sake: Aing wasn't concerned with changes to Requirement IDs or Test IDs across releases. She just wanted Story IDs to stay consistent because those are essentially the top-level definition of "what the product does" as far as validation is concerned, so she tracks which stories are added or removed from release to release.

kyleam:
> She just wanted Story IDs to stay consistent

@seth127, that's good to know, but either way, from our perspective, we want all of these IDs to stay stable (stories, reqs, and tests) once they make it into a release, right?  https://github.com/metrumresearchgroup/bbr/pull/535#issuecomment-1219754770

seth127:
> we want all of these IDs to stay stable (stories, reqs, and tests) once they make it into a release, right?

Yes, that's right. I just wanted to note that if we need to change other ones (I'm mainly thinking of test IDs changing if/when we refactor the test suite) that it's not a big deal.

PR #554, opened 2022-09-13
Patch For Threads

rymarinelli:
Closes #553

PR #556, opened 2022-09-22
NULL Elements Patch 

### LINKED ISSUE #555: Handling of NULL values passed in bbi_args
When working on correcting the bug #553, it was observed that NULL could be set as an element in lists, and it would be passed through functions. 

```
  mod1 <- new_model(file.path(tempdir(), "test_path", "1"), .description = "original test-workflow-bbi model",
                        .tags = ORIG_TAGS,.bbi_args = list(parallel = FALSE, threads = NULL))
```

The resulting model object leaves `threads` blank.
```
> mod1

── Status ─────────────────────────────────────────────────────────────────────────────────────────────────────────────

── Finished Running ──

── Absolute Model Path ────────────────────────────────────────────────────────────────────────────────────────────────
• /tmp/Rtmp468kbm/test_path/1

── YAML & Model Files ─────────────────────────────────────────────────────────────────────────────────────────────────
• /tmp/Rtmp468kbm/test_path/1.yaml
• /tmp/Rtmp468kbm/test_path/1.ctl

── Description ────────────────────────────────────────────────────────────────────────────────────────────────────────
• original test-workflow-bbi model

── BBI Args ───────────────────────────────────────────────────────────────────────────────────────────────────────────
• parallel: FALSE
• threads: 
```

The yaml for mod looks like :
```
model_type: nonmem
description: original test-workflow-bbi model
bbi_args:
  parallel: no
  threads: ~
```

rymarinelli:
closes #555

### DIFF
--- a/R/model-summary.R
+++ b/R/model-summary.R
@@ -185,6 +185,9 @@ nonmem_summary <- function(
 ) {
 
   # check against YAML
+  .mod <- sanitize_null_bbi_args(.mod)
+
+  # check against YAML
   check_yaml_in_sync(.mod)
 
   # extract output path
--- a/R/new-model.R
+++ b/R/new-model.R
@@ -72,6 +72,8 @@ new_model <- function(
   if (!is.null(.bbi_args))    .mod <- replace_all_bbi_args(.mod, .bbi_args)
   if (!is.null(.based_on))    .mod <- replace_all_based_on(.mod, .based_on)
   if (isTRUE(.star))          .mod <- add_star(.mod)
+
+  .mod <- sanitize_null_bbi_args(.mod)
   return(.mod)
 }
 
--- a/R/submit-model.R
+++ b/R/submit-model.R
@@ -89,6 +89,8 @@ submit_nonmem_model <- function(.mod,
 
   # build command line args
   .bbi_args <- parse_args_list(.bbi_args, .mod[[YAML_BBI_ARGS]])
+
+  .bbi_args <- sanitize_null_bbi_args(.bbi_args)
   args_vec <- check_bbi_args(.bbi_args)
 
 
--- a/R/utils.R
+++ b/R/utils.R
@@ -709,6 +709,9 @@ sanitize_null_bbi_args <- function(obj)
 
     .args <- map(obj,is.null)
     null_elements <- unlist(lapply(.args, function(x) x[isTRUE(x)])) %>% names()
+
+    if(exists("null_elements") == FALSE || (is.null(null_elements) == TRUE)) return(obj)
+
     null_elements <- as.list(strsplit(null_elements, '\\s+'))
 
     for(i in 1:length(null_elements))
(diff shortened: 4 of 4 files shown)


kyleam:
> Due to `NULL` values being able to elements in a list, some problematic behavior arises.

I'm uncertain about whether it's problematic.  The user sets it to an invalid value, and then when they submit the model, they see 

```
Error in check_bbi_args(.bbi_args) :
There are 1 errors in check_bbi_args(): `` passed for arg `threads`. Expected 'numeric' type.
```

The messaging could be clearer, but that seems fair enough.  So, my concern is that this PR is adding a fairly complicated function to solve a non-problem.  Am I missing something?

Also, is your intent for this PR to supersede gh-554 given it takes the code suggestion from there?

rymarinelli:
> > Due to `NULL` values being able to elements in a list, some problematic behavior arises.
> 
> I'm uncertain about whether it's problematic. The user sets it to an invalid value, and then when they submit the model, they see
> 
> ```
> Error in check_bbi_args(.bbi_args) :
> There are 1 errors in check_bbi_args(): `` passed for arg `threads`. Expected 'numeric' type.
> ```
> 
> The messaging could be clearer, but that seems fair enough. So, my concern is that this PR is adding a fairly complicated function to solve a non-problem. Am I missing something?
> 
> Also, is your intent for this PR to supersede [gh-554](https://github.com/metrumresearchgroup/bbr/pull/554) given it takes the code suggestion from there?

Yes, I think I need to resolve this first before moving back to #554. I agree that the error that raises is minor, but the associated yaml could be problematic with some of the models. For instance, the yaml could reach this state. 
```
model_type: nonmem
description: original test-workflow-bbi model
bbi_args:
  parallel: no
  threads: ~
``` 

@seth127 What are your thoughts on this? Do you think this is worth the complexity? Also, do you have any suggestions on the organization of the two highly related PRs?

kyleam:
> the associated yaml could be problematic with some of the models. For instance, the yaml could reach this state.

Can you flesh out for me a bit more what the problem is?  Is it something beyond "the user attaches invalid argument type to a model, and the submission aborts when it checks the type of those arguments"?

rymarinelli:
> 

I think the primary concern  is with `new_model.` If you passed in `NULL` values into `.bbi_args` it could reach undesired states. 

For instance,  you might have black spaces associated with values.  
```
── BBI Args ───────────────────────────────────────────────────────────────────────────────────────────────────────────
• parallel: FALSE
• threads: 
```

The other concern was with the yaml, and if that could cause any issues with how it is being written out. 

A secondary concern is that some handling might not be working if ``` passed for arg `threads`. Expected 'numeric' type.```  is being reached. It is my understanding the `bbi_args` should not be NULL and should be promptly removed in the logic if ever set to NULL. The goal of writing the function was to try and write some more generalized handling that could be applied to wherever `bbi_args` is touched to ensure that no `NULL` values are passed.

seth127:
I think this PR can be closed without merging, given our conclusion in [this comment](https://github.com/metrumresearchgroup/bbr/issues/555#issuecomment-1258238710)

@kyleam do you agree?

PR #561, opened 2022-11-10
silence tidyselect deprecation warnings

kyleam:
Running the test suite with the most recent tidyselect (1.2.0, released 2022/10) leads to a bunch of warnings that boil down to the same two messages.  The first two commits deal with each case separately.  The third commit is a spot that produces the same warning as the spots handled in the second commit, but the resolution here is different and less straightforward, so I've put it in its own commit.

MPN oldest passes, so we should be good in terms of backward compatibility.


barrettk (approved):
This all looks good. Im surprised they deprecated the use of `.data$`, but im kind of in favor of it (probably a controversial stance). `pmtables` definitely has a lot of these occurrences as well. `mutate_at()` and `vars()` have been superseded for a while, so im glad we're updating those as well.

kyleam:
> Im surprised they deprecated the use of `.data$`, but im kind of in favor of it (probably a controversial stance).

You may be aware of this, but to clarify for anyone else reading this: note that its use isn't entirely deprecated, just in selections: https://www.tidyverse.org/blog/2022/10/tidyselect-1-2-0/

PR #563, opened 2022-11-28
`*_log(.recurse=FALSE)` refactor

### LINKED ISSUE #492: Consider changing all _log() functions to default to .recurse = FALSE

seth127:
Closes #492

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: bbr
 Title: R package for bbi
-Version: 1.4.0
+Version: 1.4.0.8000
 Authors@R: 
     c(person(given = "Seth",
              family = "Green",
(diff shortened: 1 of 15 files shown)


seth127:
Code changes in first commit, tests in second, docs and validation in third. Thanks!

kyleam (approved):
> Wherever possible, I left the default untouched/unspecified. However many tests were using a reference file that used subdirectories, [...]

Makes sense.  Thanks for providing the rationale.

PR #564, opened 2022-11-29
`nm_file()` no longer uppercases columns

### LINKED ISSUE #558: update_run_num and nm_join problematic case studies
Hi all, I just wanted to give some feedback on some snags I ran into when using these functions on a project and suggest some possible solutions you could implement. 

1) `update_run_number` : When I used this on a Bayes run, the .chn and .msf got passed into a `toupper` somewhere in the function, and my control stream was replaced with .CHN and .MSF. this is problematic because I had other functions set up to read in `.chn`, and the uppercase threw that off. I had to manually go into every control stream and change .CHN and .MSF to .chn and .msf. Note: The toupper does NOT effect the .tab, par.tab, etc., so there is something inconsistent about just the .chn and .msf being changed, but not the others. 

Here are some pieces from a sort of working example:

```
> bbi_version() 
[1] "3.2.1"

modTEST <- new_model(file.path(MODEL_DIR, 999999)) 

# here is what is in model 999999:

$EST METHOD=CHAIN FILE=999999.chn NSAMPLE=4 ISAMPLE=0 SEED=10 CTYPE=0 IACCEPT=0.2 DF=10 DFS=0
;$EST METHOD=NUTS SEED=1 NBURN=1000 NITER=2000 PRINT=1 MSFO=./999999.msf RANMETHOD=P PARAFPRINT=10000 BAYES_PHI_STORE=1

;$TABLE ID NUM AUCss BASE KIN KOUT IMAX AUC50 PLAC ETAS(1:LAST) PRED IPRED EPRED EWRES NPDE NOPRINT ONEHEADER FILE=./999999.tab RANMETHOD=P 


copy_model_from(modTEST, "000000", .overwrite = T) %>% update_run_number()

# And here is what it gets changed to via update_run_number:

$EST METHOD=CHAIN FILE=000000.CHN NSAMPLE=4 ISAMPLE=0 SEED=10 CTYPE=0 IACCEPT=0.2 DF=10 DFS=0
;$EST METHOD=NUTS SEED=1 NBURN=1000 NITER=2000 PRINT=1 MSFO=./000000.MSF RANMETHOD=P PARAFPRINT=10000 BAYES_PHI_STORE=1

;$TABLE ID NUM AUCss BASE KIN KOUT IMAX AUC50 PLAC ETAS(1:LAST) PRED IPRED EPRED EWRES NPDE NOPRINT ONEHEADER FILE=./000000.tab RANMETHOD=P 

```

I think the most important thing here is to be consistent -- the chn, msf, tab and par.tab etc. should be treated the same IMO, otherwise makes for problems post-processing when you have to guess which file is uppercase and which is lowercase.

2) `nm_join` : I also had some issues here that I assume were caused by `toupper` hiding somehwere in the function. I ran `nm_join`, grabbed that output, and fed it into my mrgsolve model. The mrgsolve model did not work correctly because one of the parameters got changed to uppercase. I had "AUCss" tabled out from the Nonmem run, then I had "AUCss" as an input parameter to my mrgsolve model. However because `nm_join` turned it into "AUCSS", the parameter wasn't fed into my mrgsolve model causing problems. Maybe I should have used all caps to follow best coding practices and probably will from here on out, but why is the `toupper` implemented in nm_join? Any good reason to keep it there? Was it so that the input data set and the Nonmem tabled out data would all be in the same case and therefore be able to join? If so that totally makes sense to me, but it was a little dangerous in this case here. Food for thought! 

Thanks :)

seth127:
Closes #558

### DIFF
--- a/R/nm-file.R
+++ b/R/nm-file.R
@@ -17,7 +17,6 @@
 #' [PKPDmisc::read_nonmem](https://metrumresearchgroup.github.io/PKPDmisc/reference/read_nonmem.html).
 #'
 #' @return A tibble with the data from the specified file and estimation method.
-#'   All column names will be converted to uppercase.
 #'
 #' @param .mod Either a `bbi_nonmem_model`, `bbi_nonmem_summary`, or a path to a
 #'   file to read in. If passing model object to `nm_file()`, must also pass `.suffix` that
@@ -157,7 +156,6 @@ nm_file_impl <- function(.path) {
   }
 
   # format, message, and return
-  names(.d) <- toupper(names(.d))
   verbose_msg(glue("  rows: {nrow(.d)}"))
   verbose_msg(glue("  cols: {ncol(.d)}"))
   verbose_msg("") # for newline
--- a/R/nm-file.R
+++ b/R/nm-file.R
@@ -13,8 +13,7 @@
 #' `nm_file()` (and family) are _not_ compatible with files that have multiple
 #' tables, for example an `.ext` file for a model with multiple estimation
 #' methods or a table file from a model using `$SIM`. For these kinds of files,
-#' consider using
-#' [PKPDmisc::read_nonmem](https://metrumresearchgroup.github.io/PKPDmisc/reference/read_nonmem.html).
+#' consider using `data.table::fread()` with the `skip` and `nrows` arguments.
 #'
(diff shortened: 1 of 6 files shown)


seth127:
First commit makes the change, second one does some unrelated housekeeping of the docs.

kyleam (approved):

PR #565, opened 2022-12-01
use_bbi() directory feature/bug

### LINKED ISSUE #552: discrepancy in dir vs file path between options and use_bbi()
# Current state

* `options("bbr.bbi_exe_path")` needs to be set to a path to a working `bbi` executable (i.e. the _file_)
* `use_bbi()` defaults to installing `bbi` to the path set in `options("bbr.bbi_exe_path")`.
* `use_bbi()` can accept either:
  *  a path to a _file_ (the path where the installed executable will be)
  * a path to a _directory_ that the executable should be installed into. If it gets a directory path (i.e. `"/data/apps"`) it will install the binary _within_ that directory (i.e. `"/data/apps/bbi"`).
* If `options("bbr.bbi_exe_path")` is set to a _directory_, [`check_bbi_exe()`](https://github.com/metrumresearchgroup/bbr/blob/main/R/bbr.R#L178-L195) will fail whenever something tries to call `bbi` (i.e. it won't be able to find the executable)

# Resulting confusion

This results in a confusing state where the user can do the following:

* set `options("bbr.bbi_exe_path" = "/data/apps")`
* call `use_bbi()` and have `bbi` successfully install to `/data/apps/bbi`
* try to use `bbi` and have it fail, saying it can't find `bbi` at `/data/apps`

(This actually happened in a MeRGE training, which is where this issue came from.)

# Solution options

Below are some options I see. They are in order of my current preference, though I'm very open to feedback.

1. If [`bbi_exec()`](https://github.com/metrumresearchgroup/bbr/blob/main/R/bbr.R#L72-L79) (or likely [`check_bbi_exe()`](https://github.com/metrumresearchgroup/bbr/blob/main/R/bbr.R#L178-L195)) is passed a directory path, let it look for `bbi` inside that directory, and (if found) ask the user if they would like to reset `options("bbr.bbi_exe_path")` to the executable path.
  a. Potentially, first check that `options("bbr.bbi_exe_path")` is indeed set to that directory path, because at that point we won't technically know if that's where the path came from.
  b. If it _isn't_ set to that directory path (again, I think this is unlikely, but possible) we proceed to error, as we currently do [here](https://github.com/metrumresearchgroup/bbr/blob/main/R/bbr.R#L185).
3. If `bbi_exec()` (or likely `bbi_check_exe()`) is passed a directory path, let it silently look for `bbi` inside that directory, and use that instead.
4. If `bbi_exec()` (or likely `bbi_check_exe()`) is passed a directory path, error and say something like "`some/dir/you/passed` is a directory, did you mean `some/dir/you/passed/bbi`?"
5. Make `use_bbi()` no longer accept a directory path, so that we catch the issue sooner and make sure users are setting the path to the executable instead of the directory.

seth127:
Closes #552

### DIFF
--- a/R/use-bbi.R
+++ b/R/use-bbi.R
@@ -1,8 +1,8 @@
 #' @title Installs most current release of bbi
 #' @description Identifies system running and pulls the relevant tarball for the
-#'   current release of bbi from GitHub, and then installs it in the directory
-#'   passed to `.dir`. If used in an interactive session, will open an
-#'   installation menu confirming the installed version. This function will
+#'   current release of bbi from GitHub, and then installs it at `.path` (see
+#'   Details section for defaults). If used in an interactive session, will open
+#'   an installation menu confirming the installed version. This function will
 #'   print information about the installed version. This **printing can be
 #'   suppressed** by setting `options(bbr.verbose = FALSE)`.
 #'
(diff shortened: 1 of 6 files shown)


kyleam (approved):
LGTM, thanks

PR #568, opened 2022-12-09
Function for converting PsN runs to bbr model objects

barrettk:


barrettk:
Current concerns:
 - multiple `modelfit_dir`'s 
      - wouldn't always be the max --> could be a bootstrap or some other method where multiple fits occur. 
 - multiple `NM_run`'s
      - Multiple folders are generated if the model fails to converge for parameter estimates. I think it can happen for other **_intentional_** scientific reasons as well though.
 -  PsN calls outside of `execute` (i.e. `bootstrap`, `scm`, etc.)
 - The threads attribute I pulled from the PsN generated `meta.yaml` was wrong. It said I had 5 threads, when A) I never specified any (should be 1), and B) I only had 4 cores for my workflow. However, its possible this is the default value, and it wasn't actually used since it wasn't parallelized, which brings me to the next point...
 - Couldnt find a setting for whether the model was run in parallel or not.
 - Some parameters were not retrievable to match a legitimate bbi json file
      - The way I handled the `nonmem` chunk of the bbi json may not be good enough
 - We should probably append the json with a parameter indicating whether the model was run via PsN or not. We could then use this parameter to prevent certain functions from working (if there's no workaround). Would need to have a deeper discussion about appending the bbi json though; just in case it has other implications im not aware of.
 - Other assumptions I mentioned in inline code comments

barrettk:
Just realized I cant take the max of a folder (like I was doing for `NM_run` directories) because they use literal sorting rather than numeric:

```r
psn_fit_dirs <- psn_fit_dirs[grepl("modelfit", psn_fit_dirs)] %>% sort()
> psn_fit_dirs
 [1] "/data/pirana_examples/modelfit_dir1"  "/data/pirana_examples/modelfit_dir10" "/data/pirana_examples/modelfit_dir11"
 [4] "/data/pirana_examples/modelfit_dir12" "/data/pirana_examples/modelfit_dir13" "/data/pirana_examples/modelfit_dir14"
 [7] "/data/pirana_examples/modelfit_dir2"  "/data/pirana_examples/modelfit_dir3"  "/data/pirana_examples/modelfit_dir4" 
[10] "/data/pirana_examples/modelfit_dir5"  "/data/pirana_examples/modelfit_dir6"  "/data/pirana_examples/modelfit_dir7" 
[13] "/data/pirana_examples/modelfit_dir8"  "/data/pirana_examples/modelfit_dir9" 
> max(psn_fit_dirs)
[1] "/data/pirana_examples/modelfit_dir9"
```

barrettk:
@seth127 it seems the pirana examples dont have a `meta.yaml` file, which was introduced in [PsN 4.8.0](https://uupharmacometrics.github.io/PsN/)
image

Given that we're on 5.2.6, we must have executed these models a long while back. Id really prefer not to remove the `meta.yaml` dependency given all the submission info it captures, but open to suggestions. Will see what other options we have.

barrettk:
@kyleam have you seen anything like the most recent drone failure in this PR? Seems like it failed during cloning with a bizarre message. Not the first time i've seen this.

kyleam:
@barrettk Hmm, that is a bizarre failure.  From drone:

```
+ git fetch origin +refs/heads/main:
From https://github.com/metrumresearchgroup/bbr
 * branch            main       -> FETCH_HEAD
 * [new branch]      main       -> origin/main
+ git checkout main
Branch 'main' set up to track remote branch 'main' from 'origin'.
Switched to a new branch 'main'
+ git fetch origin refs/pull/568/head:
From https://github.com/metrumresearchgroup/bbr
 * branch            refs/pull/568/head -> FETCH_HEAD
+ git merge d4f5b5ef1d498904dc19886141d065ce88634bb4
merge: d4f5b5ef1d498904dc19886141d065ce88634bb4 - not something we can merge
```

Nothing there looks strange, aside from the failure message of course, and the commit is something it can merge:

```
$ git checkout origin/main
Previous HEAD position was d4f5b5ef Bug fix: now properly finds the latest NM_run folder
HEAD is now at 89f9f930 Merge pull request #567 from metrumresearchgroup/release/1.5.0
$ git fetch origin refs/pull/568/head:
From github.com:metrumresearchgroup/bbr
 * branch              refs/pull/568/head -> FETCH_HEAD
$ git rev-parse FETCH_HEAD
d4f5b5ef1d498904dc19886141d065ce88634bb4
$ git merge FETCH_HEAD
Updating 89f9f930..d4f5b5ef
Fast-forward
 NAMESPACE                    |   1 +
 R/convert-psn.R              | 314 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 man/convert_psn.Rd           |  50 ++++++++++++++
 man/convert_psn_data_path.Rd |  17 +++++
 man/copy_psn_run_files.Rd    |  27 ++++++++
 man/copy_psn_tables.Rd       |  24 +++++++
 man/create_psn_json.Rd       |  17 +++++
 7 files changed, 450 insertions(+)
 create mode 100644 R/convert-psn.R
 create mode 100644 man/convert_psn.Rd
 create mode 100644 man/convert_psn_data_path.Rd
 create mode 100644 man/copy_psn_run_files.Rd
 create mode 100644 man/copy_psn_tables.Rd
 create mode 100644 man/create_psn_json.Rd
```

```
$ git log --oneline origin/main..d4f5b5ef1d498904dc19886141d065ce88634bb4
d4f5b5ef (HEAD, origin/feat/PsN_output_helper, refs/pull/origin/568) Bug fix: now properly finds the latest NM_run folder
6668585b refactored conversion to use better arguments
aedeeba5 updated .mod_path variable definition for several helper functions
abfee5c8 added clarification comment
d3af42a0 incomplete comment
9160b12d PsN converter is now functional (with a lot of assumptions)
e1a71fc5 Psn output conversion (WIP)
```

Does it persist if you restart?

barrettk:
> Does it persist if you restart?

Restart my R session? I dont have any issues locally. I bet my next commit won't run into the drone issue too (just a hunch), but I dont see what restarting would do. Are you suggesting restarting before my next commit?

kyleam:
> Restart my R session?

The drone build, via the button on the right:

image

barrettk:
FYI @seth127 

I discovered we will likely need additional handling for **model and data paths** in the event the model was **executed on a different system**, and then just copied over. This causes the `meta.yaml` and/or `version_and_option_info.txt` file to reference file paths that do not exist on the system.

I added a bunch of handling/guessing for model paths, but data paths have not been handled yet. The pirana examples wont work without allowing users to specify a datapath as well (or we could just error out). 

At the moment, **if no model file is specified**, we will do one of two things:
 - if PsN is greater than 4.8.0 (if theres a yaml file), we will grab the data path described there (accurate if run on the same machine)
 - If PsN is less than 4.8.0 (there's no yaml), we will grab the directory from the text file (`modelfit_dir`), and check to see if the model file is there. If not, error out and prompt the user to specify a path. We _could_ grab the `psn.mod`, but that has the aforementioned formatting issues.

To handle datapath discrepancies however, we only have the option of either A) erroring out if we cant find it, or B) allowing the user to specify a datapath (would only ever be used for models run on a different machine). In all honesty I dont like either solution, but would lean towards option B. If we chose option A), all the model_path handling I just implemented would  pointless, since we'd error out as soon as we recognized some file paths didnt exist.

seth127:
This is all looking good @barrettk. Overall, I think we should make a few more tweaks and then reach out to some users to test is out and give feedback. Discussion on those below:

# Path answers

> I added a bunch of handling/guessing for model paths...  

This sounds good. I think we want to make best effort to find the original model file, but if we can't then we just need to error and tell the user to pass it in via `. psn_model_file`. I don't think we want to grab the badly formatted one in the run dir. If they really want to use that, then they can pass that path manually.

> data paths have not been handled yet... we have the option of either A) erroring out if we cant find it, or B) allowing the user to specify a datapath 

I think we should basically do a two-step thing:

1. try to find the model file (error if we can't, as we do now)
2. parse the `$DATA` from that and then search for the data file from there. That path will be relative to the location of the model file that we just found. If we can't find it, let's give a **warning** saying `"expected to find data at {path we parsed} but no file is there. Please adjust $DATA in {path to new control stream}."` I don't think we actually need to error (though we'll want to get user feedback on this)

One more consideration, copied from above, when dealing with the data path:

> * If `.bbr_dir` is not `dirname(.modelfit_dir)` (aka the proposed default), then we should probably modify the data path to be relative to the new location of the control stream. This might take some `fs::path_rel()` gymnastics.

Basically we want to be sure that, if we found the data file, the new copied `.mod` file still has the right path to it.

# Tangent about `.mod` file in execution dir
One last note... I had written this above:

> When copying the .mod control stream, we don't change the `$DATA` path in the original `.mod` file, but we _do_ need to add `../` to the `$DATA` path in the `.mod` that is copied into the output/execution dir. _This is to imitate what bbi would've done if it ran this model._

but now I'm thinking... do we even need to copy in the `.mod` file to the execution directory? Do we use that for anything, once execution is finished? I'm thinking we might want to try not copying it and see if anything breaks. It just seems... unnecessary and confusing. (To be clear, I know we need [this](https://github.com/metrumresearchgroup/bbr/blob/590cdd2e45ec1562aab4aff902f46b4425371645/R/convert-psn.R#L82-L83), I'm talking about getting rid of [this](https://github.com/metrumresearchgroup/bbr/blob/590cdd2e45ec1562aab4aff902f46b4425371645/R/convert-psn.R#L94-L95).)

seth127:
> what are your thoughts on keeping it just for consistency's sake?

At this point I'm thinking we should _not_ copy it, unless we need it, for the following reasons:

* We want these to function as if run by `bbr`, but in terms of consistency, I think it's probably beneficial if it's obvious that this actually wasn't run with `bbr`, and so some of the reproducibility promises aren't the same. In that sense, some inconsistency might be semi-desirable, because it just makes that more obvious to future users.
* This copied control stream is really just an artifact of how `bbi` executes. Similar to previous comment: missing it just shows that it wasn't run with `bbi`.
* The copied control stream isn't necessary for re-running the model. In fact, if the user decides to re-run the model, it would get overwritten anyway.

seth127:
Testing this and it's looking good. Some small changes.

* `.data_path` in the `bbi_config.json` should be relative or else `get_data_path()` breaks. [This line](https://github.com/metrumresearchgroup/bbr/blob/ee1b62f98450fa88d23ed5bfa6772c6376cda5a4/R/convert-psn.R#L400) should be `data_path = fs::path_rel(.data_path, .bbr_run_dir),`
* _Don't_ add `.tab` to the table files. Copy them as is.
* Warn instead of error when we can't find the table files referenced in the control stream
* Make `.bbi_dir` a required argument

R/convert-psn.R
```diff
@@ -21,20 +20,23 @@
...
+    stop("`.bbr_dir` cannot be set to the same location as `.modelfit_dir`")
+  }
+
 
-  # TODO: Make sure bbi path is set BEFORE copying any files
+  # Make sure bbi path is set BEFORE copying any files
   # this is needed to create a fake bbi json, and we dont want to copy everything just to fail at the end
-  # is_empty wont capture all cases - revisit how to properly address this concern
-  if(rlang::is_empty(bbi_version())){
+  if(!nzchar(bbi_version())){
```
seth127, line 39: maybe I'm missing something but do they actually need to have `bbi` installed to run this? They'll need it to do `model_summary()`, which is a likely next step. But I don't think we actually need `bbi` for this function, right?

R/convert-psn.R
```diff
@@ -193,7 +182,7 @@ get_psn_submission_info <- function(.modelfit_dir, .psn_model_file = NULL){
 #'
 #' @keywords internal
 parse_psn_submission_info <- function(.submission_file, .psn_model_file){
-# browser()
+  # browser()
```
seth127, line 185: delete this browser() call

inst/dev-examples/convert-psn-to-bbr.R
```diff
@@ -54,7 +54,10 @@ mod_sum %>% get_theta()
 # one of the nice things about bbr is how it integrates with other MeRGE packages
 # here we show how to make simple diagnositic plots with the pmplots package
 #   see more here: https://merge.metrumrg.com/expo/expo1-nonmem-foce/posts/intro-to-pmplots.html
-if (!nzchar(packageVersion("pmplots"))) remotes::install_github("metrumresearchgroup/pmplots")
+if(require(pmplots)){
```
seth127, line 57: Using `require()` is fine, but you have the logic backwards. This is saying to install _if_ it's already installed.

You want the following (without my comments)

```
if (!require(pmplots)) { # load the library, if not installed `!require` will return TRUE
  remotes::install_github("metrumresearchgroup/pmplots") # install
  library(pmplots) # load the library, now that it's installed
}
```

barrettk: Ahh good catch. Yeah I found that `nzchar` returned an error for me. Didnt look into it too much, but with this fix I think it will be more reliable

seth127:
This looks good. As stated previously, we're not going to merge this now because we want to get some user feedback (and write some formal tests, etc.).

Issue #398, opened 2021-09-24
edit_model()


seth127:
Opens control stream in Rstudio GUI with `file.edit()`

# Discussions points

* might want to look into what `file.edit()` does when you're _not_ in an interactive Rstudio session (i.e. in the terminal or in an Rmd that gets knit)
* look at [comment below](https://github.com/metrumresearchgroup/bbr/issues/398#issuecomment-926668621) about other similar helpers.

# Proposed code

```r
#' Helper to open control stream for editing in Rstudio
#' @param .mod a `bb{.model_type}_model` object
edit_model <- function(.mod){
  get_model_path(.mod) %>% 
    file.edit()
}
```

seth127:
Consider adding `edit_` (or possibly `view_`) helpers (all using `file.edit()` for other files too:

```r
view_output_file <- function(.mod, .extension){
  ## grab file with .extension
  tmpdir <- get_output_dir(.mod)
  tmpfile <- list.files(tmpdir, pattern=regex(.extension, ignore_case = TRUE))
  
  ## return NULL if file doesn't exist
  if(length(tmpfile) == 0){
    warning("File does not exist")
    return(NULL)
  }
  
  ## open file
  file.edit(file.path(tmpdir,tmpfile))
}

view_lst <- function(.mod){
  view_output_file(.mod, .extension="lst")
}
```

kyleam:
We've added various helpers to bbr.bayes for Stan-related files, including the model file.  Notably, there's a file.edit kludge to make it work better with RStudio server.

So, if we add the proposed function

 * we should absorb the file.edit kludge from bbr.bayes
 * adjust bbr.bayes to 1) remove the kludge and 2) make its "open model" helper an alias or wrapper around bbr's function

seth127:
I like this idea @kyleam . We should also settle on naming for all of these... I think the `open_*` convention from the `bbr.bayes` is my current favorite.

PR #570, opened 2023-01-03
add open_model_file

kyleam:
This PR brings in the `file_edit()` helper from bbr.bayes and adds the "open control stream" helper proposed in #398 ([related bbr.bayes discussion](https://github.com/metrumresearchgroup/bbr.bayes/pull/28#issuecomment-1359947078)).

This doesn't come with any automated tests (or associated validation doc updates) because I don't have any good ideas for how to meaningfully test this, but please let me know if you have any suggestions.  I've tested it manually in RStudio on model paths, model objects, and model summary objects.

---

From gh-398:

> might want to look into what file.edit() does when you're not in an interactive Rstudio session (i.e. in the terminal or in an Rmd that gets knit)

`file.edit` doesn't play well with a non-interactive session:

image

Given the interactive/convenience nature of `file.edit()` and the `open_model_file` wrapper, that behavior seems okay to me.  @seth127, do you think we should add special handling for the non-interactive context?


kyleam:
The plan is to rewrite bbr.bayes to use this.  I've pushed [another branch](https://github.com/metrumresearchgroup/bbr/compare/bbr-bayes-split...open-model-file-bb) that merges this PR's branch into `bbr-bayes-split`.  I'll open a draft PR in the bbr.bayes repo that pulls that branch in.

update steps:

- [ ] bbr repo: merge this current PR into main
- [ ] bbr repo: merge main into bbr-bayes-split
- [ ] bbr repo: bump dev version and push dev tag
- [ ] bbr.bayes PR: switch drone to dev tag

seth127:
Thanks for doing this Kyle.

> Given the interactive/convenience nature of file.edit() and the open_model_file wrapper, that behavior seems okay to me. @seth127, do you think we should add special handling for the non-interactive context?

I do think it would be nice if we could handle this more gracefully (i.e. maybe just return NULL). My thinking is that people will likely put this code into things like RMD docs and then might later knit that without remembering to comment them out.

What do you think about just putting a `if (!interactive()) return(invisible(NULL))` at the top of [our `file_edit()`](https://github.com/metrumresearchgroup/bbr/blob/c9100f013bcc256c0ec31d71fefee0a2435cfe58/R/open.R#L25-L36)? Is there a downside to that?

R/open.R
```diff
@@ -0,0 +1,36 @@
...
+#' Open main model definition file in text editor
+#'
+#' This is a light wrapper around [get_model_path()] to enable quickly jumping
+#' to a model definition file in RStudio or any other editor supported by
+#' [utils::file.edit()].
+#'
+#' @param .mod Model object or a file path to a model.
+#' @param ... Other arguments passed on to methods.
+#' @export
+open_model_file <- function(.mod, ...) {
```
kyleam, line 11: Hmm, actually, I don't see much reason to go with the method here.  (I'm not spotting where that was every explicitly proposed, so I'm not sure why I went that way.). `get_model_path` should return what's necessary for any model type and having `.character` method is a little silly because you'd just call `file.edit` if you already had the path.

I'll plan to rework this.

kyleam:
> What do you think about just putting a if (!interactive()) return(invisible(NULL)) at the top of [our file_edit()](https://github.com/metrumresearchgroup/bbr/blob/c9100f013bcc256c0ec31d71fefee0a2435cfe58/R/open.R#L25-L36)? Is there a downside to that?

Encourage people to leave function calls in their Rmd that shouldn't be there :). But, no, I don't see any real downsides.  I'll make that change.  Are you okay with also giving a warning?

seth127 (approved):
looks good

PR #577, opened 2023-02-14
friendlier messages for delete_models()

### LINKED ISSUE #560: Improve `delete_models()` messages
`delete_models()` was originally conceived for cleaning up models created by `test_threads()` and for this reason it defaults to `.tags = "test threads"`. This is fine, but when attempted to delete an arbitrary model, you see this:

```r
> delete_models(mod107)
The following tags were not found:
- test threads 

Error in delete_models(mod107) : None of specified tags were found
```

This makes sense, but we could probably add a more helpful message about that being the default value and that users should pass `.tags = NULL` to delete arbitrary models, regardless of tag.

Also, when the user _does_ pass `.tags = NULL` they get this message about `NA`

```r
> delete_models(mod107, .tags = NULL)
Are you sure you want to remove 1 models with the following tags?: `NA` (Yes/no/cancel) yes
Removed 1 models with the following tags:
- NA
```

That message could also be better. I'm not sure what the right message is in that case, but maybe just not mentioned tags at all (if the user passes `.tags = NULL`).

seth127:
Closes #560

### DIFF
--- a/R/test-threads.R
+++ b/R/test-threads.R
@@ -290,10 +290,14 @@ delete_models <- function(.mods, .tags = "test threads", .force = FALSE){
     if (!isTRUE(delete_prompt)) return(invisible(NULL))
   }
 
-  msg_remove <- paste0(
-    paste("Removed", length(mod_paths), "models with the following tags:\n"),
-    paste("-",mods_removed, collapse = "\n")
-  )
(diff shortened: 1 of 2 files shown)


kyleam (approved):
LGTM

PR #579, opened 2023-02-23
use_bbi fixes

### LINKED ISSUE #576: windows: use_bbi issues
Two problems were encountered using `use_bbi` on Windows:

 1. no bbi reported as being installed despite `use_bbi` downloading it

 2. the "please either set..." message being shown despite that being set

The first one looks to be because `Sys.which` will come up empty on Windows unless the path ends with ".exe".  When on Windows, we might want to check that the binary path ends with ".exe", but we should at least change `build_bbi_install_path` so that it tacks on ".exe":

https://github.com/metrumresearchgroup/bbr/blob/d4453a98a63051b76dcc20380e69646a1acd9177/R/use-bbi.R#L224-L227

Note also that there is already this:

https://github.com/metrumresearchgroup/bbr/blob/d4453a98a63051b76dcc20380e69646a1acd9177/R/aaa.R#L1-L5

The second issue is because `install_menu` uses `normalizePath`, and `add_to_path_message` compares that value to `bbr.bbi_exe_path`.  `normalizePath` by default converts "/" to "\\", so the `bbi_exe_path` value needs to use the same form.  (Note that `file.path` on windows uses "/" by default.)  This fix here is probably to make `add_to_path_message` feed `bbr.bbi_exe_path` to `normalizePath` before comparing.

kyleam:
This series fixes a few `use_bbi` issues.  These were prompted/surfaced by Windows testing (see gh-576), but most of these are also problems outside of Windows.

Main issues fixed:

 - "set bbr.bbi_exe_path or add to PATH" message was shown even if caller aborted

 - "set bbr.bbi_exe_path or add to PATH" message misfired on valid `bbr.bbi_exe_path` if it didn't exactly match the `normalizePath()` output

 - "set bbr.bbi_exe_path or add to PATH" did _not_ fire if the paths directory was in `PATH` but overshadowed

 - relative path like `use_bbi("bin/bbi")` triggered a dev error

 - ensure target path on Windows ends with ".exe"

Closes #576

(I'll have one followup PR that uses this as a base. edit: that's gh-581)

---

todo:

- [x] final round of interactive testing on windows
- [x] adjust for compatibility with mpn oldest ([failing build](https://github-drone.metrumrg.com/metrumresearchgroup/bbr/3432/2/1))

### DIFF
--- a/R/aaa.R
+++ b/R/aaa.R
@@ -1,4 +1,5 @@
-BBI_DEFAULT_PATH <- if (.Platform$OS.type == "windows") {
+ON_WINDOWS <- identical(.Platform$OS.type, "windows")
+BBI_DEFAULT_PATH <- if (ON_WINDOWS) {
   "bbi.exe"
 } else {
   "bbi"
--- a/R/use-bbi.R
+++ b/R/use-bbi.R
@@ -341,9 +341,18 @@ version_message <- function(local_v, current_v){
                  .trim = FALSE))
 }
 
-try_path_real <- function(...) {
-  tryCatch(fs::path_real(...),
-           ENOENT = function(e) NULL)
+if (utils::packageVersion("fs") < "1.4.2") {
+  try_path_real <- function(path) {
+    if (!all(fs::file_exists(path))) {
+      return(NULL)
+    }
+    fs::path_real(path)
+  }
+} else {
+  try_path_real <- function(path) {
+    tryCatch(fs::path_real(path),
+             ENOENT = function(e) NULL)
+  }
 }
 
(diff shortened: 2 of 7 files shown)


kyleam:
> [prevent indented newline from being trimmed](https://github.com/metrumresearchgroup/bbr/pull/579/commits/e70c368239c510074e098601f30c8aa35c18d210)

Don't think this is going to trip anyone up, but that should say "intended" not "indented" :x

seth127 (approved):
These all look good to me. Per usual, I appreciate the detailed commit messages.

PR #580, opened 2023-02-23
set bbr.bbi_exe_mode default to 'local' unless on linux

kyleam:
"local" is likely the better default for a user on Windows or macOS (though based on light searching, it may be possible to install SGE on Windows and macOS).


seth127 (approved):
As discussed offline, I think this is a good change.

PR #581, opened 2023-02-23
check_bbi_exe: warn on Windows unless bbi is at least v3.2.2

kyleam:
bbi 3.2.2 has Windows fixes for `bbi init` and `bbi nonmem run`. Without these, submitting a model is unlikely to be successful, but still warn rather than error because 1) there may be some configurations/setups where a previous version would work and 2) older bbi versions should work for older commands like `bbi nonmem summary` that don't depend on running NONMEM.

---

Merging should wait until:

- [x] base has been merged (#579)
- [x] bbi 3.2.2 has been released


seth127 (approved):
Good stuff.

PR #583, opened 2023-02-24
tests: account for bbr.bbi_exe_mode default change

kyleam:
When test-submit-model.R and test-submit-models.R check the dry run output, they assume the default mode is sge, but that's no longer the case as of e4619f05 (set bbr.bbi_exe_mode default to 'local' unless on linux, 2023-02-23).

Instead grab the value from bbr.bbi_exe_mode.  While touching these lines, also move the leading prefix into a variable because that doesn't change for most calls.

---

 * gh-580 introduced test-submit-model.R and test-submit-models.R failures on Windows.  The test suite isn't passing on Windows yet, so this is just about not introducing new ones.

 * This change will also put us in a better spot for the upcoming SLURM transition.


seth127 (approved):
Nice. These changes make sense, and also make the tests a bit more readable, which doesn't hurt.

Also this

> This change will also put us in a better spot for the upcoming SLURM transition.

is a great point.

PR #588, opened 2023-03-07
DO NOT MERGE: Feat/inherit estimates

barrettk:
Will update later

`inherit_param_estimates` will copy over parameter estimates from a parent model and overwrite the control stream file


barrettk:
@seth127 is it safe to close this PR?

seth127:
I think so (and we can delete the branch too). 

Just for posterity: we're closing this because this work was completed in #623 instead.

PR #592, opened 2023-04-12
Prepare for 1.7.0 release

kyleam:
* mark some things as internal to get a clean pkgdown build

* NEWS item (this release is a bit different than usual, so let me know if you have thoughts about what we should say/include here)

* DESCRIPTION version update

changeset: https://github.com/metrumresearchgroup/bbr/compare/1.6.0...release/1.7.0

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: bbr
 Title: R package for bbi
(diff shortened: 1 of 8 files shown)


seth127 (approved):
Looks good to me. Can't wait for that `bbr.bayes` to drop...

PR #593, opened 2023-04-13
DESCRIPTION: add pkgdown site to URL field

kyleam:
The pkgdown site needs to be listed in the DESCRIPTION's URL field in order for pkgdown to auto-link to bbr pages when building another package's site.

https://pkgdown.r-lib.org/articles/linking.html#across-packages

---

With the bbr.bayes pkgdown site, I've confirmed that this change is sufficient to get things like "\`new_model()\`" to auto-link to <https://metrumresearchgroup.github.io/bbr/reference/new_model.html> instead of the fallback target of <https://rdrr.io/pkg/bbr/man/new_model.html>.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -47,7 +47,7 @@ Description: R package for making calls to bbi for running
     NONMEM and other models. Also manages model metadata, workflow, and
     run logs.
 License: MIT + file LICENSE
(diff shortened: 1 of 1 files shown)


seth127 (approved):
Nice work

Issue #599, opened 2023-06-29
Refactor current functions to use `nmrec`


seth127:
There are several current `bbr` functions that rely on raw string parsing of the control stream and would likely be more robust if they used the new `nmrec` package instead. Specifically, once the changes from https://github.com/metrumresearchgroup/nmrec/pull/5 are released (discussion on release begins [here](https://github.com/metrumresearchgroup/nmrec/pull/5#issuecomment-1613418533)).

Several examples:

* [x] `test_threads()`, notably [here](https://github.com/metrumresearchgroup/bbr/blob/main/R/test-threads.R#L351-L376)
* [x] `nm_table_files()`, specifically [here](https://github.com/metrumresearchgroup/bbr/blob/main/R/nm-tables.R#L90-L96)
* [x] `copy_control_stream()` (internal), specifically [here](https://github.com/metrumresearchgroup/bbr/blob/main/R/copy-model-from.R#L227-L232) (side note, it's not clear to me that [this line](https://github.com/metrumresearchgroup/bbr/blob/main/R/copy-model-from.R#L232) needs to be there...)
* [ ] `update_model_id()`, really [most of the function](https://github.com/metrumresearchgroup/bbr/blob/main/R/copy-model-helpers.R#L74-L97)
   skipped: see discussion starting at https://github.com/metrumresearchgroup/bbr/pull/605#issuecomment-1660822495

These don't need to be done all at once, and some may not be worth doing at all, but they were related enough to put them in the same issue. If anyone sees more cases like this in the code base, feel free to add them to this issue.

kyleam:
> Specifically, once the changes from https://github.com/metrumresearchgroup/nmrec/pull/5 are released

This is probably for discussion outside of GitHub, but, in the case for bbr (at least for non-optional code), won't a bbr release that depends on nmrec need to wait for nmrec make it to MPN?

kyleam:
And thanks for pulling together these examples.

seth127:
> won't a bbr release that depends on nmrec need to wait for nmrec make it to MPN?

Hmmm, good thought there. Yes, let's discuss offline. I'm not sure that's a hard requirement, but probably is the smoothest path forward.

kyleam:
I don't think there's anything in-flight here.  It probably makes sense to open a dedicated PR for anything else that comes up, but please reopen if I'm missing something.

PR #600, opened 2023-06-30
test-threads refactor

barrettk:
- now relies on `nmrec` instead of string manipulation
 - we should require a miniumum version of nmrec once it's officially released

addresses one of the bullets in https://github.com/metrumresearchgroup/bbr/issues/599


barrettk:
@kyleam This PR is ready to review with the exception of being able to fetch `nmrec` and require a minimum version in the DESCRIPTION file

kyleam:
Okay, thanks.  I'm working on $PROB support for nmrec.  That shouldn't take long, so I'd like to get that into the next nmrec release.

Two initial thoughts :

 * Using nmrec in this case can essentially be viewed as a fix.  It's a good thing if the current tests pass for you, but shouldn't this also add regression cases for at least a few examples that were broken with the old code?

 * Before this is merged, we should discuss (offline) how to handle nmrec as a dependency: <https://github.com/metrumresearchgroup/bbr/issues/599#issuecomment-1613813457>.  It's probably best to mark this as a draft until then, even when it's ready for a full review.  (What we decide in that discussion main mean that more than an nmrec release would block the merge here.)

Also, if you'd like to get a passing drone before the final adjustments, you can temporarily add something like

```
  - git clone --depth 1 https://github.com/metrumresearchgroup/nmrec.git /tmp/nmrec
  - R -s -e 'devtools::install("/tmp/nmrec", upgrade = "never")'
```

to the drone install steps.

barrettk:
> * Using nmrec in this case can essentially be viewed as a fix.  It's a good thing if the current tests pass for you, but shouldn't this also add regression cases for at least a few examples that were broken with the old code?

I initially didn't think there were any cases where the old code would have broken, but after thinking on it a bit there definitely were - mainly models that used one of the weird spellings of `maxevals` or something. We captured all the frequent cases with regex, but definitely did not support all of them (dont think we supported lower case either). Can add a test for this type of thing unless there were any other cases you had in mind. Im changing some of the tests now to rely on nmrec as well, so our existing tests will be a bit better.


> * Before this is merged, we should discuss (offline) how to handle nmrec as a dependency: [Refactor current functions to use `nmrec` #599 (comment)](https://github.com/metrumresearchgroup/bbr/issues/599#issuecomment-1613813457).  It's probably best to mark this as a draft until then, even when it's ready for a full review.  (What we decide in that discussion main mean that more than an nmrec release would block the merge here.)

Yeah this sounds good. As an aside, my intention was just to address the test-threads changes in this PR, but we could broaden the scope to include all the changes mentioned in that issue as well.


>Also, if you'd like to get a passing drone before the final adjustments, you can temporarily add something like
```
  - git clone --depth 1 https://github.com/metrumresearchgroup/nmrec.git /tmp/nmrec
  - R -s -e 'devtools::install("/tmp/nmrec", upgrade = "never")'
```

Thanks, will try this out once I fix the tests

barrettk:
@kyleam FYI it seems like it's failing solely because of `gregexec` in `nmrec:::split_to_elements`, which is not included in R 4.0. I have reported a [similar issue](https://github.com/davidgohel/flextable/issues/457) in the past for an MPN package, so im wondering if we want to make a fix for this on the `nmrec` side

kyleam:
> FYI it seems like it's failing solely because of `gregexec` in `nmrec:::split_to_elements`

Thanks, didn't realize that was a recent addition.  Fixed in https://github.com/metrumresearchgroup/nmrec/pull/11

kyleam:
> We captured all the frequent cases with regex, but definitely did not support all of them (dont think we supported lower case either).

Hmm, I'm not as confident as you are about what's frequent, but, yeah, the idea is that, by using nmrec, you support all the spellings, as well as little things that the previous bbr code doesn't consider.  For example, the option name could be `NBURN = 1000` or even `NBURN 1000`.

> Can add a test for this type of thing unless there were any other cases you had in mind.

Thanks.  I don't care about any particular case (and I don't think it's necessary to exhaustively cover weird cases; that's nmrec's job) but having a few cases that wouldn't work before this change is a good thing in my view.

> As an aside, my intention was just to address the test-threads changes in this PR, but we could broaden the scope to include all the changes mentioned in that issue as well.

My vote is to keep this PR focused on test-threads.

barrettk:
@kyleam got drone to work by re-running it. Seems like we will wait to actually merge this until nmrec is officially released, but figured i'd still ask for a review now (no rush).

R/test-threads.R
```diff
@@ -330,90 +330,59 @@ adjust_estimation_options <- function(.mods, .cap_iterations){
...
-        } else if(any(niter_detect)) {
-          est_block.i[niter_detect] <- replace_est_opt(est_block.i[niter_detect], "NITER=\\d+", .cap_iterations)
-          if (any(nburn_detect) && !any(str_detect(est_block.i, "NBURN=0\\b"))) {
-            est_block.i[nburn_detect] <- replace_est_opt(est_block.i[nburn_detect], "NBURN=\\d+", .cap_iterations)
-          }else if(!any(nburn_detect)){
+      for(i in seq_along(ests)){
+        est_block.i <- ests[[i]]
+        est_block.i$parse()
+
+        max_detect <- any(purrr::map_lgl(est_block.i$get_options(), ~ .x$name == "maxevals"))
```
kyleam, line 347: This and the similar lines below are fine as is, but I'm guessing they came before the switch to making `get_record_option` return `NULL` in nmrec's [47228bd](https://github.com/metrumresearchgroup/nmrec/commit/47228bddb5736f371781c25bd48ac63daa6ca476). With that, this could also be written as is

```R
max_detect <- !is.null(get_record_option(est_block.i, "maxevals"))
```

But, again, I think either way is okay.

R/test-threads.R
```diff
@@ -330,90 +330,59 @@ adjust_estimation_options <- function(.mods, .cap_iterations){
...
+          nmrec::set_record_option(est_block.i, "niter", .cap_iterations)
+          if(nburn_detect && !check_record_val(est_block.i, "nburn", 0)) {
+            # We don't want to overwrite NBURN if it was set to 0
+            nmrec::set_record_option(est_block.i, "nburn", .cap_iterations)
+          }else if(!nburn_detect){
             # NITER was specified, but NBURN wasn't (which means the default is being relied on). Force specification to cap.
-            message(glue("     Adding NBURN declaration to {est_block.i[niter_detect]} in {basename(mod_path)}"))
-            est_block.i[niter_detect] <- glue("{est_block.i[niter_detect]} NBURN={.cap_iterations}")
+            message(glue("Adding NBURN={.cap_iterations} declaration to {basename(mod_path)} to avoid using the default value"))
+            nmrec::set_record_option(est_block.i, "NBURN", .cap_iterations)
```
kyleam, line 361: All of the other spots that call `nmrec::set_record_option` are modifying an existing option, so they keep whatever name the caller used originally, which is good. This spot is adding a new option, so the spelling you give will go in.

Makes sense, and I like that you took care to go with upper case here because it's my impression that that's the predominant style for NONMEM options.

tests/testthat/test-test-threads.R
```diff
@@ -20,25 +20,33 @@ cleanup_bbi <- function(.recreate_dir = FALSE) {
...
-    values <- map(1:length(str_values), ~ {
-      est_loc <- grepl("METHOD", str_values[[.x]])
-      est_method <- str_values[[.x]][est_loc]
-      str_loc <- grepl(search_str, str_values[[.x]])
-      eval_method <- gsub('[[:digit:]]+|=', '', str_values[[.x]][str_loc])
-      value <- as.numeric(gsub(glue('{search_str}|='), '', str_values[[.x]][str_loc]))
-      names(value) = paste0(est_method, ", ",eval_method)
-      value
-    }) %>% unlist()
+    ctl <- nmrec::parse_ctl(mod_lines)
```
kyleam, line 29: Okay as is, but `mod_lines` is only used in the `parse_ctl` call, so you could just use `read_ctl` directly.

tests/testthat/test-test-threads.R
```diff
@@ -20,25 +20,33 @@ cleanup_bbi <- function(.recreate_dir = FALSE) {
...
+          val <- as.numeric(opt$value)
+          names(val) <- opt$name_raw
+          val
+        } else{
+          NULL
+        }
+      }) %>% purrr::flatten_dbl()
+
+      if(!rlang::is_empty(opts)) opts else NULL
+    })  #%>% purrr::flatten_dbl()
```
kyleam, line 46: Leftover dead code?

tests/testthat/test-test-threads.R
```diff
@@ -127,42 +135,47 @@ withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
     mods_complex <- test_threads(mod_complex, .threads = c(2, 4), .cap_iterations = 100, .mode = "local", .dry_run = TRUE)
 
     # Dont overwrite NBURN if set to 0
-    max_evals <- get_est_options("MAXEVAL|NITER|NBURN", mods_complex)
+    max_evals <- get_est_options(mods_complex)
 
+    # Confirm that estimation method didnt change, and that MAXEVAL/NITER was preserved
```
kyleam, line 140: This is no longer confirming "the estimation method didn't change", right? (I think that's okay, but comment should be updated.)

Same comment for line 163.

tests/testthat/test-test-threads.R
```diff
@@ -173,57 +186,49 @@ withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
...
   })
 
-  test_that("test_threads(.dry_run=T) threads are set correctly [BBR-TSTT-003]", {
-    mod_threads <- lapply(mods, function(mod.x){mod.x$bbi_args$threads}) %>% unlist()
-    expect_equal(mod_threads[1], 2)
-    expect_equal(mod_threads[2], 4)
-  })
 
+  test_that("test_threads(.dry_run=T) correctly changes maxeval/niter: handles different spelling [BBR-TSTT-002]", {
+    # recursive test from adding `nmrec` dependency - lower case and different spellings previously not supported
```
kyleam, line 191: What's "recursive" referring to?

tests/testthat/test-test-threads.R
```diff
@@ -127,42 +135,47 @@ withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
...
 
-    max_evals <- get_est_options("MAXEVAL|NITER|NBURN", mods_complex)
+    max_evals <- get_est_options(mods_complex)
 
     for(i in 1:length(max_evals)){
-      expect_equal(unname(max_evals[[i]]), c(9999, 0, 10))
-      expect_equal(names(max_evals[[i]]), c("METHOD=1, MAXEVAL", "METHOD=BAYES, NBURN", "METHOD=BAYES, NITER"))
+      max_evals_i <- unlist(unname(max_evals[[i]]))
+      expect_equal(unname(max_evals_i), c(9999, 10, 0))
+      expect_equal(names(max_evals_i), c("MAXEVAL", "NITER", "NBURN"))
```
kyleam, line 178: Hmm, there aren't two "BAYES" methods in acop-fake-bayes.ctl:

<https://github.com/metrumresearchgroup/bbr/blob/feddd576f3f8cf9f091487b761391dedba481040/inst/model/nonmem/complex/acop-fake-bayes.ctl#L38-L40>

So I guess `get_est_options` *before* this PR was returning an incorrect result. Is that related to why the 0 and 10 swap places in the expected value?

tests/testthat/test-test-threads.R
```diff
@@ -173,57 +186,49 @@ withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
     )
   })
 
-  test_that("test_threads(.dry_run=T) threads are set correctly [BBR-TSTT-003]", {
-    mod_threads <- lapply(mods, function(mod.x){mod.x$bbi_args$threads}) %>% unlist()
-    expect_equal(mod_threads[1], 2)
-    expect_equal(mod_threads[2], 4)
-  })
 
+  test_that("test_threads(.dry_run=T) correctly changes maxeval/niter: handles different spelling [BBR-TSTT-002]", {
```
kyleam, line 190: Nice, thanks for adding tests that would fail before this PR's changes.

tests/testthat/test-test-threads.R
```diff
@@ -173,57 +186,49 @@ withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
...
+        expect_equal(names(max_evals_i), c("MAXEVAL", "NITER", "NBURN"))
+      }else if (i==2){ # modified model (NBURN=0 --> nburn=10)
+        expect_equal(unname(max_evals_i), c(9999, 10, 10))
+        expect_equal(names(max_evals_i), c("MAXEVAL", "NITER", "nburn"))
+      }else{ # created with test_threads (capped iterations)
+        expect_equal(unname(max_evals_i), c(5, 5, 5))
+        expect_equal(names(max_evals_i), c("MAXEVAL", "NITER", "nburn"))
+      }
+    }
+    delete_models(c(list(mod_complex_edit), mods_fake), .force = T)
```
kyleam, line 223: That `list` is unnecessary, right? At a glance, think that can just be `c(mod_complex_edit, mods_fake)`.

kyleam, line 223: Unlikely to actually bite us, but, for package code, better to use `TRUE` than `T` shorthand because `T` may be overridden to anything (set `T <- FALSE` for maximum confusion :x)

kyleam (approved):
The reworked approach looks good, and you fit it nicely into the previous existing conditions, making it easy to review regarding whether it preserves the existing intended logic.

The test code took me a bit longer to wrap my head around (and I'm not sure I completely succeeded), but I think that's mostly a matter of needing to refresh on what they were doing, not the propose changes (which in my view make things simpler and easier to understand). I left a few questions and comments on the tests, but nothing that I consider a blocker.

R/test-threads.R
barrettk: Yeah I was conflicted with the case - thought about determining the case of other options, but that seemed a bit overkill

tests/testthat/test-test-threads.R
barrettk: No it was correct before, but was formatted differently and IMO more confusingly. It would attach the method to each of the options, so both of those expectations corresponded to the same "BAYES" method. The flipping of the 10 and 0 is just because the order of the options changed

tests/testthat/test-test-threads.R
barrettk: No `delete_models` expects model objects or lists:
```r
>     delete_models(c(mod_complex_edit, mods_fake), .force = T)
Error in check_model_object_list(.mods, .mod_types = c(NM_MOD_CLASS, NM_SUM_CLASS)) :
Passed list must contain only model objects, but found 5 invalid objects at indices: 1, 2, 3, 4, 5

>     delete_models(c(list(mod_complex_edit), mods_fake), .force = T)
Removed 3 models with the following tags:
- test threads
```

tests/testthat/test-test-threads.R
barrettk: Fixed

tests/testthat/test-test-threads.R
barrettk: Fixed

kyleam:
> requested a review from [kyleam](https://github.com/kyleam) [now](https://github.com/metrumresearchgroup/bbr/pull/600#event-10298068177)

I've already approved.  Is there something in particular you want me to look at?  If not, merge if you're happy with its state.

tests/testthat/test-test-threads.R
kyleam: Hmm, I can't make heads or tails of it with a quick glance, but I'll take you word for it.  Thanks.

tests/testthat/test-test-threads.R
kyleam: Doh, because model object is a list underneath.  Thanks for the correction.

barrettk:
> I've already approved. Is there something in particular you want me to look at? If not, merge if you're happy with its state.

Yeah fair - I made some changes based on some of your comments, and just wanted to make sure you were good with the changes. Will merge now

tests/testthat/test-test-threads.R
barrettk: Oh jeez yeah brain fart

PR #603, opened 2023-07-21
nm_table_files: use nmrec instead of regex

barrettk:
Addresses the second bullet of issue https://github.com/metrumresearchgroup/bbr/issues/599

### DIFF
--- a/R/nm-tables.R
+++ b/R/nm-tables.R
@@ -89,9 +89,12 @@ nm_table_files <- function(.mod, .check_exists = TRUE) {
 
   # get file names from table statements and construct paths
   .f <- nmrec::select_records(.l, "table") %>%
-    purrr::map_chr(function(.x) nmrec::get_record_option(.x, "file")$value) %>%
-    str_replace("^\\.\\/", "") %>%
+    purrr::map(function(.x) nmrec::get_record_option(.x, "file")$value) %>%
+    unlist() %>% str_replace("^\\.\\/", "") %>%
     file.path(out_dir, .)
+
+  # Coerce to the same type of quote
+  .f <- sub("^'(.*)'$", "\\1", .f) %>% sub('^"(.*)"$', "\\1", .)
 
   if(rlang::is_empty(.f)){
     stop(glue("No table files were found in {.p}"))
--- a/R/nm-tables.R
+++ b/R/nm-tables.R
@@ -89,7 +89,7 @@ nm_table_files <- function(.mod, .check_exists = TRUE) {
 
   # get file names from table statements and construct paths
   .f <- nmrec::select_records(.l, "table") %>%
-    purrr::map_chr(~nmrec::get_record_option(.x, "file")$value) %>%
+    purrr::map_chr(function(.x) nmrec::get_record_option(.x, "file")$value) %>%
     str_replace("^\\.\\/", "") %>%
     file.path(out_dir, .)
 
--- a/R/nm-tables.R
+++ b/R/nm-tables.R
@@ -90,11 +90,13 @@ nm_table_files <- function(.mod, .check_exists = TRUE) {
   # get file names from table statements and construct paths
   .f <- nmrec::select_records(.l, "table") %>%
     purrr::map(function(.x) nmrec::get_record_option(.x, "file")$value) %>%
-    unlist() %>% str_replace("^\\.\\/", "") %>%
-    file.path(out_dir, .)
+    unlist() %>% str_replace("^\\.\\/", "")
 
-  # Coerce to the same type of quote
-  .f <- sub("^'(.*)'$", "\\1", .f) %>% sub('^"(.*)"$', "\\1", .)
+  # Unquote file paths
+  .f <- unquote_filename(.f)
+
+  # Absolute paths
+  .f <- .f %>% file.path(out_dir, .)
 
   if(rlang::is_empty(.f)){
     stop(glue("No table files were found in {.p}"))
@@ -158,4 +160,17 @@ parse_ctl_to_list <- function(.path) {
   names(spec) <- labs
 
   return(spec)
+}
+
+#' Unquote file paths for inclusion within a NONMEM control file
+#'
+#' @param .files vector of file paths
+#'
+#' @keywords internal
+unquote_filename <- function(.files){
+  quoted <- grepl("^'.*'$", .files) | grepl('^".*"$', .files)
+  if (any(quoted)) {
+    .files[quoted] <- stringr::str_sub(.files[quoted], 2, -2)
+  }
+  return(.files)
 }
--- a/R/nm-tables.R
+++ b/R/nm-tables.R
@@ -84,14 +84,13 @@ nm_tables <- function(
 #' @export
 nm_table_files <- function(.mod, .check_exists = TRUE) {
   .p <- get_model_path(.mod)
-  .p <- get_model_path(.mod)
   .l <- nmrec::read_ctl(.p)
   out_dir <- get_output_dir(.mod, .check_exists = .check_exists)
 
   # get file names from table statements and construct paths
   .f <- nmrec::select_records(.l, "table") %>%
     purrr::map(~nmrec::get_record_option(.x, "file")$value) %>%
-    purrr::list_c() %>%
+    purrr::flatten_chr() %>%
     str_replace("^\\.\\/", "") %>%
     file.path(out_dir, .)
 
--- a/R/nm-tables.R
+++ b/R/nm-tables.R
@@ -84,14 +84,14 @@ nm_tables <- function(
 #' @export
 nm_table_files <- function(.mod, .check_exists = TRUE) {
   .p <- get_model_path(.mod)
-  .l <- parse_ctl_to_list(.p)
+  .p <- get_model_path(.mod)
+  .l <- nmrec::read_ctl(.p)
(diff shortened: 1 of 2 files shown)


R/nm-tables.R
```diff
@@ -84,14 +84,12 @@ nm_tables <- function(
...
+  .l <- nmrec::read_ctl(.p)
   out_dir <- get_output_dir(.mod, .check_exists = .check_exists)
 
   # get file names from table statements and construct paths
-  .f <- .l[grep("^TAB", names(.l))] %>%
-    map_chr(~paste(.x, collapse = " ")) %>%
-    str_extract("\\bFILE\\s*=\\s*([^ ]+)") %>%
-    str_replace("\\bFILE\\s*=\\s*", "") %>%
+  .f <- nmrec::select_records(.l, "table") %>%
+    purrr::map_chr(~nmrec::get_record_option(.x, "file")$value) %>%
```
kyleam, line 92: It's okay as is, and I know purrr's anonymous function syntax is used a lot in this code base, but using `function(...) ...` instead would allow `R CMD check` to flag/warn about issues inside the function.

kyleam, line 92: A table record doesn't necessarily have a `FILE` option, in which case `NULL` will be returned by this function. `map_chr` will signal an error in that case. To stick with `map_chr`, you could tack on an `... %||% NA_character_` at the end of the function body and then filter NAs out downstream, but it's probably cleaner to instead do

```R
purrr::map(...) %>%
  unlist()
```

R/nm-tables.R
```diff
@@ -84,14 +84,12 @@ nm_tables <- function(
...
     str_replace("^\\.\\/", "") %>%
     file.path(out_dir, .)
```
kyleam, line 95: This isn't an issue introduced by the changes proposed here, but I think it'd be okay to address with this PR (up to you): the file value may be quoted with single or double quotes. Could append something like this to end of pipeline:

```
> .f
[1] "foo"     "'bar'"   "\"baz\""
> sub("^'(.*)'$", "\\1", .f) %>% sub('^"(.*)"$', "\\1", .)
[1] "foo" "bar" "baz"
```

kyleam (changes requested):
My main comment (inline) is that the proposed change will error if every table record doesn't have a file option (which isn't required by NONMEM).

Also, it'd be good to test at least one case that wouldn't have worked before this switch (e.g., "FILE {filename}" rather than "FILE={filename}"). `tests/testthat/test-nm-tables.R` already has some pretty simple tests:

<https://github.com/metrumresearchgroup/bbr/blob/5973247ef6269fcc6d3a1b123fd8bb24c969d865/tests/testthat/test-nm-tables.R#L62-L76>

barrettk:
>My main comment (inline) is that the proposed change will error if every table record doesn't have a file option (which isn't required by NONMEM).

My main question here is: What is the expected result of `nm_table_files` in these cases? If no filename is specified, we currently wouldnt look for that table and by extension, join that data in `nm_join`. Is it expected that we ignore these tables if no filename is specified?

kyleam:
> Is it expected that we ignore these tables if no filename is specified?

My read of the NONMEM docs is that, if no filename is specified, the output is just printed in the NONMEM output.  So, yes, I'd say the only thing to do is ignore such table records.

seth127:
> the only thing to do is ignore such table records

This makes sense to me. It also made me think about what we do if there _aren't any_ table files. The default path looks like it errors out ([here](https://github.com/metrumresearchgroup/bbr/blob/dev/nmrec-nm_table_files/R/nm-tables.R#L97)). Passing `NULL` (or anything else that doesn't lead to actual files on disk) looks like it will error out down the line somewhere (either [here](https://github.com/metrumresearchgroup/bbr/blob/dev/nmrec-nm_table_files/R/nm-tables.R#L46) or when it's passed to `nm_file()` [here](https://github.com/metrumresearchgroup/bbr/blob/dev/nmrec-nm_table_files/R/nm-tables.R#L74)).

Either way, I think that makes sense to error out if no table files are found. In this case, there's no value in `nm_join()` and the caller could just use `nm_data()` directly.

As an aside: I was looking into it because I was considering what to do for this "no table files found" scenario in https://github.com/metrumresearchgroup/bbr.bayes/issues/96. That gets a bit more complicated (because of the simulations downstream), but that discussion doesn't belong here.

R/nm-tables.R
barrettk: fixed

barrettk: fixed

R/nm-tables.R
barrettk: added - tested that this worked well too

barrettk:
>Either way, I think that makes sense to error out if no table files are found. In this case, there's no value in nm_join() and the caller could just use nm_data() directly.

@seth127 are you suggesting we add support for this in this PR?

R/nm-tables.R
```diff
@@ -84,17 +84,18 @@ nm_tables <- function(
...
-    map_chr(~paste(.x, collapse = " ")) %>%
-    str_extract("\\bFILE\\s*=\\s*([^ ]+)") %>%
-    str_replace("\\bFILE\\s*=\\s*", "") %>%
-    str_replace("^\\.\\/", "") %>%
+  .f <- nmrec::select_records(.l, "table") %>%
+    purrr::map(function(.x) nmrec::get_record_option(.x, "file")$value) %>%
+    unlist() %>% str_replace("^\\.\\/", "") %>%
     file.path(out_dir, .)
 
+  # Coerce to the same type of quote
```
kyleam, line 96: I don't follow this comment.  The intent is to "unquote" the file name.

R/nm-tables.R
```diff
@@ -84,17 +84,18 @@ nm_tables <- function(
...
+  .f <- sub("^'(.*)'$", "\\1", .f) %>% sub('^"(.*)"$', "\\1", .)
```
kyleam, line 97: Sorry, I realize this was my suggestion, but this code will incorrectly strips a set of inner double quotes, while the expected quoting rules would be that only the outer layer is stripped.

I've confirmed that this is how NONMEM treats the quoting:

```
$ grep TAB 1.ctl
$TABLE NUM IPRED NPDE CWRES NOPRINT ONEHEADER FILE='"1.tab"'

$ ls --quoting-style=literal 1/*tab*
1/"1.tab"
```

How about something like this instead?

```R
quoted <- grepl("^'.*'$", .f) | grepl('^".*"$', .f)
if (any(quoted)) {
  .f[quoted] <- stringr::str_sub(.f[quoted], 2, -2)
}
```

You could of course pull that into something like an `unquote_filename`, if you find that clearer.

kyleam (changes requested):
Thanks for the update, including the additional tests.

Sorry, I overlooked something in my last review when I suggested the "unquote" code.  Other than that, this looks good to me.

R/nm-tables.R
barrettk: @kyleam in your first code snippet, is that the expected behavior, or an example of incorrect filename specifications?

Trying to get a better understanding of what is accepted. Is it A) no quotes or B) two sets of quotes (double for the inner)?

kyleam: > in your first code snippet, is that the expected behavior, or an example of incorrect filename specifications?

It's verification that NONMEM will only strip the outer layer of quotes (as expected for shell quoting).   Line 97 in the current code will incorrectly strip the inner layer of quotes.

> Trying to get a better understanding of what is accepted. Is it A) no quotes or B) two sets of quotes (double for the inner)?

Sorry, I don't follow.  Accepted by what?

Go into a bash shell and do this:

```
$ touch '"foo"'
```

That creates a file `"foo"` (double quotes are part of the actual name).  Likewise, `touch "'foo'"` would create a file `'foo'`.  NONMEM follows that expected behavior of only removing a single layer of quoting.

For an R return value, we want to return the name of the actual file, without any quoting.  The current code on line 97 doesn't because it will strip `'"foo"'` to `foo`, when the actual file name is `"foo"`.

barrettk: Thanks for the added clarity. I went with creating a new function like you suggested, though I opted for more of an integration test (as opposed to testing `unquote_filename` specifically)

kyleam (approved):
Thanks for the update, and sorry again for initially suggesting an incorrect approach.

tests/testthat/test-nm-tables.R
```diff
@@ -75,3 +75,40 @@ test_that("nm_table_files() errors with non-existent file [BBR-NMT-003]", {
...
+test_that("nm_table_files() works if no filename is present [BBR-NMT-003]", {
+  # Table records with no specified filename will be filtered out
+  perturb_file(get_model_path(MOD1), txt = "\n$TABLE NUM CL KA\n")
+
+  res <- nm_table_files(MOD1, .check_exists = FALSE)
+  expect_true(all(fs::is_absolute_path(res)))
+  expect_equal(basename(res), MOD1_TABLE_FILES)
+})
+
+test_that("nm_table_files() works with quoted file paths [BBR-NMT-003]", {
```
kyleam, line 97: Great, thanks for adding this test.

kyleam:
> FYI we need to run `devtools::document()` in the parent branch before merging.

Okay.  So it doesn't get forgotten, I'd say it's fine for you to do that now and push directly to dev-nmrec.

PR #604, opened 2023-07-31
copy_control_stream: use nmrec instead of regex

barrettk:
addresses `copy_control_stream()` part of https://github.com/metrumresearchgroup/bbr/issues/599

### DIFF
--- a/R/copy-model-from.R
+++ b/R/copy-model-from.R
@@ -222,17 +222,16 @@ copy_control_stream <- function(.parent_model_path, .new_model_path, .overwrite,
   if (.update_model_file) {
 
     # read parent control stream
-    mod_str <- .parent_model_path %>% read_file()
+    mod_str <- nmrec::read_ctl(.parent_model_path)
 
     # replace the $PROBLEM line(s)
     new_mod_id <- get_model_id(.new_model_path)
-    mod_str <- str_replace(mod_str,
-                           "\\$PROB(.|\n)*?\\$",
-                           as.character(glue("$PROBLEM From bbr: see {new_mod_id}.yaml for details\n\n$")))
-    glue("")
+    prob_rec <- nmrec::select_records(mod_str, "prob")[[1]]
+    prob_str <- nmrec::get_record_option(prob_rec, "text")
+    prob_str$value <- as.character(glue("From bbr: see {new_mod_id}.yaml for details"))
 
-    # read parent control stream
-    write_file(mod_str, .new_model_path)
+    # write parent control stream
+    nmrec::write_ctl(mod_str, .new_model_path)
   } else {
     fs::file_copy(.parent_model_path, .new_model_path)
   }
--- a/R/copy-model-from.R
+++ b/R/copy-model-from.R
@@ -60,15 +60,15 @@
 #' @seealso [new_model()], [update_model_id()]
 #' @export
 copy_model_from <- function(
-  .parent_mod,
-  .new_model = NULL,
-  .description = NULL,
-  .based_on_additional = NULL,
-  .add_tags = NULL,
-  .star = NULL,
-  .inherit_tags = FALSE,
-  .update_model_file = TRUE,
-  .overwrite = FALSE
+    .parent_mod,
+    .new_model = NULL,
+    .description = NULL,
+    .based_on_additional = NULL,
+    .add_tags = NULL,
+    .star = NULL,
+    .inherit_tags = FALSE,
+    .update_model_file = TRUE,
+    .overwrite = FALSE
 ) {
   UseMethod("copy_model_from")
 }
@@ -76,15 +76,15 @@ copy_model_from <- function(
 #' @describeIn copy_model_from `.parent_mod` takes a `bbi_nonmem_model` object to use as a basis for the copy.
 #' @export
 copy_model_from.bbi_nonmem_model <- function(
-  .parent_mod,
-  .new_model = NULL,
-  .description = NULL,
-  .based_on_additional = NULL,
-  .add_tags = NULL,
-  .star = NULL,
-  .inherit_tags = FALSE,
-  .update_model_file = TRUE,
-  .overwrite = FALSE
+    .parent_mod,
+    .new_model = NULL,
+    .description = NULL,
+    .based_on_additional = NULL,
+    .add_tags = NULL,
+    .star = NULL,
+    .inherit_tags = FALSE,
+    .update_model_file = TRUE,
+    .overwrite = FALSE
 ) {
 
   .new_model <- build_new_model_path(.parent_mod, .new_model)
@@ -143,16 +143,16 @@ copy_model_from.bbi_nonmem_model <- function(
 #' @return S3 object with the same class as `.parent_mod`
 #' @keywords internal
 copy_model_from_impl <- function(
-  .parent_mod,
-  .new_model,
-  .description = NULL,
-  .based_on_additional = NULL,
-  .add_tags = NULL,
-  .star = NULL,
-  .inherit_tags = FALSE,
-  .update_model_file = TRUE,
-  .overwrite = FALSE,
-  setup_fn = NULL
+    .parent_mod,
+    .new_model,
+    .description = NULL,
+    .based_on_additional = NULL,
+    .add_tags = NULL,
+    .star = NULL,
+    .inherit_tags = FALSE,
+    .update_model_file = TRUE,
+    .overwrite = FALSE,
+    setup_fn = NULL
 ) {
   check_for_existing_model(.new_model, .overwrite)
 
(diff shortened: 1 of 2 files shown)


R/copy-model-from.R
```diff
@@ -222,17 +222,16 @@ copy_control_stream <- function(.parent_model_path, .new_model_path, .overwrite,
...
-    mod_str <- .parent_model_path %>% read_file()
+    mod_str <- nmrec::read_ctl(.parent_model_path)
 
     # replace the $PROBLEM line(s)
     new_mod_id <- get_model_id(.new_model_path)
-    mod_str <- str_replace(mod_str,
-                           "\\$PROB(.|\n)*?\\$",
-                           as.character(glue("$PROBLEM From bbr: see {new_mod_id}.yaml for details\n\n$")))
-    glue("")
+    prob_rec <- nmrec::select_records(mod_str, "prob")[[1]]
```
kyleam, line 229: Control streams must have a problem record, and nmrec requires there only be a single record. So, if there weren't one record, the `nmrec::read_ctl` call a few lines up would fail.

Still, it'd be nice if this was a little more defensive and abort if things looked unexpected. For example, something like this (untested):

```diff
diff --git a/R/copy-model-from.R b/R/copy-model-from.R
index 584ca3b3..b63100a5 100644
--- a/R/copy-model-from.R
+++ b/R/copy-model-from.R
@@ -226,8 +226,11 @@ copy_control_stream <- function(.parent_model_path, .new_model_path, .overwrite,

     # replace the $PROBLEM line(s)
     new_mod_id <- get_model_id(.new_model_path)
-    prob_rec <- nmrec::select_records(mod_str, "prob")[[1]]
-    prob_str <- nmrec::get_record_option(prob_rec, "text")
+    prob_recs <- nmrec::select_records(mod_str, "prob")
+    if (length(prob_recs) != 1) {
+      dev_error("select_records should always return a single $PROBLEM record")
+    }
+    prob_str <- nmrec::get_record_option(prob_recs[[1]], "text")
     prob_str$value <- as.character(glue("From bbr: see {new_mod_id}.yaml for details"))

     # write parent control stream
```

R/copy-model-from.R
```diff
@@ -222,17 +222,16 @@ copy_control_stream <- function(.parent_model_path, .new_model_path, .overwrite,
...
+    prob_str <- nmrec::get_record_option(prob_rec, "text")
```
kyleam, line 230: As we discussed a bit offline, there's a wrinkle here in that NONMEM will accept `$PROB{newline}` as a valid record. [61c85f7](https://github.com/metrumresearchgroup/nmrec/commit/61c85f7a0996df46b0240c59a59ec841d1b26864) (define parser for problem records, 2023-06-30) explains the situation a bit.

In this case, there won't be a `text` option. At the very least, this spot should detect a `NULL` return from `get_record_option`, stop trying to update it, and warn the user.

A more involved fix could be something like this (again, untested):

```diff
diff --git a/R/copy-model-from.R b/R/copy-model-from.R
index 584ca3b3..266fe5b9 100644
--- a/R/copy-model-from.R
+++ b/R/copy-model-from.R
@@ -227,8 +227,21 @@ copy_control_stream <- function(.parent_model_path, .new_model_path, .overwrite,
     # replace the $PROBLEM line(s)
     new_mod_id <- get_model_id(.new_model_path)
     prob_rec <- nmrec::select_records(mod_str, "prob")[[1]]
+    new_text <- as.character(glue("From bbr: see {new_mod_id}.yaml for details"))
+
     prob_str <- nmrec::get_record_option(prob_rec, "text")
-    prob_str$value <- as.character(glue("From bbr: see {new_mod_id}.yaml for details"))
+    if (is.null(prob_str)) {
+      # NONMEM permits a bare "$PROB{newline}", in which case there won't be a
+      # text option.
+      prob_rec$parse()
+      idx <- purrr::detect_index(
+        prob_rec$values,
+        function(x) inherits(x, "nmrec_option_record_name")
+      )
+      prob_rec$values <- append(prob_rec$values, paste0(" ", new_text), idx)
+    } else {
+      prob_str$value <- new_text
+    }

     # write parent control stream
     nmrec::write_ctl(mod_str, .new_model_path)
```

(That could be a bit more careful and signal a dev error if `idx == 0`.)

tests/testthat/test-copy-model-from.R
```diff
@@ -153,7 +153,8 @@ test_that("copy_from_model .overwrite=FALSE works [BBR-CMF-005]", {
 })
 
 test_that("copy_model_from() supports `.new_model` containing a period [BBR-CMF-006]", {
-  temp_mod_path <- create_temp_model()
+  mod_content <- readLines(ctl_ext(MOD1_PATH)) %>% paste(collapse = "\n")
+  temp_mod_path <- create_temp_model(mod_content = mod_content)
```
kyleam, line 157: The need for these test adjustments makes sense.

The test above already covers the main `.update_model_file = TRUE` case, but it'd be nice to have a test case that covered something that would have failed before this transition (e.g., a lower case `$prob ...`). It'd also be good to test the bare `$PROB{newline}` case mentioned above.

kyleam (changes requested):
Thanks @barrettk. Looking good overall. Left some comments mostly focusing on edge cases.

R/copy-model-from.R
```diff
@@ -222,17 +222,16 @@ copy_control_stream <- function(.parent_model_path, .new_model_path, .overwrite,
   if (.update_model_file) {
 
     # read parent control stream
-    mod_str <- .parent_model_path %>% read_file()
+    mod_str <- nmrec::read_ctl(.parent_model_path)
 
     # replace the $PROBLEM line(s)
```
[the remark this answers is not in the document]
kyleam: > Seems like it would be preferable to handle this in `bbi` [...]

What, in particular, do you have in mind?  

If you're saying that `bbi` should support multiple problems, that might be the ideal state, but I don't see that happening anytime soon.  And even if bbi/bbr did, nmrec would still need to be updated to handle multiple problems.

Or are you saying that bbi should give an informative error (for all relevant entry points: `run`, `summary`, ...)?  Even if that's the case, I don't think that's a guarantee that we don't hit into it here (it's a hand-edited file, and IIRC copying works fine with un-executed models).

So, I think the question is just, given that we think multiple problems aren't used in the wild with bbi/bbr, 1) are we okay with just letting an informative error bubble up from nmrec? or 2) do we want to catch it and warn that `.update_model_file = TRUE ` isn't in effect, and skip the "update model id" logic?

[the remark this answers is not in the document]
barrettk: > Or are you saying that bbi should give an informative error (for all relevant entry points: run, summary, ...)? Even if that's the case, I don't think that's a guarantee that we don't hit into it here (it's a hand-edited file, and IIRC copying works fine with un-executed models).

Yeah I was thinking more along these lines. But yeah fair point in that we could still run into the error here regardless of `bbi`'s handling. I did end up adding that catch in the latest commit though.

tests/testthat/test-copy-model-from.R
barrettk: oh just noticed the lower case `$prob` test case. Will add that now before you review

barrettk: Ok added that case in [789b9fe](https://github.com/metrumresearchgroup/bbr/pull/604/commits/789b9fefbf06b7a93987df6e48f67478f5f9ffea)

R/copy-model-from.R
```diff
@@ -60,31 +60,31 @@
...
-  .overwrite = FALSE
+    .parent_mod,
+    .new_model = NULL,
+    .description = NULL,
+    .based_on_additional = NULL,
+    .add_tags = NULL,
+    .star = NULL,
+    .inherit_tags = FALSE,
+    .update_model_file = TRUE,
+    .overwrite = FALSE
```
barrettk, line 71: These changes (and similar ones below) happened because I auto-indented

R/copy-model-from.R
```diff
@@ -222,17 +222,42 @@ copy_control_stream <- function(.parent_model_path, .new_model_path, .overwrite,
...
-                           "\\$PROB(.|\n)*?\\$",
-                           as.character(glue("$PROBLEM From bbr: see {new_mod_id}.yaml for details\n\n$")))
-    glue("")
+    mod_str <- tryCatch(nmrec::read_ctl(.parent_model_path), error = identity)
+
+    if(inherits(mod_str, "error")){
+      warning(
+        glue::glue(".update_model_file = TRUE was not performed. Reason: {mod_str$message}")
+      )
+      fs::file_copy(.parent_model_path, .new_model_path)
```
barrettk, line 231: I thought about adding another layer here...and checking for a specific error message (containing `$PROBLEM`), but opted not to. Let me know if doing so would make more sense though

tests/testthat/test-copy-model-from.R
```diff
@@ -213,3 +213,53 @@ test_that("copy_model_from(.new_model=NULL) errors when no models are valid inte
...
+
+  # Test lower case ($prob)
+  mod_content <- readLines(ctl_ext(MOD1_PATH))
+  mod_content[1] <- gsub("PROBLEM", "problem", mod_content[1])
+  mod_content <- mod_content %>% paste(collapse = "\n")
+  temp_mod_path <- create_temp_model(mod_content = mod_content)
+  temp_mod <- read_model(temp_mod_path)
+
+  new_mod_path <- "lowercase"
+  new_mod <- copy_model_from(temp_mod, new_mod_path, .overwrite = TRUE)
```
barrettk, line 234: I added `.overwrite = TRUE` here because subsequent runs would fail due to the existence of previously copied models. This shouldnt be needed, and I think this indicates a flaw in the `cleanup` function, but I couldn't narrow it down and figured we could address that in a separate PR if need be

R/copy-model-from.R
kyleam: What you went with makes sense to me.

tests/testthat/test-copy-model-from.R
kyleam: Okay, thanks for noting.

tests/testthat/test-copy-model-from.R
```diff
@@ -213,3 +213,53 @@ test_that("copy_model_from(.new_model=NULL) errors when no models are valid inte
     new_mod2 <- copy_model_from(new_mod1)
   }, regexp = "no models.+integer")
 })
+
+test_that("copy_model_from works with various PROBLEM declarations [BBR-CMF-007]", {
```
kyleam, line 217: Thanks for the extensive tests.

(I know it's just copied from above, but might as well avoid adding new test IDs at this point.)

kyleam (approved):
`copy_control_stream` and test updates look good to me.

Issue #582, opened 2023-02-24
Creation of new model with inital parameter values from final parameter values of prior model


oskarrgh:
Missing function for creating new model with initial theta/omega/sigma parameter values taken from final estimates of a prior model. Could perhaps be either stand-alone function or addition to copy_model_from function. Think this would add great value!

seth127:
Thanks Oskar. We've heard this from a few other users as well. We are definitely considering adding this feature. 

One of the primary concerns with this in the past has been how reliably we can parse the parameter blocks from the control stream, given the flexibility that NONMEM allows for defining them. I remember that OMEGA blocks in particular were a concern. @kylebaron do you have an input here? Is there anything you've done in `mrgsolve` or elsewhere that we could piggy-back off of, in terms of parsing parameter blocks from a NONMEM control stream?

timwaterhouse:
I agree that this would be a good addition. An alternative to trying to parse the control stream could involve using an `.msf` file somehow. I'm not sure how tricky that's going to be to feed in info from one model to another, though.

seth127:
This has come along way since the last comment on this issue. I think we're pretty close to having it implemented, with the bulk of the new code actually living in `nmrec` (https://github.com/metrumresearchgroup/nmrec/pull/18/ , currently being reviewed).

@oskarrgh I wanted to get some user review at this point, if possible. Would you (and/or anyone else) be willing to take a look at the [test cases in the `nmrec` PR](https://github.com/metrumresearchgroup/nmrec/blob/4fba4053ca02a59ecd1e78495d2b5f5eac19d2d5/tests/testthat/test-set-param.R) and verify that the replacement is happening the way that you would expect it to? Any comments can be left on https://github.com/metrumresearchgroup/nmrec/pull/18/ .

## Some explanation on the test cases

**These are testing internal helper functions** which a `bbr` user would _not_ interact with directly. Those functions are defined and documented in [`nmrec/R/set-param.R`](https://github.com/metrumresearchgroup/nmrec/blob/4fba4053ca02a59ecd1e78495d2b5f5eac19d2d5/R/set-param.R), if you are interested in looking at them.

Each test case is set up like so:

```r

list(
  lines = "$omega 1 2 3",
  values = c(
    4,
    0, 5,
    0, 0, 6
  ),
  want = "$omega 4 5 6"
)
```

* `lines` is a snippet from an "original" NONMEM control stream
* `values` are the _final estimates_ from a parent model (more on this format below)
[...]

PR #606, opened 2023-09-07
Add nmrec dependency

seth127:
_Work in progress... adds nmrec as a dependency and refactors some internal functions to use it, in place of legacy code that primarily relied on regex._


seth127:
@kyleam do you have an opinion about whether it would be easier/smoother to merge this and make a bbr release before or after nmrec gets added to MPN?

In other words:
- nmrec will be on the Sept 2023 snapshot
- should we make a release with this stuff and put it on the Sept 2023 snapshot, or is it easier to wait until after that snapshot?

kyleam:
> is it easier to wait until after that snapshot?

Unless you think there's any urgency for this rework to land on MPN (I'm not aware of any), I'd prefer to wait until nmrec makes it to snapshot.  We've got a good number of moving parts already with the snapshot and mrg releases, and waiting would let us finalize things (e.g., the drone config) without requiring post-release followups.

seth127:
Ok that makes sense to me @kyleam . Let's plan to revisit this in early October, once that MPN snapshot is out.

kyleam:
The snapshot is out; I've adjusted the Drone config to drop the custom nmrec installation in all spots aside from the oldest build, which still requires it (2e1edb91).

https://github-drone.metrumrg.com/metrumresearchgroup/bbr/3549

seth127:
@kyleam @barrettk what's the status of this branch at this point? I know it was serving as a dev branch for awhile... are there still pending PRs coming into here, or should we go ahead and merge it?

Related, @barrettk is actively working on #582 , so we could until that is done to merge all of this. That will likely be a little while though.

I'm fine with whatever y'all want to do, I just wanted to make sure this wasn't just loitering at this point. Thanks.

kyleam:
@seth127, once nmrec hit MPN, I pushed an update for the CI (2e1edb9), so it's ready to merge in my view. We no longer need a branch to merge nmrec PRs now that that nmrec is on MPN.

(Note that there's the unmerged gh-605, but [my opinion](https://github.com/metrumresearchgroup/bbr/pull/605#issuecomment-1660822495) is that that can be closed without merge.)

> Related, @barrettk is actively working on #582 [...]

I don't think that we should wait for that to come in.

> I'm fine with whatever y'all want to do, I just wanted to make sure this wasn't just loitering at this point.

As the opener of the PR, I was expecting you to switch it to "ready for review" and guide it to completion (probably just requesting a review and hitting merge at this point, though things may pop up).  Would you like someone to take over that?

.drone.yml
```diff
@@ -100,6 +100,9 @@ steps:
   image: 906087756158.dkr.ecr.us-east-1.amazonaws.com/r-dev-ci-mpn-4.0:2020-06-08
   commands:
   - R -s -e 'devtools::install_deps(upgrade = '"'"'never'"'"')'
+  # nmrec isn't available until the 2023-09-19 snapshot.
+  - git clone --depth 1 https://github.com/metrumresearchgroup/nmrec.git /tmp/nmrec
+  - R -s -e 'devtools::install("/tmp/nmrec", upgrade = "never")'
```
barrettk, line 105: I think we want to change these lines before merging FYI

seth127:
Thanks for the updates.

> As the opener of the PR, I was expecting you to switch it to "ready for review" and guide it to completion (probably just requesting a review and hitting merge at this point, though things may pop up). Would you like someone to take over that?

Yes, fair point. Although I would be happy if you don't mind taking it over @kyleam . I think you've reviewed most (all?) of the PRs that came _into_ this branch, so hopefully it will be pretty quick look this one over. Let me know if I'm missing any nuance there and we can discuss a different plan, if needed.

.drone.yml
kyleam: > I think we want to change these lines before merging FYI

What is your suggested change?

kyleam: As the comment on line 103 says, nmrec isn't available on this image's (`r-dev-ci-mpn-4.0:2020-06-08`) snapshot, so we need to keep the custom install.  We could update the custom command to install nmrec from the 2023-09-19 snapshot instead of the Git repo (or equivalently just make this use the v0.2.0 tag from the nmrec repo).  

I could see the argument that pinning the nmrec version is more inline with the "test oldest" goal of this build.  Is that the reason for your suggestion?

kyleam: > I could see the argument that pinning the nmrec version is more inline with the "test oldest" goal of this build

a92c9bf4 pins the nmrec version.

kyleam (approved):
The core changes of this series have been reviewed elsewhere: 600, 603, 604

Here are the non-merge commits directly to this branch:

 * 5a1b11b7 (add nmrec dependency to drone)
 * 1734b8d6 (add nmrec to description)
 * 8f8e40de (document unquote_filename)
 * 2e1edb91 (ci: drop now unneeded nmrec installations)
 * a92c9bf4 (ci: pin oldest pipeline to nmrec 0.2.0)

Those are non-R changes: tweak the Drone configuration, update `DESCRIPTION` to include nmrec as dependency, and regenerate missing docs.  Those all look fine to me.

---

At least on my end, GitHub isn't showing the Drone builds as completed.  Here they are (both green):

 * https://github-drone.metrumrg.com/metrumresearchgroup/bbr/3555
 * https://github-drone.metrumrg.com/metrumresearchgroup/bbr/3556

kyleam:
@barrettk If all looks good to you, please do the honors of merging to main.

.drone.yml
barrettk: >I could see the argument that pinning the nmrec version is more inline with the "test oldest" goal of this build. Is that the reason for your suggestion?

I thought about that too, but mainly it was just to reduce some of the variability in what version of `nmrec` we pull in drone. In the event `nmrec` and `bbr` were undergoing larger refactors, I figured it would be good to have a more stable version of `nmrec` rather than whatever was on `main` at the time. Pointing to a specific version definitely takes care of that and IMO is actually better than pointing to an MPN snapshot anyways (so we can track `nmrec` and `bbi` versions similarly).

barrettk:
>At least on my end, GitHub isn't showing the Drone builds as completed. Here they are (both green)

Yeah that is the same on my end as well

.drone.yml
kyleam: > mainly it was just to reduce some of the variability in what version of `nmrec` we pull in drone.

Having the nmrec version be different across the pipelines is fine, and in many cases desirable because it increases what you're testing.  On average nmrec's main should be more stable than a release because it has fixes, but in the event that it does break bbr compatibility, we want to know because it shouldn't.

In any case, for the oldest build using an old nmrec for consistency as a92c9bf works for me, so we're set for the purposes of this PR.

kyleam:
The good old off/on approach didn't work to get the status checks to show up.  (Given it's not urgent to get this in, I'm okay holding this up to see if we can resolve it; however, I'm also fine with admin-bypassing this and waiting to see whether it comes up again.)

seth127:
Per [this comment above](https://github.com/metrumresearchgroup/bbr/pull/606#pullrequestreview-1668573637) everything is good, we just have a glitch in the Drone build showing up. I am merging with my admin privileges to avoid having to dig any deeper into the Drone issue. We'll dig deeper on it if this keeps happening on other PRs.

PR #610, opened 2023-10-26
Add model_tree() function

barrettk:
- Will operate similar to `run_log` and other related log functions.
 - Defaults to an interactive HTML plot and optionally a static plot
 - Control tooltip settings and optionally append model_summary output (defaults to including this)


seth127:
After discussing today, @barrettk is going to open a new branch that cleans this up and focuses on the `collapsibleTree` approach.

Some updates needed:

* Streamline the static option to print to PNG
* Implement S3 methods, similar to other places in `bbr` (e.g. [`get_based_on()`](https://github.com/metrumresearchgroup/bbr/blob/main/R/get-based-on.R))
  * The `character` method 
    * expects a (scalar) directory path
    will need additional args to pass through to `run_log()`
  * The `bbi_log_df` method
    * need to check that the necessary columns are in the passed df (`based_on`, `absolute_model_path`, everything in `.info`)
    * Consider whether to have the log method be on `bbi_log_df` or `bbi_run_log_df`
  * check that the error looks reasonable if we _don't_ add a `default` method 
    * if not, should error and tell the user what the acceptable inputs are (dir path or log df)

barrettk:
New PR: https://github.com/metrumresearchgroup/bbr/pull/673

Issue #611, opened 2023-10-30
update_model_id mishandles based_on field


kyleam:
The goal of `update_model_id` is to replace parent model ID with the new model ID.  It gets the parent ID directly from the `based_on` field:

https://github.com/metrumresearchgroup/bbr/blob/727b65b5cfba817a49d29d290ccc5c47ead050e8/R/copy-model-helpers.R#L68

`based_on` is the relative path from the YAML, which means it can be something like `../1100`.  `update_model_id` mishandles this in two ways:

 * it doesn't account for `based_on` being a path that only corresponds to the model ID when the parent is in the same directory as the child model.

   Two potential fixes:

   * take the `basename` of the `based_on` field
   * use `get_model_id` on the `based_on` value (`get_model_id(get_based_on(...))`)

   I like the second approach because it more clearly communicates the desired result, and it lets the method worry about the implementation details (though it's unlikely we'll ever have a model type that breaks the `basename` -> `model ID` mapping).

 * it uses the value as a regular expression pattern without protecting it as literal value (e.g., with `\\Q...\\E`)

<details>
<summary>example of bad update</summary>

Here's an example of a bad update:

```
$ tail -n3 inst/model/nonmem/bayes/{1100/1100-1.yaml,1100.ctl,1100/1100-1.ctl}
==> inst/model/nonmem/bayes/1100/1100-1.yaml <==
description: Chain 1
based_on:
- ../1100

==> inst/model/nonmem/bayes/1100.ctl <==
$EST METHOD=BAYES CTYPE=0 SEED=1 NBURN=10 NITER=50 PRINT=10 MSFO=./1100.msf RANMETHOD=P PARAFPRINT=10000 BAYES_PHI_STORE=1

$TABLE NUM CL V2 Q V3 KA ETAS(1:LAST) EPRED IPRED NPDE EWRES NOPRINT ONEHEADER FILE=1100.tab RANMETHOD=P

==> inst/model/nonmem/bayes/1100/1100-1.ctl <==
$EST METHOD=BAYES CTYPE=0 SEED=1 NBURN=10 NITER=50 PRINT=10 MSFO1100-1.msf RANMETHOD=P PARAFPRINT=10000 BAYES_PHI_STORE=1

$TABLE NUM CL V2 Q V3 KA ETAS(1:LAST) EPRED IPRED NPDE EWRES NOPRINT ONEHEADER FILE=1100.tab RANMETHOD=P
```

[...]

PR #614, opened 2023-10-31
update_model_id: work with child models in sub-directories

barrettk:
Previously, `update_model_id` would replace the id with the `based_on` field, which was actually a relative path rather than a model id. In other words, this function did not support child models created in sub-directories (or otherwise a different directory than the parent model). This PR addresses that by capturing the id of the parent model instead of the path. See referenced issue for more details.

closes https://github.com/metrumresearchgroup/bbr/issues/611

### DIFF
--- a/R/copy-model-helpers.R
+++ b/R/copy-model-helpers.R
@@ -65,11 +65,13 @@ update_model_id <- function(
   check_yaml_in_sync(.mod)
   mod_id <- get_model_id(.mod)
   modelfile <- get_model_path(.mod)
-  based_on <- .mod$based_on[1]
+  based_on <- get_based_on(.mod)
   if (is.null(based_on)) {
     stop(glue("Cannot call update_model_id() because .mod$based_on is empty for model {mod_id}"))
+  }else{
+    based_on_id <- get_model_id(based_on)
   }
-  message(glue("replacing {based_on} with {mod_id} in {modelfile}"))
+  message(glue("replacing {based_on_id} with {mod_id} in {modelfile}"))
 
   ## construct suffixes regex string
   assert_character(.suffixes)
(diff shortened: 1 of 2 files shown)


kyleam (changes requested):
Thanks.  The fix to make it use the model ID looks good.  Could you also add a commit that prevents regex characters in the parent model ID from triggering unexpected matches (second issue described at gh-614)?

Here's a test to demonstrate the issue:

```diff
diff --git a/tests/testthat/test-copy-model-helpers.R b/tests/testthat/test-copy-model-helpers.R
index c364fc7d..5721ea98 100644
--- a/tests/testthat/test-copy-model-helpers.R
+++ b/tests/testthat/test-copy-model-helpers.R
@@ -135,3 +135,19 @@ test_that("update_model_id() works with models in different directories", {
     paste0(get_model_id(new_mod), c(DEFAULT_SUFFIXES, new_suff))
   )
 })
+
+test_that("update_model_id() protects regex characts in parent model ID", {
+  withr::with_tempdir({
+    mod1a <- copy_model_from(MOD1, file.path(getwd(), "1+"))
+    mod2 <- copy_model_from(mod1a, 2)
+
+    mod2_path <- get_model_path(mod2)
+    readr::write_lines(c("1+.tab", "11.tab"), mod2_path)
+
+    update_model_id(mod2)
+    expect_identical(
+      readr::read_lines(mod2_path),
+      c("2.tab", "11.tab")
+    )
+  })
+})
```

Note that I'm avoiding "." as the regex character because that triggers a failure due to `safe_based_on` stripping the right-side as the extension.  At first glance that's a bug because it's valid to have a period in the model name (`copy_model_from` even has "100.1" as an example model name), but this PR doesn't need to go down that rabbit hole.

kyleam (approved):
Updated code looks good to me, thanks.

> add support for regex in parent model ID.

I think this commit message is likely to cause confusion if anyone stumbles on this commit in future troubleshooting.  You're not adding support for the model ID being treated as a regular expression; you're preventing it from being treated as one.

Not a blocker given commit messages aren't typically reviewed in this project, but I would be in favor of you amending the commit to reword.

PR #621, opened 2023-11-20
tests: add missing skips for SUM1-based tests

kyleam:
When not running on Metworx or Drone, setup-workflow-ref.R doesn't define SUM1.  Most spots that use SUM1 call
skip_if_not_drone_or_metworx() to avoid an "object 'SUM1' not found" error.  Fix the two spots that don't.


kyleam:
> When not running on Metworx

.. where "running on Metworx" is detected by `METWORX_VERSION` being set.  I hit into this in a dev Metworx environment that doesn't yet set this environment variable.

seth127 (approved):

PR #626, opened 2023-12-11
get_param: fix data length for constructed matrix

kyleam:
When creating the matrix from a vector and labels, format_matrix() calculates the data size as 2 * N, where N is the maximum label index (e.g., 2 * 3 for a 3x3 matrix).  The correct size is N * N.

Fortunately that error doesn't functionally matter because matrix's ncol and nrow arguments, not the input data length, control the final shape.

R 4.2 or later flags the issue in cases where 2 * N isn't a multiple of N * N.  Several spots in test-get-param.R now trigger a warning: "data length differs from size of matrix".

Fix the issue by dropping the rep() call entirely and passing a scalar.  matrix() will repeat that as many times as needed.

### DIFF
--- a/R/get-param.R
+++ b/R/get-param.R
@@ -127,7 +127,7 @@ format_matrix <- function(.values, .labels, .type = c("OMEGA", "SIGMA")){
   .cols <- map(matrix_pos, 2) %>% unlist() %>% as.numeric()
 
   # Assign values to corresponding location
-  .mat <- matrix(rep(0,max(.rows, .cols)*2), nrow = max(.rows), ncol = max(.cols))
+  .mat <- matrix(0, nrow = max(.rows), ncol = max(.cols))
   for(i in seq_along(.rows)){
     .mat[.rows[i],.cols[i]] <- .values[i]
     .mat[.cols[i],.rows[i]] <- .values[i]


barrettk (approved):
Good catch - code reads better now too/makes more sense. 

As an aside, you shouldnt be able to have a 2 x 3 matrix for an OMEGA matrix from my understanding though (not sure if that was just an example). Shouldnt the _full_ output matrix always be symmetrical with `nrow==ncol`?  I assume you'd still see that warning for a 3 x 3 matrix since 2 x N is not a multiple of 9, but just want to make sure I fully understand your commit message

barrettk:
@kyleam No I see what youre saying by that - I agree that it's wrong. I honestly recall thinking about this when I implemented it (and knew it should be N x N in terms of dimension). I wish I could remember why I changed it to 2 x N, but now I am not able to make sense of why I did it that way. Would bet I saw some weird behavior in the console or something, but dont think it really matters at this point. Regardless, a single static value makes the most sense since the row and column specification ultimately puts it in the correct format/dimension.

But no I think your commit message is fine and still approve, I just wanted to make sure a 2 x 3 matrix wasnt a possible output from `get_omega()`/`get_param()`.

kyleam:
@barrettk Got it.  Thanks for the review.

(Obviously this is just a focused fix, but I wouldn't be opposed to a rewrite of `format_matrix` that makes it bit clear that this should always be a square matrix.  For example, get max index for rows and columns and assert they're the same.)

Issue #607, opened 2023-09-12
bbi_init: clarify .nonmem_dir


kyleam:
cc: @seth127 

The location of `.nonmem_dir` is a frequent source of confusion.  `bbi_init` (and `bbi init` underneath) expect to be pointed to a directory that has multiple installations.  Using Metworx as an example, `.nonmem_dir` should be `/opt/NONMEM`:

```
$ ls -1 /opt/NONMEM/
nm73gf
nm73gf_nmfe
nm74gf
nm74gf_nmfe
nm75
```

Underneath, each of those is a NONMEM installation:
```
$ ls -d /opt/NONMEM/nm75/run
/opt/NONMEM/nm75/run
```

Here's what the docs currently say:

https://github.com/metrumresearchgroup/bbr/blob/1fd58de8f4eddc8c9d2cc02c947e85b47a5aaf16/R/bbr.R#L326-L326

It'd be good to rephrase/extend that to make it clearer that `/opt/NONMEM` not `/opt/NONMEM/nm75` is the expected input.

seth127:
@kyleam @barrettk we just had another external (to Metrum) user reach out with this confusion. Let's put this on the list to knock out before the next release too. Obviously it would be preferable to point them to the docs instead of pointing them to an open issue about how we intend to update the docs...

barrettk:
@seth127 @kyleam I can take this on today/tomorrow if you guys want. Would like to take a look at the support ticket though just to get a better sense of how we can make the documentation as clear as possible.

kyleam:
Great, thanks @barrettk.  

I don't know of a support ticket offhand.  (This came up a bit in gh-596, but there were several other things going on there, so I'm not sure that helps to look through.)   Perhaps @seth127 can point you to a support ticket or some other issue, but I think the main confusion boils down to thinking you should pass the NONMEM installation directory rather than the parent directory that contains that installation (which I think is understandable given bbi_init's description for `.nonmem_dir`).

Fwiw bbi's documentation is this "a directory in which to look for NonMem Installations"

https://github.com/metrumresearchgroup/bbi/blob/835953efed10fff2d7024ac1c3dd88949b268c05/cmd/init.go#L64

Issue #596, opened 2023-05-15
Error assigning nm version 


ldbro0:
Hi 

I'm trying to get `bbr` and `bbi` working and I'm getting the following error 

```
> bbi_init(.dir = "nonmem/study_id/poppk", .nonmem_dir = "/gstore/scratch/u/id_12345/7.5.0-foss-2017a", .nonmem_version = "nmfe75")

Error in bbi_init(.dir = "nonmem/study_id/poppk", .nonmem_dir = "/gstore/scratch/u/id_12345/7.5.0-foss-2017a",  : 
  Must specify a valid `.nonmem_version` for bbi_init(). nonmem/study_id/poppk/bbi.yaml contains the following options: ``
```

I'm also have difficulties getting this to work in an .Rmd document w/ really odd relative paths.  For example if I do the follow in .Rmd based on the ACOP demo

```{r}

MODEL_DIR <- here("nonmem/study_id/poppk")

print(bbi_version())

bbi_init(.dir = MODEL_DIR,
         .nonmem_dir = "/gstore/scratch/u/brooksl2/7.5.0-foss-2017a", 
         .nonmem_version = "nmfe75")

```

I get the following error where there seems to be a call to a working directory overriding the here() call? 
```
[1] "3.2.2"
Error: Cannot find id_12345/project/r_code/nonmem//id_12345/project/nonmem/study_id/poppk Make sure you are in the correct working directory and have passed the correct path to bbi_init(.dir)
```

kyleam:
> bbi_init(.dir = "nonmem/study_id/poppk", .nonmem_dir = "/gstore/scratch/u/id_12345/7.5.0-foss-2017a", .nonmem_version = "nmfe75")
> [...]
> Must specify a valid `.nonmem_version` for bbi_init(). nonmem/study_id/poppk/bbi.yaml contains the following options: ``

Underneath, `bbi init` looks through `.nonmem_dir` and tries to detect subdirectories for different NONMEM versions ([code](https://github.com/metrumresearchgroup/bbi/blob/f8a2c83869381a2d1221aeb302da880093079b91/cmd/init.go#L39-L76)).  Under each of those subdirectories, it looks for things like expected sub-directories, a license file, and an NONMEM executable.

As an example, on Metworx `.nonmem_dir` is `/opt/NONMEM/`.  Here's how that looks:

```
$ ls -1 /opt/NONMEM | grep nm
nm73gf
nm73gf_nmfe
nm74gf
[...]

PR #627, opened 2023-12-13
Add more documentation to `bbi_init`: make `.nonmem_dir` argument more clear

barrettk:
- It became apparent that we needed more clear documentation surrounding the NONMEM installation directory (`.nonmem_dir` argument), as users were incorrectly passing the directory containing the executable file, rather than the parent directory.

 - Added more context to the parameter itself, along with some additional details and an example.

closes https://github.com/metrumresearchgroup/bbr/issues/607

### DIFF
--- a/R/bbr.R
+++ b/R/bbr.R
@@ -324,13 +324,32 @@ bbi_help <- function(.cmd_args=NULL) {
 #'  **Notes on `.nonmem_dir`**
 #'
 #'   `bbi` tries to detect subdirectories for different NONMEM versions, along
-#'   with a license file and executable within the subdirectory. `.nonmem_dir`
-#'   should therefore point to the *parent* directory containing the executable.
+#'   with a license file and executable within the subdirectory.
 #'
-#'   If your NONMEM executable exists in the the location `/opt/NONMEM/nm75/run`,
-#'   `.nonmem_dir` should point to `/opt/NONMEM/`, rather than the subdirectory
-#'   `/opt/NONMEM/nm75`. You would then denote the preferred NONMEM version by
-#'   specifying `.nonmem_version = 'nm75'`
+#'   `.nonmem_dir` should point to the **parent** directory that contains one
+#'   or more NONMEM installations.  It should *not* point to the installation
+#'   directory itself.  Here is an example:
+#'
+#' ```
+#'     /opt/NONMEM/
+#'     |-- nm74
+#'     |   |-- license
+#'     |   |-- run
+#'     |   |-- source
+#'     |   | ...
+#'     |   `-- util
+#'     `-- nm75
+#'         |-- license
+#'         |-- run
+#'         |-- source
+#'         | ...
+#'         `-- util
+#' ```
+#'
+#'   `/opt/NONMEM/` contains two NONMEM installations, one for NONMEM 7.5
+#'   and one for NONMEM 7.4.  In this case, you should pass `"/opt/NONMEM"`
+#'   as `.nonmem_dir` and either `"nm74"` or `"nm75"` as `.nonmem_version`.
+#'
 #'
 #' @param .dir Path to directory to run `init` in (and put the resulting
 #'   `bbi.yaml` file)
@@ -355,10 +374,10 @@ bbi_help <- function(.cmd_args=NULL) {
 #' @examples
 #' \dontrun{
 #'
-#' # NONMEM executable found at `/opt/NONMEM/nm75/run`:
+#' # Top-level of NONMEM 7.5 installation is at `/opt/NONMEM/nm75`:
 #'
 #' bbi_init(
-#'    .dir = "inst/model/nonmem/basic",
+#'    .dir = MODEL_DIR,
 #'    .nonmem_dir = "/opt/NONMEM",
 #'    .nonmem_version = "nm75"
 #' )
--- a/R/bbr.R
+++ b/R/bbr.R
@@ -321,9 +321,22 @@ bbi_help <- function(.cmd_args=NULL) {
 #'   model. `submit_model()` and `submit_models()` also support passing a
 #'   configuration file via the `.config_path` argument.
 #'
+#'  **Notes on `.nonmem_dir`**
+#'
+#'   `bbi` tries to detect subdirectories for different NONMEM versions, along
+#'   with a license file and executable within the subdirectory. `.nonmem_dir`
+#'   should therefore point to the *parent* directory containing the executable.
+#'
+#'   If your NONMEM executable exists in the the location `/opt/NONMEM/nm75/run`,
+#'   `.nonmem_dir` should point to `/opt/NONMEM/`, rather than the subdirectory
+#'   `/opt/NONMEM/nm75`. You would then denote the preferred NONMEM version by
+#'   specifying `.nonmem_version = 'nm75'`
+#'
 #' @param .dir Path to directory to run `init` in (and put the resulting
 #'   `bbi.yaml` file)
-#' @param .nonmem_dir Path to directory with the NONMEM installation.
+#' @param .nonmem_dir Path to directory with the NONMEM installation. This
+#'   should be the **parent** directory that contains that installation. See
+#'   details for more information.
 #' @param .nonmem_version Character scalar for default version of NONMEM to use.
 #'   If left NULL, function will exit and tell you which versions were found in
 #'   `.nonmem_dir`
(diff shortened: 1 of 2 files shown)


barrettk:
@kyleam @seth127 I imagine we'll have a little back and forth regarding the documentation updates. Requesting you both to review in the event you both want to give your two cents for added clarity regarding this issue. In all honesty the confusion seems well founded, and I would probably have assumed the same based on the documentation if I hadnt worked with `bbr` as much. Point being I want to make sure we are as clear as possible, as I would intuitively thing to pass the directory containing the executable, rather than the parent directory. I think the example and details help, but let me know if it's too wordy or not concise enough.

R/bbr.R
```diff
@@ -321,9 +321,22 @@ bbi_help <- function(.cmd_args=NULL) {
 #'   model. `submit_model()` and `submit_models()` also support passing a
 #'   configuration file via the `.config_path` argument.
 #'
+#'  **Notes on `.nonmem_dir`**
+#'
+#'   `bbi` tries to detect subdirectories for different NONMEM versions, along
+#'   with a license file and executable within the subdirectory. `.nonmem_dir`
+#'   should therefore point to the *parent* directory containing the executable.
```
kyleam, line 328: Hmm, I'm not sure what "the executable" is here (I know you expand on it a bit in the next paragraph, but I'm explaining how I'd be confused by it as a reader).  The parent directory that should be passed as `.nonmem_dir` will typically just contain (assuming contain means first-level contents) one or more directories for the NONMEM installation.

Even _within_ a NONMEM installation's `run/` directory, I think saying "the executable" would be confusing because there are several executables.  (The one bbi cares about is `nmfe*`, but I don't think we should get into that level of detail here.)

R/bbr.R
```diff
@@ -321,9 +321,22 @@ bbi_help <- function(.cmd_args=NULL) {
...
+#'
+#'   If your NONMEM executable exists in the the location `/opt/NONMEM/nm75/run`,
```
kyleam, line 330: typo: "the the"

kyleam, line 330: "exists in the location `/opt/NONMEM/nm75/run`"  here must mean under the `run/` directory, but that could be interpreted as though run itself is the executable file of interest.

I wonder if this is the kind of thing were an example layout would help a lot.  For example

```
**Notes on `.nonmem_dir`**

`.nonmem_dir` should point to the parent directory that contains one
or more NONMEM installations.  It should not point to the installation
directory itself.  Here is an example:

    /opt/NONMEM/
    |-- nm74
    |   |-- license
    |   |-- run
    |   |-- source
    |   | ...
    |   `-- util
    `-- nm75
        |-- license
        |-- run
        |-- source
        | ...
        `-- util

`/opt/NONMEM/` contains two NONMEM installations, one for NONMEM 7.5
and one for NONMEM 7.5.  In this case, you should pass "/opt/NONMEM"
as `.nonmem_dir` and either "nm74" or "nm75" as `.nonmem_version`.
```

R/bbr.R
```diff
@@ -338,6 +351,18 @@ bbi_help <- function(.cmd_args=NULL) {
 #'
 #' @importFrom yaml read_yaml write_yaml
 #' @importFrom fs dir_exists
+#'
+#' @examples
+#' \dontrun{
+#'
+#' # NONMEM executable found at `/opt/NONMEM/nm75/run`:
```
kyleam, line 358: Same comment about it not being confusing what executable this is referring to.  Perhaps "# Top-level of NONMEM 7.5 installation is at `/opt/NONMEM/nm75`:"?

R/bbr.R
```diff
@@ -338,6 +351,18 @@ bbi_help <- function(.cmd_args=NULL) {
...
+#'
+#' bbi_init(
+#'    .dir = "inst/model/nonmem/basic",
```
kyleam, line 361: From the bbr checkout, using this `inst/` path is of course handy to test, but that's an unusual directory from the user perspective.  What about just using a `MODEL_DIR` stand-in?  That would align with what's used expo1.

kyleam (changes requested):
Thanks for working on this.  I think this is a clear improvement and would help avoid the confusion.

I'm leaving several comments and suggestions, but the only thing that I consider a blocker is revising the parts that talk about "the executable".

> as users were incorrectly passing the directory containing the executable file, rather than the parent directory.

Hmm, I'm confused by this part of the commit message.  I assume this is talking about the nmfe executable under the installations `run/` subdirectory.  In that case, I don't think this accurately captures the problem.  I think most users that were passing the wrong directory were passing the top-level installation directory (say `/path/to/nm-install/`) not the directory containing the nmfe executable (`/path/to/nm-install/run/`).

R/bbr.R
barrettk: I was trying to reword your comment [here](https://github.com/metrumresearchgroup/bbr/issues/596#issuecomment-1548797780):

> Underneath, bbi init looks through .nonmem_dir and tries to detect subdirectories for different NONMEM versions ([code](https://github.com/metrumresearchgroup/bbi/blob/f8a2c83869381a2d1221aeb302da880093079b91/cmd/init.go#L39-L76)). Under each of those subdirectories, it looks for things like expected subdirectories, a license file, and an NONMEM executable.

I think the example layout would definitely help alot, but was worried we wouldnt want to mention Metworx. Since you posted that example without mentioning Metworx ill definitely add that. I do think it adds clarity to mention the executable, or some file that we can point to and know it's 2 directories up from there. I'll try to rephrase and potentially remove that word altogether, but I thought your comment on that issue added some clarity

kyleam: > I was trying to reword your comment [...]

I see.  Yeah, the original comment is trying to distinguish a couple levels of directories (e.g., "Under of of those subdirectories ..."), and I think in the comment here it'd be unclear to me that it's talking about subdirectories within the NONMEM installation directory.

> an NONMEM executable.

Aside from my "an" typo, yeah, I was a bit sloppy there to say "NONMEM executable" and since I was describing in detail what bbi does, I should have specified `nmfe*`.  Again, in my eyes that a bit too much detail for here, but I'm okay with what you decide there.  The main thing I'm trying to avoid here is statements like "`.nonmem_dir` should therefore point to the *parent* directory containing the executable"; I have a hard time making that correct unless "containing" is interpreted recursively (which I think would be surprisingly).

R/bbr.R
barrettk: I left the first section, but removed the second mention of the executable in favor of your example layout

R/bbr.R
barrettk: Yeah I think that makes sense. Addressed.

R/bbr.R
barrettk: addressed

R/bbr.R
barrettk: I only have one mention of the executable in my most recent commit. Let me know if it be helpful to add `nmfe` before that mention, or if we should reword that sentence altogether. I kind of like mentioning those contents in the event external scientists are more familiar with specific NONMEM installation files, but definitely want it to be as obvious as possible

R/bbr.R
```diff
@@ -321,9 +321,41 @@ bbi_help <- function(.cmd_args=NULL) {
 #'   model. `submit_model()` and `submit_models()` also support passing a
 #'   configuration file via the `.config_path` argument.
 #'
+#'  **Notes on `.nonmem_dir`**
+#'
+#'   `bbi` tries to detect subdirectories for different NONMEM versions, along
+#'   with a license file and executable within the subdirectory.
```
kyleam, line 327: My preference would be to drop this entirely.  What `bbi` looks for to decide a directory is a NONMEM installation is an implementation detail, and I find it a bit confusing as a lead off paragraph.  For example

> `bbi` tries to detect subdirectories for different NONMEM versions

Where does it look for these?  The reader is supposed to infer `.nonmem_dir`, at this point they haven't even been given the details below about htat being the parent directory.

> along with a license file and executable within the subdirectory

it's not clear to me which directory "the subdirectory" is referring to.

I think the layout below makes it clear what `.nonmem_dir` should be, and this first paragraph is more likely to cause confusion.

R/bbr.R
```diff
@@ -321,9 +321,41 @@ bbi_help <- function(.cmd_args=NULL) {
...
+#'   `/opt/NONMEM/` contains two NONMEM installations, one for NONMEM 7.5
+#'   and one for NONMEM 7.4.  In this case, you should pass `"/opt/NONMEM"`
+#'   as `.nonmem_dir` and either `"nm74"` or `"nm75"` as `.nonmem_version`.
+#'
+#'
 #' @param .dir Path to directory to run `init` in (and put the resulting
 #'   `bbi.yaml` file)
-#' @param .nonmem_dir Path to directory with the NONMEM installation.
+#' @param .nonmem_dir Path to directory with the NONMEM installation. This
+#'   should be the **parent** directory that contains that installation. See
```
kyleam, line 357: [ This was suppose to be a part of my first review, but I don't see it; I must have forgot to post it.  Sorry. ]

The second sentence would point people in the right direction, but I find the combination of the first and second sentences a bit confusing.  To me it almost reads as the though the second sentence corrects what the first just said.  (That all depends on how the reader interprets "directory with", which I suppose is probably the core of the original issue.)

I'd be in favor of reducing the first two sentences down to one.  Perhaps "Path to parent directory containing one or more NONMEM installations. See [...]".

kyleam (approved):
Thanks for the updates. 

I think this PR makes things this argument much clearer.  

Approving, but please see my second comment because that was supposed to come with my _first_ review, but I somehow left it out.

seth127:
Thanks for all the work on this, both of you. I think this is much better at this point. I especially think [showing the potential directory layout](https://github.com/metrumresearchgroup/bbr/blob/f0aecbd53eb33d2e46ebafb5c7512db3c9eec6e1/R/bbr.R#L324-L351) will be very helpful. It would be for me, at least.

kyleam (approved):

PR #630, opened 2024-01-03
test-workflow-bbi.R tweaks for upcoming Metworx

kyleam:
* The first commit of this series reworks some assertions in `test-workflow-bbi.R` so that they pass on the upcoming Metworx (in addition to the current Metworx).

 * The second commit was motivated by the upcoming Metworx but, due to changes on the Metworx side, is no longer strictly necessary.  I think it's still a good change because it eliminates some cases where we'd need to make a bbr release to accommodate Metworx-side changes.

### DIFF
--- a/R/helpers-for-tests.R
+++ b/R/helpers-for-tests.R
@@ -70,3 +70,19 @@ skip_if_old_nmrec <- function(v) {
       glue("nmrec version is {utils::packageVersion('nmrec')}. Test requires version {v} or later"))
   }
 }
+
+#' Return path to mpiexec executable to use in tests.
+#'
+#' @noRd
+get_mpiexec_path <- function() {
+  res <- Sys.getenv(
+    "BBR_TESTS_MPIEXEC_PATH",
+    unset = unname(Sys.which("mpiexec"))
+  )
+  if (nchar(res)) {
+    return(res)
+  }
+
+  # Fall back to bbi default.
+  return("/usr/local/mpich3/bin/mpiexec")
+}
--- a/tests/testthat/test-workflow-bbi.R
+++ b/tests/testthat/test-workflow-bbi.R
@@ -41,7 +41,12 @@ withr::with_options(list(
   if (fs::file_exists(file.path(MODEL_DIR_BBI, "bbi.yaml"))) fs::file_delete(file.path(MODEL_DIR_BBI, "bbi.yaml"))
 
   # create new bbi.yaml
-  bbi_init(MODEL_DIR_BBI, "/opt/NONMEM", "nm74gf")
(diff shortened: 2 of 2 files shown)


kyleam:
@seth127 @barrettk These are the last pieces on my end that I'd like to get in before the release.  

@seth127 I'm requesting your review since we've already discussed both of these, but please switch it over to @barrettk if that works better.

barrettk:
@kyleam Just for additional background, is the tolerance now necessary on the upcoming Metworx BP because we added a new version of NONMEM? If not im curious why the results would differ more now

kyleam:
> is the tolerance now necessary on the upcoming Metworx BP because we added a new version of NONMEM?

That test is using `nm74gf`.  As far as I know, there's no update to that version (e.g., a patch version). 

> If not im curious why the results would differ more now

When I say "[t]he expectation is too stringent because these values may change with different NONMEM versions or different systems" in the commit message, the "different systems" in this case is the OS update from Bionic to Focal (which includes, among many things, an update to gfortran).  I think once you update the OS, all bets are off with respect to being able to produce the same value (see, for example, [all the considerations](https://mc-stan.org/docs/reference-manual/reproducibility.html) for reproducing Stan results).  If you or others have any NONMEM-specific expectations related to this, though, that's of course worth discussing so that we don't dismiss a discrepancy that we should look more into.

barrettk:
>the "different systems" in this case is the OS update from Bionic to Focal (which includes, among many things, an update to gfortran).

Had a feeling this was more of the concern, rather than an update to NONMEM. But awesome thanks for explaining. 

I looked into the `mpi_exec_path` to get a better understanding of [how/where](https://github.com/metrumresearchgroup/bbi/blob/835953efed10fff2d7024ac1c3dd88949b268c05/cmd/nonmem.go#L196-L198) it was being set. While I dont really understand what the last line is doing, I feel pretty good about the changes you introduced here in preparation for a potentially new `mpiexec` file location. 

I'll approve from my end, but it sounds like Seth may still want to look at it given your comment about you two having discussed this previously.

barrettk (approved):
Looks good to me (left comments and questions in previous messages)

kyleam:
> Looks good to me (left comments and questions in previous messages)

@barrettk Thanks for the discussion and reviews.  I'll leave this open for a bit in case Seth wants to chime in, but from my view this is ready to go (and if you're at the point where you're ready to cut a release PR, please feel free to merge this).

seth127:
This all looks good to me. Thanks both.

kyleam:
Drone is currently not accepting any jobs.  Holding off merge until it's resolved.

PR #631, opened 2024-01-03
Add inheriting parameter estimates to vignettes

barrettk:
Add section on inheriting parameter estimates to getting started vignette

### DIFF
--- a/vignettes/getting-started.Rmd
+++ b/vignettes/getting-started.Rmd
@@ -223,7 +223,7 @@ mod1 <- read_model(file.path(MODEL_DIR, 1))
 
 ## copy_model_from()
 
-Now you can create a new model object, based on the original, copying and renaming the control stream in the process. By default, the `copy_model_from()` function will increment the name of the new model to the next available integer in your modeling directory, but you can also pass a custom name to the `.new_model` argument. See [`?copy_model_from`](https://metrumresearchgroup.github.io/bbr/reference/copy_model_from.html) for more details this, and other optional arguments.
+Now you can create a new model object, based on the original, copying and renaming the control stream in the process. By default, the `copy_model_from()` function will increment the name of the new model to the next available integer in your modeling directory, but you can also pass a custom name to the `.new_model` argument. See [`?copy_model_from`](https://metrumresearchgroup.github.io/bbr/reference/copy_model_from.html) for more details and other optional arguments.
 
 The `copy_model_from()` call below will create both `2.ctl` and `2.yaml` files in the same directory as the parent model, and return the model object corresponding to them. (`copy_model_from()` also stores the model's "ancestry" which can be useful later in the project, as shown in the [Using the based_on field](https://metrumresearchgroup.github.io/bbr/articles/using-based-on.html) vignette.)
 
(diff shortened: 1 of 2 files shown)


barrettk:
@seth127 I have been unable to render the vignette locally for some reason (not surprising though). Stepping through the blocks works with no issues, but it seemingly gets hung up when trying to copy the model file:

```r
mod2 <- copy_model_from(mod1) %>%
  update_model_id()
```
image

kyleam:
@barrettk I don't know offhand what the source of your error is, but an alternative that may help you side step the issue: render the site locally with `pkgdown::build_site()`.  That errors because`inherit_param_estimates` is missing from `_pkgdown.yml`, but once that's resolved your changes render fine on my end.

barrettk:
@kyleam Thought about that but was trying to figure out if we *should* be doing anything else in terms of setting up file paths. Ended up just following your suggestion to move forward and make sure it rendered correctly. I updated the `pkgdown` file in this PR as opposed to in the separate release PR, but let me know if that's an issue.

kyleam:
> Thought about that but was trying to figure out if we _should_ be doing anything else in terms of setting up file paths. 

I don't really know what you're referring to (I'm not even spotting where you describe how you're rendering the vignettes), but of course please investigate any problems.  Does whatever approach you're using work before this PR?  Then that would definitely seem worth looking into the breakage.  If it's not a new problem, then it's less of a concern.

> Ended up just following your suggestion to move forward and make sure it rendered correctly. I updated the `pkgdown` file in this PR as opposed to in the separate release PR, but let me know if that's an issue.

It's a fix for something that should have been done in the original PR and that blocks building the site to check the vignette change here, so sure I think it makes sense as part of this PR.  (The reason the pkgdown index update is often done in the release PR isn't because it belongs there but instead because we often forget to do it in the original PR.)

barrettk:
> Does whatever approach you're using work before this PR? Then it would definitely seem worth looking into the breakage. If it's not a new problem, then it's less of a concern.

I was just clicking the "knit" button within the Rmarkdown document. It was failing without the recent changes, but I got the feeling you guys may either A) be rendering it differently, or B) have some environment variable set before rendering vignettes locally that I didnt. My Metworx disc is a little wonky, and my profile is not sourced during each R session (unless I manually do it), so im never surprised when things like this only happen for me.

kyleam:
> I was just clicking the "knit" button within the Rmarkdown document.

I see.  If I open RStudio and hit that button (with "Knit directory" set to "Project Directory"), it renders without an error for me.  That requires installing bbr into the environment, which as far as I know is [expected](https://r-pkgs.org/vignettes.html#sec-vignettes-workflow-writing).

Another way that works on my end: using `devtools::build_rmd("vignettes/getting-started.Rmd")` (no need to install bbr in that case).

> It was failing without the recent changes [...]

Ah, good to know it's not due to a change here. 

> [...] but I got the feeling you guys may either A) be rendering it differently, or B) have some environment variable set before rendering vignettes locally that I didnt.

I probably don't represent the [normal workflow](https://r-pkgs.org/vignettes.html#sec-vignettes-workflow-writing), but fwiw I tend to use `devtools::build_rmd()` (outside of RStudio) to render and check the vignette.  Rendering works for me without setting any additional options or environment variables.

kyleam:
> requested review from [seth127](https://github.com/seth127) and [kyleam](https://github.com/kyleam)

@seth127 Let me know if you prefer I take this, but otherwise I'll leave to you.  Thanks.

seth127 (approved):
Two very small changes, but then this is ready to go. Thanks.

vignettes/getting-started.Rmd
```diff
@@ -305,6 +337,8 @@ run_log(MODEL_DIR, .include = "2 compartment") %>%
 ```{r cleanup bottom, include=FALSE}
 # delete old files
 cleanup()
+# remove new tag from mod1
+mod1 <- mod1 %>% remove_tags("orig acop model")
```
barrettk, line 341: FYI this change was to revert the new tag that gets added to mod1. Probably should have been there before, but wouldnt impact anything unless it was rendered locally/the changes were committed.

PR #633, opened 2024-01-05
Prepare for 1.8.0 release

barrettk:
changeset: https://github.com/metrumresearchgroup/bbr/compare/1.7.0...release/1.8.0

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -304,6 +304,9 @@ steps:
   - git config --global user.email drone@metrumrg.com
   - git config --global user.name Drony
   - git fetch --tags
+  # clone nmrec from main branch for building package
+  - git clone --depth 1 https://github.com/metrumresearchgroup/nmrec.git /tmp/nmrec
+  - R -s -e 'devtools::install("/tmp/nmrec", upgrade = "never")'
   - R -s -e 'pkgpub::create_tagged_repo(.dir = '"'"'/ephemeral'"'"')'
   environment:
     BBI_EXE_PATH: /ephemeral/bbi
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,8 +6,8 @@
   estimates of a model using the final estimates of a previously executed model. 
   (#623)
 * Improved documentation and vignettes. (#622, #627, #631)
-* Some internal functions now rely on `nmrec` for parsing control stream files,
-  rather than regex/string manipulation. (#600, #603, #604, #606, #612)
+* Some functions now rely on `nmrec` for parsing control stream files, rather 
+  than regex/string manipulation. (#600, #603, #604, #606, #612)
 * `nm_join` now tracks the origin of each column and stores it as an attribute 
   of the return value (mostly important for `bbr.bayes`). (#617)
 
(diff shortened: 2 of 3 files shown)


barrettk:
@kyleam FYI I was uncertain about a few of the items I mentioned in the NEWS. Beyond changing any wording, let me know if you think any of those items should be removed/not highlighted in the NEWS

NEWS.md
```diff
@@ -1,3 +1,27 @@
+# bbr 1.8.0
+
+## New features and changes
+
+* New `inherit_param_estimates()` function for setting the initial parameter 
+  estimates of a model using the final estimates of a previously executed model. 
+  (#623)
+* Improved documentation and vignettes. (#622, #627, #631)
+* Some internal functions now rely on `nmrec` for parsing control stream files,
+  rather than regex/string manipulation. (#600, #603, #604, #606, #612)
```
kyleam, line 10: Maybe drop "internal" since with a quick glance at least one of the rewritten functions is exported?

NEWS.md
```diff
@@ -1,3 +1,27 @@
...
+* `nm_join` now tracks the origin of each column and stores it as an attribute 
+  of the return value (mostly important for `bbr.bayes`). (#617)
+
+## Bugs addressed
+
+* `update_model_id` would replace the id with the `based_on` field, which was a 
+  relative file path rather than a model id. In other words, this function did 
+  not support child models created in sub-directories (or otherwise a different 
+  directory than the parent model). (#614)
+* Various test fixes and improvements (#621, #630)
```
kyleam, line 20: Fine as is, but I think you could drop both of these as not user-facing.

kyleam (approved):
Left a couple of suggestions, but looks good to me.  Thanks for preparing this release.

.drone.yml
```diff
@@ -304,6 +304,9 @@ steps:
   - git config --global user.email drone@metrumrg.com
   - git config --global user.name Drony
   - git fetch --tags
+  # clone nmrec from main branch for building package
+  - git clone --depth 1 https://github.com/metrumresearchgroup/nmrec.git /tmp/nmrec
+  - R -s -e 'devtools::install("/tmp/nmrec", upgrade = "never")'
   - R -s -e 'pkgpub::create_tagged_repo(.dir = '"'"'/ephemeral'"'"')'
```
kyleam, line 310: As I mentioned out of band last Friday: it'd be better for this to use 

```
devtools::install_dev_deps(upgrade = "never")'
```

because this step should not assume that all of the dependencies are baked into the image.  And, for the purposes of the build, the nmrec version on MPN is good enough, so there's no need for the clone/install from the GitHub repo.

kyleam (changes requested):

.drone.yml
barrettk: Sorry didnt realize you responded offline! Just read through all your comments; updating now

kyleam: No worries.

[ For posterity: the tag pointing to a311e37 was already pushed and triggered the S3 upload, so @barrettk and I restored the tip of this PR to there.  There was a new tag pushed pointing to c65f9ef, but we canceled it before the release step was triggered.  We'll restore that original tag. ]

kyleam (approved):

Issue #637, opened 2024-01-17
`inherit_param_estimates` fails when flags needed for `model_summary()` call


seth127:
# Summary of problem 

In trying out the [`inherit_param_estimates()`](https://metrumresearchgroup.github.io/bbr/reference/inherit_param_estimates.html) feature (new in [1.8.0](https://github.com/metrumresearchgroup/bbr/releases/tag/1.8.0)) a user noticed the following problem:

> It seems that under the hood it uses `model_summary` in some way, and if the previous model has not converged correctly or `$COV` has not been evaluated (no `.grd` file) the function won’t evaluate. According to the error text I should be able to suppress this by using the `bbi` argument `–no-grd-file`. However, `inherit_param_estimates` does not seem to have an option to input `bbi` arguments (i.e. there is no way to pass through `.bbi_args`)?

The user shared the following screen shot:

<img width="907" alt="Screenshot 2024-01-17 at 10 03 32 AM" src="https://github.com/metrumresearchgroup/bbr/assets/7806083/739a7a98-e842-4cab-8d07-f495168e3f6e">

---

I verified that we do indeed [call `model_summary()` to get the final estimates from the parent model](https://github.com/metrumresearchgroup/bbr/blob/main/R/inherit-param-estimates.R#L70), but provide no interface for passing through arguments to it.

# Proposed solution

I lean towards handling this [the way we do in `nm_join()`](https://github.com/metrumresearchgroup/bbr/blob/main/R/nm-join.R#L90-L93), which is another function that calls `model_summary()` internally and only intends to use a portion of the returned summary object: **add `.bbi_args`, defaulting to `no_grd_file = TRUE, no_shk_file = TRUE` to the `inherit_param_estimates()` interface.** (See ['nm_join()` docs](https://metrumresearchgroup.github.io/bbr/reference/nm_join.html) for analogous rationale.)

With this approach:

1. The default would handle the (likely most common) case that the user reported here.
2. We would cover other unanticipated cases by giving users an interface to pass arguments through to `model_summary()`.

@barrettk @kyleam any thoughts on that, or alternatives that we might consider?

kyleam:
Your suggested approach makes sense to me.

PR #638, opened 2024-01-17
add .bbi_args argument to inherit_param_estimates

barrettk:
Add `.bbi_args` argument to inherit_param_estimates, defaulting to `list(no_grd_file = TRUE, no_shk_file = TRUE)`, as we only need to extract parameter estimates.

closes https://github.com/metrumresearchgroup/bbr/issues/637

### DIFF
--- a/tests/testthat/test-inherit-param-estimates.R
+++ b/tests/testthat/test-inherit-param-estimates.R
@@ -185,6 +185,12 @@ withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
         list(ext = "shk", missing = "shrinkage_details")
       )
 
+      # Ensure models get deleted if tests fail
+      on.exit({
+        try(delete_models(mod2, .tags = NULL, .force = TRUE))
+        try(delete_models(mod3, .tags = NULL, .force = TRUE))
+      })
+
       for (.tc in TEST_CASES) {
         # create new model to inherit estimates from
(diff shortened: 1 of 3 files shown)


tests/testthat/test-inherit-param-estimates.R
```diff
@@ -177,6 +177,39 @@ withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
...
+          inherit_param_estimates(mod3, .bbi_args = list()),
+          glue::glue("[Nn]o file present at.*2/2\\.{.tc$ext}")
+        )
+
+        # works correctly with flag added
+        args_list <- list()
+        args_list[[as.character(glue::glue("no_{.tc$ext}_file"))]] <- TRUE
+        mod3 <- inherit_param_estimates(mod3, .bbi_args = args_list)
+
+        delete_models(list(mod2, mod3), .tags = NULL, .force = TRUE)
```
seth127, line 210: I think we might as well do a quick test that the parameters were actually inherited, before we leave the loop here. Something quick like we do [in the test above](https://github.com/metrumresearchgroup/bbr/blob/b3baa845fa0252a0b881f049bb62b13fe7f9bcd6/tests/testthat/test-inherit-param-estimates.R#L176-L178).

seth127 (changes requested):

seth127 (approved):
Thanks for the updates. Looks good.

Issue #639, opened 2024-01-18
`inherit_param_estimates` tests fail outside of Drone or Metworx


kyleam:
From b226df7a (tests: add missing skips for SUM1-based tests, 2023-11-20, gh-621):

> When not running on Metworx or Drone, setup-workflow-ref.R doesn't
> define SUM1.  Most spots that use SUM1 call
> skip_if_not_drone_or_metworx() to avoid an "object 'SUM1' not found"
> error.  Fix the two spots that don't.

Since then, the merge of gh-623 introduced five more spots that fail with that error, all within `test-inherit-param-estimates.R`.

cc: @barrettk

seth127:
Closed by https://github.com/metrumresearchgroup/bbr/pull/641

PR #641, opened 2024-01-23
`inherit_param_estimates` tests fail outside of Drone or Metworx 

barrettk:
skip tests requiring `SUM1` if not on drone or metworx

closes https://github.com/metrumresearchgroup/bbr/issues/639


kyleam (approved):
Confirmed that all tests pass when I unset `METWORX_VERSION`.  Thanks.

PR #642, opened 2024-01-24
Prepare for 1.8.1 Release

barrettk:
Changeset: https://github.com/metrumresearchgroup/bbr/compare/1.8.0...release/1.8.1

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,10 +2,10 @@
 
 ## Bugs addressed
 
-* `inherit_param_estimates()` would occasionally fail during the `model_summary()`
-call if the previous model had not converged correctly or `$COV` has not been 
-evaluated (no `.grd` file). Added `.bbi_args` argument (similar to `nm_join()`),
-that defaults to *not* looking for `.shk` and `.grd` files. (#638)
+* `inherit_param_estimates()` calls `model_summary()` underneath and that failed
+when the parent model did not have `.grd` or `.shk` files. `inherit_param_estimates()`
+now has a `.bbi_args` argument (similar to `nm_join()`) that defaults to *not* 
+looking for `.shk` and `.grd` files. (#638)
 
 # bbr 1.8.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,12 @@
+# bbr 1.8.1
+
+## Bugs addressed
+
+* `inherit_param_estimates()` would occasionally fail during the `model_summary()`
+call if the previous model had not converged correctly or `$COV` has not been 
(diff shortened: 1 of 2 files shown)


barrettk:
## Scorecard Details
image

NEWS.md
```diff
@@ -1,3 +1,12 @@
+# bbr 1.8.1
+
+## Bugs addressed
+
+* `inherit_param_estimates()` would occasionally fail during the `model_summary()`
+call if the previous model had not converged correctly or `$COV` has not been 
+evaluated (no `.grd` file). Added `.bbi_args` argument (similar to `nm_join()`),
+that defaults to *not* looking for `.shk` and `.grd` files. (#638)
```
kyleam, line 8: Reading this, I tripped a bit on "occasionally" (does that mean even models following under the "if ..." only fail occasionally?) and the "not converged correctly" (what are all the cases that includes?).

I think I'd find it clearer with these tweaks

```
* `inherit_param_estimates()` calls `model_summary()` underneath and that
   failed when the parent model did not have `.grd` or `.shk` files.
   `inherit_param_estimates()` now has a `.bbi_args` argument (similar to
   `nm_join()`) that defaults to *not* looking for `.shk` and `.grd`
   files. (#638)
```

... but I'm all right if you prefer to stick with what you've got.

NEWS.md
```diff
@@ -1,3 +1,12 @@
+# bbr 1.8.1
+
+## Bugs addressed
+
+* `inherit_param_estimates()` would occasionally fail during the `model_summary()`
+call if the previous model had not converged correctly or `$COV` has not been 
+evaluated (no `.grd` file). Added `.bbi_args` argument (similar to `nm_join()`),
```
kyleam, line 7: Nit-pick: I think the comma at the end of this sentence is unnecessary.

kyleam (approved):
Thanks.  Left some suggestions for your consideration, but works for me as is.

---

On bf4cff7d9, I get a covr score of 0.8463.  Is your lower score because you ran it with the nmrec on MPN rather than the latest (required for the inherit parameter tests)?

kyleam:
> @kyleam I ran with these args:

Ah, perfect, so it's not that.

> However, I was also surprised to see it change from `1.8.0`. Perhaps some tests were skipped on my end?

I think it's just a matter of the bbi version, then.  I'm running with `v3.3.0`.  I suspect the numbers you're reporting for this PR are with a version before 3.2.0 (perhaps 3.1.1, which ships with the current Metworx).

In terms of the difference with what you reported in gh-633, could you have been using a newer bbi version for that?

(This is all speculation; I'll confirm that I see the same numbers when I run with bbi 3.1.1.)

PR #643, opened 2024-01-29
run_log, config_log, summary_log: ignore nested models

kyleam:
For NONMEM Bayes models (model_type=nmbayes), bbr.bayes creates a series of submodels (of model_type=nonmem) in the output directory of the main model.  These models are an internal implementation detail and are not meant to be discovered run_log(), config_log(), or summary_log().

In general, there's not a clear case where we'd need or want to support finding models within the output directory of another model. Teach find_models (used underneath by the three *_log() functions) to ignored submodels.

Re: https://github.com/metrumresearchgroup/bbr.bayes/issues/40

---

- [x] Verify that bbr.bayes CI passes ([drone run](https://github-drone.metrumrg.com/metrumresearchgroup/bbr.bayes/438), metrumresearchgroup/bbr.bayes#123)

PR #644, opened 2024-01-29
config_log: make compatible with nmbayes

kyleam:
`config_log` is incompatible with nmbayes models: https://github.com/metrumresearchgroup/bbr.bayes/issues/40

To make it work, we need

 * to not collect nested models (handled by #643)
 * add a custom `config_log_make_entry` method for nmbayes (todo: link)
 * adjust the `config_log` logic to work with the nmbayes model structure

This series handles the last point.

---

- [x] Verify that bbr.bayes CI passes ([drone run](https://github-drone.metrumrg.com/metrumresearchgroup/bbr.bayes/438), metrumresearchgroup/bbr.bayes#123)


seth127 (approved):
Looks great

PR #645, opened 2024-01-29
summary_log: make compatible with nmbayes

kyleam:
`summary_log` is incompatible with nmbayes models: https://github.com/metrumresearchgroup/bbr.bayes/issues/40

To make it work, we need

  * to not collect nested models (handled by https://github.com/metrumresearchgroup/bbr/pull/643)
  * adjust the `summary_log` logic to work with the nmbayes, in particular the object returned by `model_summary.bbi_nmbayes_model()`. (Edit: The bbr.bayes `model_summary` methods now error; [discussion](https://github.com/metrumresearchgroup/bbr/pull/645#discussion_r1472279335))

This series handles the second item above.

---

- [x] Verify that bbr.bayes CI passes ([drone run](https://github-drone.metrumrg.com/metrumresearchgroup/bbr.bayes/438), metrumresearchgroup/bbr.bayes#123)


R/summary-log.R
```diff
@@ -83,10 +83,10 @@ summary_log_impl <- function(.mods, ...) {
   check_model_object_list(.mods)
 
   # bbr.bayes kludge: this is sneaking in information about a warning from
-  # model_summary.bbi_stan_model().
+  # model_summary.bbi_stan_model() and model_summary.bbi_nmbayes_model().
   res_list <- suppressSpecificWarning(
     model_summaries(.mods, ...),
```
seth127, line 88: # Summary 

I found something that I think is a problem here: In https://github.com/metrumresearchgroup/bbr.bayes/issues/40 I had written a [comment](https://github.com/metrumresearchgroup/bbr.bayes/issues/40#issuecomment-1319163438) saying

> we should at least makes sure that it gives back some reasonable "failure mode" and doesn't try to do something like stick an entire draws array into each row of the tibble (I worry that might be what it would do now...)

I was wrong about that happening for Stan models, but it appears to be exactly what is happening for `nmbayes` models at this point. 

# Examples

_I found this testing on [`bbr.bayes@nmbayes-log`](https://github.com/metrumresearchgroup/bbr.bayes/tree/nmbayes-log), so these examples below are using `bbr.bayes` models. This felt like the right place for the comment though, since this is the code that will likely need to address it._

## nmbayes pulling in draws_array

```r
slog <- summary_log("inst/model/nonmem/bayes")
class(dplyr::pull(slog, bbi_summary)[[1]])
# [1] "draws_nmbayes" "draws_array"   "draws"         "array" 
```

<details><summary>and just to show it actually has all the draws...</summary>

```r
dplyr::pull(slog, bbi_summary)[[1]]
# # A draws_array: 50 iterations, 2 chains, and 985 variables
# , , variable = THETA[1]
# 
#          chain
# iteration     1      2
#         1 -0.20 -0.169
#         2 -0.14 -0.049
#         3 -0.17 -0.063
#         4 -0.15 -0.036
#         5 -0.13 -0.016
# 
# , , variable = THETA[2]
# 
#          chain
# iteration   1   2
#         1 3.7 3.7
#         2 3.8 3.8
#         3 3.7 3.8
#         4 3.8 3.8
#         5 3.8 3.9
# 
# , , variable = THETA[3]
# 
#          chain
# iteration   1   2
#         1 1.1 1.1
#         2 1.2 1.2
#         3 1.1 1.1
#         4 1.2 1.2
#         5 1.2 1.2
# 
# , , variable = THETA[4]
# 
#          chain
# iteration   1   2
#         1 4.5 4.5
#         2 4.4 4.5
#         3 4.4 4.5
#         4 4.4 4.5
#         5 4.4 4.4
# 
# # ... with 45 more iterations, and 981 more variables

```

</details>

This will likely be a problem, once people are using this on real models.

## stan model pulling in fit object

```r
slog2 <- summary_log("inst/model/stan")
purrr::map(dplyr::pull(slog2, bbi_summary), class)
# [[1]]
# [1] "CmdStanGQ"  "CmdStanFit" "R6"        
# 
# [[2]]
# [1] "CmdStanMCMC" "CmdStanFit"  "R6"         
# 
# [[3]]
# [1] "CmdStanMCMC" "CmdStanFit"  "R6"   
```

This might be fine, because it doesn't actually pull in the draws. I still don't think we want to stick it in the `summary_log()` output though... It's hard for me to imagine anyone finding that useful.

# What to do?

This probably merits some discussion. You can get a high-level view of a "mixed" summary log with `summary_log("inst/model/", .recurse = T)`. The first few columns of that look like

Screenshot 2024-01-30 at 10 46 49 PM

I _think_ my ideal state would be to have rows 2-5 of that tibble have `NA` in the `bbi_summary` column, and some kind of useful message in the `error_msg` column. That feels to me like what the user would expect, given the current (lack of) functionality. 

That said, it's not obviously clear to me how we'd implement that... `model_summaries()` currently falls through mapping over `model_summary()` (via `model_summaries_onetry()`, as of this series). I'm pretty sure we _don't_ want `model_summary()` to error.

One idea would be to trigger off the warning: instead of suppressing it here, catch it in `model_summaries_onetry()` and relay it to the `error_msg` column (while tossing the "summary" object that we get back). That would mean (for either Stan or `nmbayes` models) you would have `NA` in the `bbi_summary` column, and some version of the following in the `error_msg` column`:

```
Warning message:
model_summary.bbi_nmbayes_model() is not implemented.
A draws object will be returned for /data/home/sethg/metrumresearchgroup/bbr.bayes/inst/model/nonmem/bayes/1100
Consider calling `read_fit_model()` directly instead.
```

That seems pretty reasonable to me, given the current state. It's not the most helpful, but I don't think we're going to be able to give the user anything _actually_ helpful until we decide to [implement some version of `model_summary()` for these model types](https://github.com/metrumresearchgroup/bbr.bayes/issues/27).

Those are my thoughts at this point. I'm open to thoughts and ideas on this, but I will reiterate that I _don't_ think returning the entire `draws_array` object in each row of the tibble is a good idea.

seth127 (changes requested):
This mostly looks great, with the exception of the potentially large issue in the inline comment in `R/summary-log.R`. Marking this as requesting changes, because we at least need to discuss that issue, and I think probably make some (hopefully small) changes.

R/summary-log.R
kyleam: > nmbayes pulling in draws_array

Excellent point.  After making `summary_log` not fail on the array, I should've stepped back and asked whether it should be there at all.

> [The Stan object] doesn't actually pull in the draws. I still don't think we want to stick it in the `summary_log()` output though... It's hard for me to imagine anyone finding that useful.

Agreed.  IIRC it's done that from the beginning and like you say, it's not as big of an issue given it avoids in all the draws, but I don't see the value in putting it there.

> I _think_ my ideal state would be to have rows 2-5 of that tibble have `NA` in the `bbi_summary` column, and some kind of useful message in the `error_msg` column.

Right.

> That said, it's not obviously clear to me how we'd implement that... `model_summaries()` currently falls through mapping over `model_summary()` (via `model_summaries_onetry()`, as of this series). I'm pretty sure we _don't_ want `model_summary()` to error.

Hmm, the error seems like a good option to me.  Ignoring the fact that `summary_log` is very tailored to `bbi nonmem summary --json` output, it seems to me that the core conceptual problem is that these `model_summary` methods are returning something that's very much not a summary.

What value do you see in having `model_summary` warn and relay the `read_fit_model` result (as it does for Stan and nmbayes)?  Why not just error and point to `read_fit_model`?

seth127: > it seems to me that the core conceptual problem is that these model_summary methods are returning something that's very much not a summary.

Yes, this is well said.

> What value do you see in having `model_summary` warn and relay the `read_fit_model` result (as it does for Stan and `nmbayes`)? Why not just error and point to `read_fit_model`?

Hmm, that's fair. The more I think about it... I actually don't see much value in the current behavior. I think I'm leaning towards agreeing with this, but I'm going to record a bit of due diligence below:

* I would be concerned about breaking existing functionality (by replacing a warning with an error), but we're still on such an early version and it's not clear that anyone is currently relying on this functionality. In that regard, now would be the right time to change it.
* We don't mention `model_summary()` anywhere in [Expo2](https://merge.metrumrg.com/expo/expo2-stan/), that I see. We do show `read_fit_model()` in several places (e.g. for [MCMC diagnostics](https://merge.metrumrg.com/expo/expo2-stan/posts/initial-model-diagnostics.html#mcmc-diagnostics))
* We do mention it in our ["Getting Started" vignette](https://metrumresearchgroup.github.io/bbr.bayes/articles/getting-started-stan.html#using-read_fit_model), but it clearly says it's essentially a no-op that passes to `read_fit_model()`. Updating that to say that it errors and recommends `read_fit_model()` seems reasonable.
  * The [`model_summary()` function docs](https://metrumresearchgroup.github.io/bbr.bayes/reference/model_summary.bbi_stan_model.html) are pretty much the same.
  * As long as we remember to update both of these, I don't see much issue.

Let's take a minute to ask a few users (offline) what they think, but I'm feeling like that's probably the right path forward. And in that case... this code wouldn't change, right? I think we would just change it on the `bbr.bayes` side to error instead of warn.

seth127: > breakage here seems low risk and unlikely, not to mention the plan was always to change the value at some point (hence the warning), so it was going to break at some point anyway.

Yep, I agree. And the conversations I've had with Metrum users today confirms that this will be very low (perhaps no) impact.

So yes, I'd say let's go ahead with switching that to an error. I'm fine with you doing it as part of https://github.com/metrumresearchgroup/bbr.bayes/pull/123 if that's your preference.

Let me know when you get whatever small changes need to happen in here pushed and I'll come back and approve.

kyleam:
> Once those changes are pushed to [metrumresearchgroup/bbr.bayes#123](https://github.com/metrumresearchgroup/bbr.bayes/pull/123), I'll add a comment here pointing to the Drone run.

That's here: <https://github-drone.metrumrg.com/metrumresearchgroup/bbr.bayes/445/1/4>.  The failure is unrelated and probably due to rate limiting.  I'll update this message with all green build once we have one, but this is ready for review.

seth127 (approved):
The changes here look good. For posterity, the bulk of the changes relevant to the above `summary_log()` discussion are on `bbr.bayes` side in https://github.com/metrumresearchgroup/bbr.bayes/pull/123/commits/327df7379df965be74f9ecdf81e74cab57bba002 . Thanks @kyleam .

kyleam:
@seth127 Thanks for taking another look (and for catching the issue with putting draws in the summary_log output)

Issue #632, opened 2024-01-05
New Feature: jitter/tweak initial parameter estimates


barrettk:
## Background
[Some Background Discussion](https://github.com/metrumresearchgroup/bbr/issues/582#issuecomment-1834615092)

Sample code for how we propose updating the estimates is shown below, though note that we will be parsing the control stream file rather than pulling parameter estimates from a model summary of a previously executed model.
```r
thetas <- get_theta(model_summary(mod106))

# user input, how much to tweak by
.p <- 0.1

# actual percentage to tweak each one by
tweak_perc <- runif(length(thetas), -.p, .p)

# actual tweak is made (this is how PsN works)
thetas <- thetas + (thetas * tweak_perc)
```

Details on how PsN/Pirana uses this feature:
```
 -degree = fraction
            Default not set. After updating the initial estimates in the output file,
            randomly perturb them by degree=fraction, i.e. change estimate to
            a value randomly chosen in the range estimate +/- estimate*fraction
            while respecting upper and lower boundaries, if set in the model file.
            Degree is set to 0.1, a 10% perturbation, when option tweak_inits is
            set in execute. EOF
```

## Feedback requested
The overall procedure for doing this is somewhat simple, though there remain a few outstanding questions that require scientist feedback:
 - How should we handle bounded `THETA` values?
    - By default, we think we should respect the original bounds. In other words, if a user wants to tweak the initial value by 20%, but this falls outside the bounds, the tweak for this parameter will be constrained to the bounds. i.e. this parameter might only be able to be jittered by 10%, while the rest are capable of fluctuating +/- 20%. In the event this happens, it could be good to warn the user, though feedback here would be appreciated.
    - *We have also considered* allowing the bounds to be altered to encapsulate the provided tweak percentage. Using the above example, if the sampled tweak percentage fell outside a bound, the bound would be adjusted to be, say, 10% larger than the tweaked initial value. The feedback we want here is as follows:
       - Should we allow for this at all via an argument?
       - If so, how should the bounds be adjusted? Is the above suggestion sufficient, or is there another mechanism that would make more sense?
 - How should we handle `FIXED` parameters? 
    - Should these values be skipped when tweaking the initial values always?
    - Should there be an argument that allows the user to choose between including/excluding these parameters?
 - Do we want nuanced control over tweaking various record blocks? Some things to think about:
[...]

PR #646, opened 2024-01-31
Get initial estimates and tweak initial theta estimates

barrettk:
See previous discussions here: https://github.com/metrumresearchgroup/bbr/pull/635

### Getting initial parameter estimates

The goal was to return an object that looked similar to `param_estimates()`. The internal function `get_initial_est()`, returns all values (i.e. the full diagonally concatenated `OMEGA` matrix if multiple records are provided) in a list format, whereas the wrapper `initial_estimates()`, only displays values defined in the control stream file
<details closed>
<summary>`initial_estimates()`  example</summary>

```r
MOD1 <- read_model(
 system.file("model", "nonmem", "basic", "1", package = "bbr")
)

> initial_estimates(MOD1, flag_fixed = TRUE)
# A tibble: 8 × 7
  parameter_names  init lower_bound upper_bound fixed record_type record_number
  <chr>           <dbl>       <dbl>       <dbl> <lgl> <chr>               <int>
1 THETA(1)         2              0          NA FALSE theta                   1
2 THETA(2)         3              0          NA FALSE theta                   1
3 THETA(3)        10              0          NA FALSE theta                   1
4 THETA(4)         0.02          NA          NA FALSE theta                   1
5 THETA(5)         1             NA          NA FALSE theta                   1
6 OMEGA(1,1)       0.05          NA          NA FALSE omega                   1
7 OMEGA(2,2)       0.2           NA          NA FALSE omega                   1
8 SIGMA(1,1)       1             NA          NA TRUE  sigma                   1
```

If you would like to format OMEGA or SIGMA records as full matrices, they are stored as attributes:
```r
 initial_est <- initial_estimates(.mod)
 attr(initial_est, "omega_mat")
 attr(initial_est, "sigma_mat")

# Run
> attr(initial_est, "omega_mat")
     [,1] [,2]
[1,] 0.05   NA
[2,]   NA  0.2
```
</details>

# Tweaking initial estimates

### Usage
```r
MOD1 <- read_model(
    system.file("model", "nonmem", "basic", "1", package = "bbr")
)

# Base usage
mod2 <- copy_model_from(MOD1, "mod2") %>%
  tweak_initial_estimates(.p = 0.2)

# This function may be paired with `inherit_param_estimates()`:
mod2 <- copy_model_from(MOD1, "mod2") %>%
  inherit_param_estimates() %>% tweak_initial_estimates(.p = 0.2)

```

<details closed>
<summary>Walkthrough</summary>

### Starting Record
```
;Initial THETAs
$THETA
(0,,1) ;[LCLM2]
( 0.7 FIX) ;[LCLM]
(0.67, 0.7, 0.72) ;[LCLF]
( 2 )   ;[CLAM]
( 2.0);[CLAF]
( 0.7 ) ;[LV1M]
( 0.7 ) ;[LV1F]
( 2.0 )   ;[V1AM]
( 2.0 )   ;[V1AF]
( 0.7 ) ;[MU_3]
(  0.7 );[MU_4]
( 0.3 )     ;[SDSL]
```

### Tweak values
```r
tweak_initial_estimates(
    .mod
    .p = 0.1,
    digits = 2
)
```

### After Tweaking
```
;Initial THETAs
$THETA
(0,,1) ;[LCLM2] # If no estimate is found, treat as fixed (skip)
( 0.7 FIX) ;[LCLM] # skip fixed parameters
(0.67, 0.72, 0.72) ;[LCLF] # If tweaked value would be outside of one of the bounds, set to the bound
( 1.9 )   ;[CLAM] # Otherwise tweak the value by the sampled percentage
( 2.1);[CLAF]
( 0.76 ) ;[LV1M]
( 0.67 ) ;[LV1F]
( 2.2 )   ;[V1AM]
( 1.8 )   ;[V1AF]
( 0.73 ) ;[MU_3]
(  0.68 );[MU_4]
( 0.27 )     ;[SDSL]
```
</details>

closes https://github.com/metrumresearchgroup/bbr/issues/632


tests/testthat/test-initial-estimates.R
```diff
@@ -0,0 +1,87 @@
...
+      length(omega_inits[,1])
+    )
+    expect_equal(
+      length(which(rec_nums == 2)),
+      # lower triangular matrix minus the first column
+      sum(lower.tri(omega_inits, diag = TRUE)[,-1])
+    )
+  })
+
+  it("initial_estimates: integration", {
```
seth127, line 63: I would like to have a couple more `initial_estimates()` tests on different models. Mostly to check that the matrices are the expected shape and that the number of specified parameters we parse are what we expect.

Some examples I think are worth testing:

* `inst/model/nonmem/complex/iovmm`
  * usage of `SAME` 
  * multiple `OMEGA` blocks, some of which will show up and some won't (i.e. `record_number` will have `1, 2, 4`)
  * `attr("omega_mat")` is the size we expect it to be
* `inst/model/nonmem/complex/example2_saemimp`
  * expect to see "This model appears to be using... to specify priors" warning
  * check the number of each type (to see that it does indeed parse those priors through)
  * check `record_number` column is what we expect (multiple THETA and OMEGA blocks)

This might be overkill, but there are a lot of moving pieces here and I think it's worth being sure that everything is being passed along the way we want it to.

R/tweak-initial-estimates.R
```diff
@@ -0,0 +1,192 @@
+
+#' Tweak the initial parameter estimates
+#'
+#' @param .mod model object to update.
+#' @param .p Percent to tweak the initial parameter estimates by. Represented as
+#'   a decimal.
+#' @param tweak type of estimates to tweak in the model. Only `"theta"` is
+#'   currently supported, though omega and sigma will be options at a later date.
```
seth127, line 8: `"omega"` and `"sigma"` (to match `"theta"` above)

R/tweak-initial-estimates.R
```diff
@@ -0,0 +1,192 @@
...
+
+  # Initialize formatted version of initial value and bounds
+  init_thetas_adj <- init_thetas %>% mutate(
+    new_fmt = sprintf(fmt_digits, .data$new),
+    up_fmt = sprintf(fmt_digits, .data$up),
+    low_fmt = sprintf(fmt_digits, .data$low)
+  )
+
+  # Set to `bound_perc` (e.g, 90%) of the way between init and bound
+  adjust_theta_bounds <- function(init_thetas, bound_perc = 0.9, fmt_digits){
```
seth127, line 143: Small thing, but I don't think we want a default value for `bound_perc`. We'll always be passing something to this and this default just feels like it's asking for mysterious unintended behavior.

R/tweak-initial-estimates.R
```diff
@@ -0,0 +1,192 @@
...
+#'  For bounded `THETA` estimates:
+#'   - If an initial `THETA` has bounds **and** an initial estimate
+#'  (e.g., `(0, 0.5, 1)`, `(0,1)`), the bounds will be respected when sampling
+#'  a percent to tweak by. If the tweaked value would fall below the lower bound,
+#'  the initial estimate will be set to 90% of the way between the initial value
+#'  and lower bound. If *after rounding*, the value would still not be within the
+#'  bounds, the difference between the initial value and the bound will be
+#'  iteratively reduced  until it is. The same is true for upper bounds.
+#'     - e.g., `(0, 0.5, 1)` --> tweak initially, falls outside bound (`(0, 1.2, 1)`)
+#'     --> set to value within bounds (first iteration: `(1-0.5)*0.9)`:  (`(0, 0.45, 1)`)
```
seth127, line 28: I'm pretty sure the end of this should say `(0, 0.95, 1)` not `(0, 0.45, 1)`. I was worried that we had implemented it wrong, but I ran through the code a bit and I think the code is right and these docs are just wrong. Let's talk through this together next week though.

seth127 (changes requested):
This is looking great overall. I have a bunch of comments, but they're almost all very small things.

The one thing that I want to look into deeper next week is the `adjust_theta_bounds` part. See my comment on this part of the docs above

```
#' --> set to value within bounds (first iteration: `(1-0.5)*0.9)`: (`(0, 0.45, 1)`)
```

I think we're actually good, but I want to walk through this with you to make sure we're on the same page with the algorithm that's been implemented and the test cases for it.

R/initial-estimates.R
```diff
@@ -0,0 +1,257 @@
...
+    omega_inits %>% dplyr::mutate(record_type = "omega"),
+    sigma_inits %>% dplyr::mutate(record_type = "sigma")
+  ) %>% dplyr::relocate("parameter_names") %>%
+    dplyr::relocate("record_number", .after = "record_type")
+
+  # Filter out NA values, as these were not specified in the control stream file
+  initial_est_df <- initial_est_df %>% dplyr::filter(!is.na(.data$init))
+
+  # Add original matrices as attributes if needed
+  attr(initial_est_df, "omega_mat") <- suppressWarnings(get_initial_est(.mod)) %>%
```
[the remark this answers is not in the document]
barrettk: I mostly did it as to only attach the matrix in this case (and not all the other attributes). I can rework this as to decrease the number of calls. The suppressWarnings was just to suppress `prior` warnings, that would have already been displayed during the first call.

R/initial-estimates.R
```diff
@@ -0,0 +1,257 @@
...
+
+  return(inits)
+}
+
+
+#' Helper function for getting the initial theta values, including bounds and
+#' the presence of `FIXED` parameters.
+#'
+#' @param ctl an `nmrec` ctl object.
+#' @param mark_flags TODO
```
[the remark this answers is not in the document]
barrettk: Good catch: never populated after copying over/updating `nmrec`

R/initial-estimates.R
```diff
@@ -0,0 +1,257 @@
...
+
+  return(initial_est_df)
+}
+
+
+#' Retrieve initial parameter estimates
+#'
+#' @inheritParams initial_estimates
+#'
+#' @keywords internal
```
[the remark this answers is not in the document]
barrettk: That sounds good for here. However, in some cases I do like generating helper docs for internal functions, in the event it makes it easier to use or recall how to specify things. Do you have any thoughts on still generating docs for *some* internal functions?

tests/testthat/test-initial-estimates.R
seth127: Sorry, my mistake here. I meant to say `"complex/acop-iov"` instead of `"complex/iovmm"` for the first example. Sorry about that.

Using your example code, it looks like this. Notice the `1, 2, 4` in the `record_number` column for OMEGAs (because it skips 3 and 5, which are `SAME` blocks).

```r
> mod_c <- read_model(file.path(MODEL_DIR_X, "acop-iov"))
> initial_est_df <- initial_estimates(mod_c)
> dplyr::select(initial_est_df, parameter_names, init, record_type, record_number)
# A tibble: 10 × 4
   parameter_names  init record_type record_number
   <chr>           <dbl> <chr>               <int>
 1 THETA(1)         2    theta                   1
 2 THETA(2)         3    theta                   1
 3 THETA(3)        10    theta                   1
 4 THETA(4)         0.02 theta                   1
 5 THETA(5)         1    theta                   1
 6 OMEGA(1,1)       0.05 omega                   1
 7 OMEGA(2,2)       0.2  omega                   1
 8 OMEGA(3,3)       0.2  omega                   2
 9 OMEGA(33,33)     0.2  omega                   4
10 SIGMA(1,1)       1    sigma                   1
```

R/initial-estimates.R
```diff
@@ -0,0 +1,257 @@
...
+  subroutine_recs <- nmrec::select_records(ctl, "sub")
+
+  # Check for PRIOR subroutines
+  if(length(subroutine_recs) > 0){
+    prior_subs <- purrr::map_lgl(subroutine_recs, function(prior_sub){
+      prior_sub_str <- prior_sub$get_lines()
+      # filter out comments
+      prior_sub_str <- gsub(";.*", "", prior_sub_str)
+      # look for priors
+      any(stringr::str_detect(prior_sub_str, "(?i)PRIOR"))
```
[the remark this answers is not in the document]
barrettk: @seth127 the `(?i)` denotes case-insensitivity in regex. Example:
```r
> stringr::str_detect(c("prior", "PRIOR", "Prior", "pRiOr"), "(?i)PRIOR")
[1] TRUE TRUE TRUE TRUE
```

R/tweak-initial-estimates.R
barrettk: Confirmed you were right, updated docs in [490f4ca](https://github.com/metrumresearchgroup/bbr/pull/646/commits/490f4ca2e472d796240576143c8698e1d3de343b)

seth127 (changes requested):
This is looking great. A couple tiny clean up comments, but I think we're good to merge very soon.

tests/testthat/test-tweak-initial-estimates.R
```diff
@@ -44,41 +44,12 @@ describe("tweak_initial_estimates", {
...
-    thetas_tweaked <- get_initial_est(mod_tweak_new, flag_fixed = TRUE)$thetas
-
-    # All values should be within bounds
-    init_within_bounds <- (thetas_tweaked$init < thetas_tweaked$up) &
-      (thetas_tweaked$init > thetas_tweaked$low)
-    expect_true(all(init_within_bounds[!is.na(init_within_bounds)]))
-  })
 
-  it("theta bounds - adjust_tweaked_theta()", {
+    # Note: the comments next to each theta record do not necessarily need to
```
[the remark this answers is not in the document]
barrettk: The comments I had next to each record were originally true when I was setting the seed and used 3 digits. After making that test a loop those comments were no longer true. I thought they could add some clarity to how the function was intended to work, but I agree it sounds confusing/misleading. Opted to remove them.

seth127 (approved):
Great work here.

PR #648, opened 2024-02-01
param_estimates_batch: don't rely on fread's header guess

### LINKED ISSUE #647: param_estimates_batch breaks with newest data.table release
In one of the ongoing PRs, drone [failed](https://github-drone.metrumrg.com/metrumresearchgroup/bbr/3797/3/4) when pulling from CRAN.

it seems to be related to how `fread` works [here](https://github.com/metrumresearchgroup/bbr/blob/main/R/param-estimates-batch.R#L81), and originates from the latest data.table [release](https://github.com/Rdatatable/data.table/releases/tag/1.15.0) (happened yesterday).

The actual error that happens with `data.table 1.15.0`:
```r
>     param_tbl <- BATCH_PARAM_TEST_DIR %>% param_estimates_batch()
Error in `rename()` at dplyr/R/mutate.R:146:3:                                                                                                                                     
! Can't rename columns that don't exist.
✖ Column `dir` doesn't exist.
Run `rlang::last_trace()` to see where the error occurred.
```

With `data.table 1.15.0`, and after running line 81 of `param-estimates-batch.R` (same as the link above), we end up with this data structure:
```r
> df <- as_tibble(fread(text = res$stdout)) # line 81
> df
# A tibble: 2 × 4
  V1                  V2                   V3            V4   
  <chr>               <chr>                <chr>         <lgl>
1 dir                 error                "termination" NA   
2 test_batch_params/7 no ext file contents ""            NA  
```

With the previous version of `data.table`, we would get this:
```r
> df <- as_tibble(fread(text = res$stdout)) # line 81
> df
# A tibble: 1 × 4
  dir                 error                termination V4   
  <chr>               <chr>                <lgl>       <lgl>
1 test_batch_params/7 no ext file contents NA          NA  
```

As you can see, the previous dataframe would fail when running [these lines](https://github.com/metrumresearchgroup/bbr/blob/main/R/param-estimates-batch.R#L90-L95) since `dir` is not a column name.

As an aside, Im not too sure what [this](https://github.com/metrumresearchgroup/bbr/blob/main/R/param-estimates-batch.R#L87) is doing (It didnt change the names in my testing, though perhaps it serves a purpose in other cases?)

kyleam:
param_estimates_batch() consumes `bbi nonmem params --dir ...` output. That has a header such as

    dir,error,termination,THETA1,[...],SIGMA(1_1),[...],OMEGA(2_2)

When feeding those lines to fread(), we do not specify that there is a header and instead let fread() guess.

The recent data.table 1.15.0 release includes a change to the guessing strategy: f83123f3 (fread improve containing header guess, 2021-11-30).  With that commit, two param_estimates_batch() test cases fail because fread() incorrectly guesses that there isn't a header. The input for those cases look like this:

    [1] "dir,error,termination,"
    [2] "dirname,msg,,"

Tell fread() that the input has a header to avoid the problem.

Closes #647.

### DIFF
--- a/R/param-estimates-batch.R
+++ b/R/param-estimates-batch.R
@@ -78,7 +78,7 @@ param_estimates_batch <- function(.path,
     }
   )
 
-  df <- as_tibble(fread(text = res$stdout))
+  df <- as_tibble(fread(text = res$stdout, header = TRUE))
 
   # throw out extra column that gets created if no models succeeded
   if ("V4" %in% names(df)) df <- select(df, -"V4")


barrettk (approved):
Looks good; thanks for doing this!

Issue #650, opened 2024-02-05
Function to run NMTRAN on a model object


barrettk:
This idea has been brought up numerous times, but the core idea is to provide a method for validating a `bbr` model object via `NMTRAN` before submitting the model. There are other applications for this as well, such as validating a recently tweaked model, or just ensuring `NONMEM` syntax is correct in general.

### Limitations
 - To use `NMTRAN`, `bbr` needs to know where the `NONMEM` executable is. We can determine this via a `bbi.yaml`, as well as via a user-provided `.config_path` (see `?submit_model` arguments), thought this doesn't necessarily match up with the version of NONMEM used when actually submitting the model. This because `bbi` determines the path a little differently.
 - This therefore requires that `bbi_init()` be called before validating the model. This requirement may limit what functions we can nest a `run_nmtran()` call into (such as the new (in-flight) `tweak_initial_estimates` function).

Below is a prototype scaffold of how we can determine the `NMTRAN` executable path:
<details closed>
<summary>Prototype functions</summary>

```r

locate_nmtran <- function(.mod, .config_path = NULL, nmtran_exe = NULL){
  
  if(is.null(nmtran_exe)){
    model_dir <- get_model_working_directory(.mod)
    config_path <- .config_path %||% file.path(model_dir, "bbi.yaml")
    
    if(!file_exists(config_path)){
      stop(paste("No bbi configuration was found in the execution directory.",
                 "Please run `bbi_init()` with the appropriate directory to continue."))
    }
    
    if (!is.null(.config_path)) {
      config_path <- normalizePath(.config_path)
    }
    
    bbi_config <- yaml::read_yaml(config_path)
    nm_config <- bbi_config$nonmem
    
    # look for default nonmem installation
    default_nm <- purrr::keep(nm_config, function(nm_ver){
      !is.null(nm_ver$default)
    })
    
    # Set nonmem path
    if(length(default_nm) > 0){
      default_nm <- default_nm[[1]]
    }else{
      # If no default, use the last one (likely higher version)
      default_nm <- nm_config[[length(nm_config)]]
    }
    
    # Set NMTRAN executable path
    nm_path <- default_nm$home
    nmtran_exe <- file.path(nm_path, "tr", "NMTRAN.exe")
  }
  
  if(!file_exists(nmtran_exe)){
    stop(glue("Could not find an NMTRAN executable at `{nmtran_exe}`"))
  }
  
  return(nmtran_exe)
}

[...]

PR #652, opened 2024-02-06
Function to run `NMTRAN` on a model

barrettk:
### Args
```r
run_nmtran(
  .mod,
  .config_path = NULL,
  nmtran_exe = NULL, # Path to an NMTRAN executable. If NULL, will look for a bbi.yaml file in the same directory as the model.
  delete_on_exit = TRUE # Logical. If FALSE, don't delete the temporary folder containing the NMTRAN run.
)
```

### Example
```r
> .mod <- MOD1
> run_nmtran(.mod)

── Status ──────────────────────────────────────────────────────────────────────────────────────────────────

── NMTRAN successful ──

── Absolute Model Path ─────────────────────────────────────────────────────────────────────────────────────
• /data/Projects/package_dev/bbr/inst/model/nonmem/basic/1.ctl

── NMTRAN Specifications ───────────────────────────────────────────────────────────────────────────────────
• NMTRAN Executable: /opt/NONMEM/nm75/tr/NMTRAN.exe
• NONMEM Version: "nm75"
• Run Directory: /data/Projects/package_dev/bbr/nmtran_1_RtmpMzE7cx

── Output ──────────────────────────────────────────────────────────────────────────────────────────────────
  
 WARNINGS AND ERRORS (IF ANY) FOR PROBLEM    1
             
 (WARNING  2) NM-TRAN INFERS THAT THE DATA ARE POPULATION.
  
Note: Analytical 2nd Derivatives are constructed in FSUBS but are never used.
      You may insert $ABBR DERIV2=NO after the first $PROB to save FSUBS construction and compilation time
  
Process finished.
```

closes https://github.com/metrumresearchgroup/bbr/issues/650


barrettk:
@seth127 when you get a second, let me know what you think of the [compare_nmtran](https://github.com/metrumresearchgroup/bbr/blob/19f25b8ecb8bb488a85cd8b811fce65162a3b26c/R/run-nmtran.R#L212-L327) function. I assume we will ditch it, but i'd like to keep that function handy during some development tasks that involve touching control stream files. I suppose we could export it, though I dont think that's too in-line with the original request. It is comparable to the bash script @kyleam was running though

seth127:
This is looking good to me, but I'll leave the final review for @andersone1

seth127:
@kyleam could you take a look at how we're [building the path to the NMTRAN executable](https://github.com/metrumresearchgroup/bbr/blob/8d4a98f0ad6e3824dd619a42cdb778aedc81562c/R/run-nmtran.R#L121-L125) and make any comment on whether we think this will work on Windows. And if not... do we know a better pattern for that?

Related, we [execute it with `processx`](https://github.com/metrumresearchgroup/bbr/blob/8d4a98f0ad6e3824dd619a42cdb778aedc81562c/R/run-nmtran.R#L150-L153). I think this will work, regardless of OS, assuming we have the `nmtran_exe` set correctly, but I wanted to double-check that with you too.

kyleam:
> @kyleam could you take a look at how we're [building the path to the NMTRAN executable](https://github.com/metrumresearchgroup/bbr/blob/8d4a98f0ad6e3824dd619a42cdb778aedc81562c/R/run-nmtran.R#L121-L125) and make any comment on whether we think this will work on Windows.

Looking just at those linked lines, nothing jumps out at me that will be an issue.  

That's assuming all the processing outside those lines is correct on Windows, but the best option would of course just be to try it out.  Before this is merged, I can test the command on our Windows instance for bbr/bbi testing.  Or perhaps better: I can get @andersone1 and @barrettk set up to access.  Let me know what you all prefer.

> Related, we [execute it with `processx`](https://github.com/metrumresearchgroup/bbr/blob/8d4a98f0ad6e3824dd619a42cdb778aedc81562c/R/run-nmtran.R#L150-L153). I think this will work, regardless of OS, assuming we have the `nmtran_exe` set correctly, but I wanted to double-check that with you too.]

Right. processx in general supports Windows.

kyleam:
I'll pull my comment from https://github.com/metrumresearchgroup/bbr/issues/650#issuecomment-1930725642 for consideration:

 
> Based on how `nmfe*` invokes `NMTRAN.exe`...
> 
> ```
> $ grep tr/NMTRAN /opt/NONMEM/nm75/run/nmfe75
> # $dir/tr/NMTRAN.exe $prdefault $tprdefault $maxlim < $1 >& FMSG
>   $dir/util/nmtran_presort < tempzzzz1 | $dir/tr/NMTRAN.exe $prdefault $tprdefault $maxlim $do2test >& FMSG
>   $dir/util/nmtran_presort < tempzzzz1 | $dir/tr/NMTRAN.exe $prdefault $tprdefault $maxlim >& FMSG
> ```
> 
> ... I think it's worth considering these questions:
> 
> * What does `nmtran_presort` do?  Should we be processing the input with that too to follow `nmfe*`?
> * `nmfe*` relays some arguments to `NMTRAN.exe`.  Is that something that something this check should do too?  Are there cases where this new NM-TRAN check would fail but the actual run would pass due to this check not passing the same args to the `NMTRAN.exe` call?
> 
> The answer may be that none of that matters in the context of this check, but that'd still be good to document here.

seth127:
@barrettk @andersone1 see both comments from @kyleam above. Could the two of you coordinate to do the following:

1. Connect with @kyleam to get access to the Windows instance he mentions, so that we can give this a test run there. (Probably best for both of you to have access to that, honestly.)
2. Someone check into the `presort` and flags thing that he mentions in the most recent comment.

Thanks all. I'm looking forward to getting this feature out there.

barrettk:
Addressing the conflicts via a rebase ended up not being worth the trouble, so opened a new PR [here](https://github.com/metrumresearchgroup/bbr/pull/705)

PR #654, opened 2024-02-07
Prepare for 1.9.0 Release

barrettk:
Changeset: https://github.com/metrumresearchgroup/bbr/compare/1.8.1...release/1.9.0

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -34,8 +34,8 @@ steps:
   image: 906087756158.dkr.ecr.us-east-1.amazonaws.com/r-dev-ci-mpn-4.0:latest
   commands:
   - R -s -e 'devtools::install_deps(upgrade = '"'"'never'"'"')'
-    # this can be removed once nmrec 0.3.0.8001 is available on MPN
-  - git clone --branch 0.3.0.8001 --depth 1 https://github.com/metrumresearchgroup/nmrec.git /tmp/nmrec
+    # this can be removed once nmrec 0.4.0 is available on MPN
+  - git clone --branch 0.4.0 --depth 1 https://github.com/metrumresearchgroup/nmrec.git /tmp/nmrec
   - R -s -e 'devtools::install("/tmp/nmrec", upgrade = "never")'
   - R -s -e 'devtools::check(env_vars = c("BBI_EXE_PATH" = "/ephemeral/bbi", "NOT_CRAN" = "true"))'
   environment:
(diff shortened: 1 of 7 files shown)


barrettk:
Apparently [drone fails](https://github-drone.metrumrg.com/metrumresearchgroup/bbr/3844/2/4) here (MPN oldest):
```
══ testthat results  ═══════════════════════════════════════════════════════════
--
242 | [ OK: 1013 \| SKIPPED: 16 \| WARNINGS: 17 \| FAILED: 1 ]
243 | 1. Error: summary_log() works correctly with nested dirs [BBR-SMLG-002] (@test-summary-log.R#42)
```

Not sure what's going on yet

kyleam:
Hmm, that's an odd one, given build was green at time those summary_log came in.  I can take a look in a bit if it doesn't jump out to you in the meantime.

https://github-drone.metrumrg.com/metrumresearchgroup/bbr/3810

barrettk:
Yeah I was just looking through [your PR](https://github.com/metrumresearchgroup/bbr/pull/645) for exactly that. Definitely odd

kyleam:
Huh, PR build passed, so I'd first see if it is there on a restart:

image

barrettk:
>Huh, PR build passed for this PR, so I'd first see if it is there on a restart

Interestingly, that was failing too until I restarted it. I hadn't restarted the push CI, but will give that a try too

barrettk:
### Scorecard Details
image

.drone.yml
```diff
@@ -107,7 +107,7 @@ steps:
   commands:
   - R -s -e 'devtools::install_deps(upgrade = '"'"'never'"'"')'
   # nmrec isn't available until the 2023-09-19 snapshot.
-  - git clone --branch 0.3.0.8001 --depth 1 https://github.com/metrumresearchgroup/nmrec.git /tmp/nmrec
+  - git clone --branch 0.4.0 --depth 1 https://github.com/metrumresearchgroup/nmrec.git /tmp/nmrec
```
kyleam, line 110: I think this is okay, but as a side remark: in the spirit of the "oldest" build, I suppose it might make sense to eventually downgrade this to the nmrec version that's on the 2023-09-19 snapshot

seth127: I could go either way on this. I don't think it practically matters at this point, and that would make it skip some tests. So I think I'd opt for keeping it this way for now.

seth127 (approved):
All looks good to me. `R CMD CHECK` passes for me on Metworx too, so we'll just assume those CI failures we're transient Drone failures.

Thanks for pulling these changes together @barrettk and @kyleam !

.drone.yml
kyleam: > I could go either way on this. I don't think it practically matters at this point, and that would make it skip some tests.

That's true, the from my standpoint that's a good thing.  The other runs already cover this, and this run would tell you that you set up your skips properly so that your tests work with an older nmrec.

I agree that's not super valuable info, especially at this point.

>  So I think I'd opt for keeping it this way for now.

Sure, I agree there.  I should of probably rephrased my comment to make it really clear I wasn't hoping for this PR to change this.  (Holding off on approving as I do a scorecard run.)

kyleam:
> so we'll just assume those CI failures we're transient Drone failures.

Yes, fingers crossed that those don't start popping up more frequently.  I'm very pleased that we haven't seen many (if any) rare-but-regular-enough-to-be-painful transient failures in our packages.

kyleam (approved):
Thanks @barrettk.  Looks good to me too.

Issue #640, opened 2024-01-18
tweak_initial_estimates: Matrix handling


barrettk:
During development of the `tweak_inital_estimates` [PR](https://github.com/metrumresearchgroup/bbr/pull/635), we determined that some additional checks and modification is required for matrix-type records (`OMEGA` and `SIGMA`).

For one, we have to check that jittered values still result in a positive-definite matrix. Altering off-diagonal values seems to increase the likelihood of this, but doesn't make the problem any more difficult to solve.

Proposal:
 - We determine if the final tweaked matrix is positive-definite. If not, pass it through a _variation_ of `Matrix::nearPD()`. The exact function may not work, as the matrix passed in to `nmrec::set_omega()` utilizes `NA` values to denote 'skipping' that index (i.e. dont overwrite it). It also may denote that value was not specified in the control stream file (most common scenario).

Other questions:
 - Does this change for correlation matrices? As in, do we need some other check, rather than a positive-definite for correlation matrices?
 - We currently 'skip' `FIXED` values or `block(n)` records. Do we want this handling for any other flag commonly found in matrix-type records?

### Examples to think about
```r
$omega 1 sd fix 2
$theta 10 11
$omega block(2) fix
3
0.1 cor 4
$omega block(2)
3 fix
0.1 uni 4
```

barrettk:
FYI It was determined that we do in fact need additional handling for `cor` matrices. In this case, the diagonals are variances, and off-diagonal values specify correlation. In these instances, we first have to turn the matrix into a correlation or covariance matrix, check for positive-definiteness/making necessary adjustments, and then convert it back to the correct format (variance on diagonal, correlation on off-diagonal)

barrettk:
We also determined that it would not be possible to check for positive-definiteness using the full matrix in the example below:

```r
> nmrec::extract_omega(ctl)
      [,1]  [,2]  [,3] [,4] [,5] [,6] [,7] [,8]
[1,] 0.500    NA    NA   NA   NA   NA   NA   NA
[2,] 0.001 0.500    NA   NA   NA   NA   NA   NA
[3,] 0.001 0.001 0.500   NA   NA   NA   NA   NA
[4,] 0.001 0.001 0.001  0.5   NA   NA   NA   NA
[5,]    NA    NA    NA   NA 0.01   NA   NA   NA
[6,]    NA    NA    NA   NA 0.00 0.01   NA   NA
[...]

PR #656, opened 2024-02-08
tweak_initial_estimates: matrix handling

barrettk:
Handling for matrix-type records. Specifically check for positive-definiteness and include handling for different formats of OMEGA and SIGMA matrices, such as:
 - `VARIANCE` (default) or` STANDARD` controls whether diagonals are taken as variance or standard deviation, respectively
 - `COVARIANCE` (default) or `CORRELATION` controls whether the off-diagonals are taken as covariance or correlations, respectively
 - `CHOLESKY` representations
 - handling for `SAME` blocks and records defined like the one below:
    ```r 
    $OMEGA BLOCK(6)
    0.1
    0.01 0.1
    (0.01)x2 0.1
    (0.01)x3 0.1
    (0.01)x4 0.1
    (0.01)x5 0.1
    ```
   - Unclear if this will be supported vs error out at the moment.
 

closes https://github.com/metrumresearchgroup/bbr/issues/640


seth127:
After discussing today, @barrettk is going to open a new branch that cleans this up and focuses on the `collapsibleTree` method

PR #657, opened 2024-02-12
check_run_times docs: fix .return_times typo in example

kyleam:
A check_run_times() example passes "All" for `.return_times`, but the function expects "all".


seth127 (approved):
good catch

PR #661, opened 2024-02-13
tail_*: don't hard code target when passed 'model path'

kyleam:
`tail_output.character` and `tail_lst.character` aren't compatible with nmbayes models.  The second commit fixes that.  

The first commit adds an argument check for various `*.character` methods in this file.  It's not conceptually related to the second commit, but I'm including it as cleanup here to avoid a merge conflict.

### DIFF
--- a/R/read-output.R
+++ b/R/read-output.R
@@ -12,6 +12,7 @@
 #' @importFrom readr read_lines
 #' @export
 check_file <- function(.file, .head = 3, .tail = 5, .print = TRUE, .return = FALSE) {
+  checkmate::assert_string(.file)
   l <- read_lines(.file)
 
   l_len <- length(l)
--- a/R/read-output.R
+++ b/R/read-output.R
@@ -70,6 +70,7 @@ tail_output <- function(.mod, ...) {
 #' @describeIn check_file Tail the OUTPUT file from a file path to a model.
 #' @export
 tail_output.character <- function(.mod, .head = 3, .tail = 5, .print = TRUE, .return = FALSE, ...) {
+  checkmate::assert_string(.mod)
   # if model path passed, construct path
   if (basename(.mod) != "OUTPUT") {
     .mod = as.character(file.path(.mod, "OUTPUT"))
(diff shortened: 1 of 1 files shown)


kyleam:
Here's where this is tested in bbr.bayes (https://github.com/metrumresearchgroup/bbr.bayes/pull/127):

 * https://github.com/metrumresearchgroup/bbr.bayes/blob/6a21da1b22e979edceddfdf8c47b45618b7372b3/tests/testthat/test-nmbayes-tail.R#L43-L50

 * https://github.com/metrumresearchgroup/bbr.bayes/blob/6a21da1b22e979edceddfdf8c47b45618b7372b3/tests/testthat/test-nmbayes-tail.R#L126-L133

seth127:
@kyleam similar to changes in 66f81caaf281e92a194af26f056da5555810fb43, should we also add `checkmate::assert_string()` [here in `check_file()`](https://github.com/metrumresearchgroup/bbr/blob/7606646090d8022d20c8559b7277c78bd6a43e4f/R/read-output.R#L14-L15)? In my quick testing, `readr::read_lines()` will take a vector of files and just concatenate what it reads in, but that seems likely to surprise or confuse a user. And we say in [the docs](https://github.com/metrumresearchgroup/bbr/blob/7606646090d8022d20c8559b7277c78bd6a43e4f/R/read-output.R#L7) that we expect a scalar (i.e. string). 

Or do you think that would technically qualify as a "breaking change", in the spirit of [Hyrum's Law](https://www.hyrumslaw.com/), despite the fact that it would actually make it consistent with what's documented?

As a sidenote, I referenced Hyrum's Law semi-humorously, but the first clause is

> With a sufficient number of users of an API...

which probably doesn't apply to this function at this point, though that's obviously a judgement call.

seth127 (approved):
Approving this so that I don't hold up https://github.com/metrumresearchgroup/bbr.bayes/pull/127 unnecessarily, but I do want to address my [comment above](https://github.com/metrumresearchgroup/bbr/pull/661#issuecomment-1947374037). That said, adding those guards is tangential to this PR anyway, so I'm fine moving ahead either way.

kyleam:
> @kyleam similar to changes in [66f81ca](https://github.com/metrumresearchgroup/bbr/commit/66f81caaf281e92a194af26f056da5555810fb43), should we also add `checkmate::assert_string()` [here in `check_file()`](https://github.com/metrumresearchgroup/bbr/blob/7606646090d8022d20c8559b7277c78bd6a43e4f/R/read-output.R#L14-L15)?

Good point, I was focused on the character methods and didn't consider `check_file`.  I'm in favor of adding it there too.

> which probably doesn't apply to this function at this point

Yes, I agree (though a [longtimeuser4](https://xkcd.com/1172/) may not).

PR #662, opened 2024-02-14
nm_join: abort and point to nm_join_bayes if passed nmbayes model

kyleam:
nm_join() is incompatible with nmbayes models (it will fail looking for the files in the top-level output directory).

bbr in general avoids embedding information about bbr.bayes, but there are already some exceptions.  Make one here too to point callers in the right direction.

Re: https://github.com/metrumresearchgroup/bbr.bayes/issues/36

### DIFF
--- a/R/nm-join.R
+++ b/R/nm-join.R
@@ -92,6 +92,12 @@ nm_join <- function(
     no_shk_file = TRUE
   )
 ) {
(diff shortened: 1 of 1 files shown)


seth127 (approved):
How does it know? It's magic.

(I'm kidding, this is exactly what we discussed in [bbr.bayes#36](https://github.com/metrumresearchgroup/bbr.bayes/issues/36))

Issue #651, opened 2024-02-06
Feature request: make `get_data_path` also read in the path from the ctl


barrettk:
As seen in the helper docs, `get_data_path`:
> **only works on models that have been successfully run** because the path to the data is constructed during model submission.

I propose an amendment to this function, though I am **not** proposing we replace the existing functionality with the one below (more on this in a second).

We can get the data path from the `ctl` file directly via `nmrec`:
```r
get_data_path_from_ctl <- function(.mod){
  mod_path <- get_model_path(.mod)
  ctl <- nmrec::read_ctl(mod_path)
  data_rec <- nmrec::select_records(ctl, "data")[[1]]
  data_path <- nmrec::get_record_option(data_rec, "filename")$value

  data_path_norm <- fs::path_norm(file.path(mod_path, data_path))

  if(!fs::file_exists(data_path_norm)){
    stop(glue("Could not find data at {data_path_norm}"))
  }

  return(data_path_norm)
}
```

I propose we attempt to fetch *both* data paths. i.e. we extract the data path from the `ctl` file, and *if* the model has been run, we also parse it from the `bbr` model object. We could compare the two, and warn the user if there is a mismatch (i.e. if the model file has been moved since it was run). We could have an argument for which one to return (with a default), to ensure a single string is usually returned.

Im not fully on board with this suggestion, but I think this could have a use case. The main thing I would change, is that we provide a method for *not* erroring out if the model has *not* been executed. If the model hasn't been executed, we could return the path defined in the control stream file.
cc @seth127

seth127:
I like this idea, and I've been mulling over something similar since filing https://github.com/metrumresearchgroup/bbr/issues/595 last year. Basically, it seems like we should have an easy way to get the path to the data (and possibly load the data) for a model that hasn't been run yet. nmrec provides an obvious solution for this. 

In terms of checking it against what's in bbi_config.json (for an already run model), that probably requires a bit more discussion. It might be "scope creep" for a function that's just supposed to return a file path, but if there's a compelling use case or reason, I'm open to it. 

Thanks for filing this.

seth127:
# Specification

Capturing some offline discussion of the changes we want to make to `get_data_path()`

* [ ] `get_data_path()` should return absolute path to the input data file (as it [currently does](https://metrumresearchgroup.github.io/bbr/reference/get_data_path.html))
[...]

Issue #595, opened 2023-05-04
nm_data doesn't work on unfinished models


seth127:
`nm_data()` doesn't work on a model that hasn't been run successfully (because it relies on [pulling the data path from the json config](https://github.com/metrumresearchgroup/bbr/blob/main/R/get-path-from-object.R#L179), which won't be there if the model hasn't be run successfully). 

Importantly, this also means `nm_join()` and friends don't work, because they call `nm_data()`.

seth127:
This just came up again and I agree that using nmrec to get this from the control stream seems like a good move. 

In terms of my comment from almost a year ago about nm_join() failing because of this... I'm not sure what I was getting there. While true, it's not obvious why a user would expect to be able to call nm_join() on a model that hasn't been run, notably because the table files wouldn't exist yet.

PR #664, opened 2024-02-15
get_data_path refactor

barrettk:
- extract and format data path using input data path set in the control stream file, rather than the `bbi.json` generated on a successful NONMEM model submission
 - extracting the path from the json is still supported via the `pull_from_config` argument
 - added support for extracting the data paths from a run_log

closes https://github.com/metrumresearchgroup/bbr/issues/651
closes https://github.com/metrumresearchgroup/bbr/issues/595


barrettk:
The following function and test changes were required due to the most recent refactor of `get_data_path`:
 - `check_up_to_date`:
    - The internal `check_up_to_date_nonmem` function just had to specify `.check_exists = FALSE`. We dont want to check the data path's existence here since `check_up_to_date` has its own handling/error messaging for when the data path has changed
    - Test `BBR-CUTD-004` was changed. Previously an empty `ctl` file was used for the test. This is now allowed whenever `nmrec` is used. I followed the example I had done previously, where I passed in model lines to `mod_content` in the `create_temp_model` function. That was the only change needed for this test, and probably should have been done before.
 - `test-nm-join.R `: `BBR-NMJ-007`
    - A couple things were changed here, though only one was really necessary. The rest just improved clarity since I was already changing the test.
    - The core change was the setup of the tests. To test duplicate rows, the model and data files were copied to a temporary directory, and an additional row was added. When copying the files to a temporary directory however, the data path was changed in the `bbi.json` file. This was relevant previously since `get_data_path` pulled the data path from there, rather than the control stream file. Since we now pull from the control stream file, we had to modify the relative data path there, rather than in the json file.
    - The other changes I made include adding comments to denote what was actually being tested, along with variable specification and formatting to improve code readability.


cc @seth127 @kyleam

barrettk:
Have to say, it is somewhat frustrating that we can't use a lot of the `rlang` and `cli` warning messages (see build errors on MPN:oldest [here](https://github-drone.metrumrg.com/metrumresearchgroup/bbr/3895/2/4)). I would very much be in favor in requiring minimum `rlang` and `cli` versions and bumping MPN:oldest in the near future, as I think the better formatted messages can go a long way

barrettk:
FYI @seth127 
> This should first check if the model has been finished running, and if not, skip this check altogether

I didn't implement it this way. I opted to error out here rather than skip it. Skipping it (without warning) when a user specified `pull_from_config = TRUE` didn't seem right, and I already had a notable amount of warnings, notes, and errors for various cases. Happy to discuss changing this, but just wanted to highlight why I went this way.

barrettk:
@seth127 as part of your review, can you specifically confirm that this PR addresses https://github.com/metrumresearchgroup/bbr/issues/595 appropriately as well? I didnt add any new tests for `nm_data`, but we could get rid of a lot of the `skip_` helper functions in the test suite. I can do that now or after you make a first pass review.

R/check-up-to-date.R
```diff
@@ -112,7 +112,7 @@ check_up_to_date_nonmem <- function(.mod) {
 
   # check necessary files for changes
   model_file <- get_model_path(.mod)
-  data_file <- get_data_path(.mod)
+  data_file <- get_data_path(.mod, .check_exists = FALSE)
```
kyleam, line 115: In the commit message, you say:

> I dont think it should pull from the config here either, as we want to interpret the currently specified path, rather than what it was when the model was submitted.

The model check already takes care of reporting if the data path configured in the model was changed. I don't see a strong reason to change this, and it seems to me that "did the data that was submitted change" is the core question this should be asking. In which case, the value recorded in the config would be the one to consider.

It is unlikely to matter in practice, but in such cases I'd prefer to stick to the current behavior and avoid unnecessary changes.

R/get-path-from-object.R
```diff
@@ -158,32 +158,188 @@ get_model_id.bbi_model <- function(.mod) {
 
 #' Get path to data file
 #'
-#' Helper to extract the path to the input data from a
-#' `bbi_{.model_type}_summary` or `bbi_{.model_type}_model` object. Note that
-#' **this function only works on models that have been successfully run**
-#' because the path to the data is constructed during model submission.
-#' @param .mod Model to use, either a `bbi_{.model_type}_model` or
-#' `bbi_{.model_type}_summary` object.
+#' Helper to extract the path to the input data from a `NONMEM` control stream file.
```
kyleam, line 161: Given this is a generic, this seems too implementation specific for the first sentence of the description. I think the previous description was more appropriate. You could then follow up with more specifics about NONMEM models.

R/get-path-from-object.R
```diff
@@ -158,32 +158,188 @@ get_model_id.bbi_model <- function(.mod) {
...
-#' @importFrom fs path_rel
+#'
+#' @details
+#'
+#' ### Notes on model extensions:
+#' `bbi` has different handling for `.mod` vs `.ctl` extensions for `NONMEM`
+#' models. When a model such as `001.ctl` is targeted with `bbi`, `bbi` will
+#' create a new directory called `001`, copy the model into it (with a `.mod`
+#' extension), and do the execution directory. For more details, see `bbi`'s
+#' [documentation](https://github.com/metrumresearchgroup/bbi/blob/main/integration/markdown/expectations.md?plain=1#L19:L21)
```
kyleam, line 180: I don't see where this explains the difference between handling for `.mod` and `.ctl` files. It mentions bbi creating the output directory and copying the file into it, but that's the same in either case.

The difference is that on copy bbi adjusts a relative data path for `.mod` files, adding on a "../". Or, said from the user perspective, writers of `.ctl` files are expected to specify a relative path for the data file one level deeper than the control stream they're editing.

This motivation for the `.mod` adjustment was apparently inspired by the PsN behavior. (See [this](https://github.com/metrumresearchgroup/bbi/blob/835953efed10fff2d7024ac1c3dd88949b268c05/cmd/nonmem.go#L981-L984) bbi comment.)

Also, I wonder whether `submit_model()` is really the more useful place to document this handling (it can then be referenced from here). Unlike `get_data_path()`, that's an entry point we know a user will go through.

R/get-path-from-object.R
```diff
@@ -158,32 +158,188 @@ get_model_id.bbi_model <- function(.mod) {
...
+get_data_path.bbi_log_df <- function(
+    .bbi_object,
+    .check_exists = TRUE,
+    pull_from_config = FALSE,
+    ...
+){
+  data_paths <- map_chr(.bbi_object[[ABS_MOD_PATH]], function(.path) {
+    .mod <- read_model(.path)
+    get_data_path_nonmem(.mod, .check_exists, pull_from_config, ...)
+  })
```
kyleam, line 227: This should call the generic method, not `get_data_path_nonmem()`, because the log can consist of models that aren't `bbi_nonmem_model`.

R/get-path-from-object.R
```diff
@@ -158,32 +158,188 @@ get_model_id.bbi_model <- function(.mod) {
...
+
+  return(data_paths)
+}
+
+
+#' @keywords internal
```
kyleam, line 233: No need to mark this as `internal` given that there isn't a docstring.

R/get-path-from-object.R
```diff
@@ -158,32 +158,188 @@ get_model_id.bbi_model <- function(.mod) {
...
+get_data_path_nonmem <- function(
+    .bbi_object,
+    .check_exists = TRUE,
+    pull_from_config = FALSE,
+    ...
+){
+
+  if(inherits(.bbi_object, NM_SUM_CLASS)){
+    .bbi_object <- read_model(.bbi_object$absolute_model_path)
+  }
```
kyleam, line 243: This seems like a better fit for the `get_data_path.bbi_nonmem_summary()` method.

R/get-path-from-object.R
```diff
@@ -158,32 +158,188 @@ get_model_id.bbi_model <- function(.mod) {
...
+  check_model_object(.bbi_object, "bbi_nonmem_model")
```
kyleam, line 244: This check seems unnecessary. This is called by methods that are already triggered by object type.

R/get-path-from-object.R
```diff
@@ -158,32 +158,188 @@ get_model_id.bbi_model <- function(.mod) {
...
+  mod_path <- get_model_path(.bbi_object)
+
+  # Get relative data path
+  data_path_rel <- get_data_path_from_ctl(.bbi_object)
```
kyleam, line 248: I'm confused by the `_path_rel` name. Isn't this coming straight from the control stream, so it may be either absolute or relative?

R/get-path-from-object.R
```diff
@@ -158,32 +158,188 @@ get_model_id.bbi_model <- function(.mod) {
...
+
+  # Handling for `.mod` extensions
```
kyleam, line 250: This comment says `.mod` but then the code is checking for `.ctl`, which is confusing to me. Worth clarifying, if you keep it, but in my opinion, you could just drop this comment. It's approaching the redundancy/noise of

```
# Read the lines.
lines <- readLines("foo")
```

R/get-path-from-object.R
```diff
@@ -158,32 +158,188 @@ get_model_id.bbi_model <- function(.mod) {
...
+  if(grepl("(?i)ctl", fs::path_ext(mod_path))){
```
kyleam, line 251: `get_model_path()` uses `find_nonmem_model_file_path()` underneath, and that's case sensitive, so there's no reason to be case insensitive here and this condition could just be

```
identical(fs::path_ext("mod_path"), "ctl")
```

(Same applies to line 269 below.)

R/get-path-from-object.R
```diff
@@ -158,32 +158,188 @@ get_model_id.bbi_model <- function(.mod) {
...
+    path_elements <- unlist(strsplit(data_path_rel, "/"))
```
kyleam, line 252: This assumes a "/" separator. On Windows, can NONMEM users write control streams with "\\" path separators?

Could use `fs::path_split()` here.

R/get-path-from-object.R
```diff
@@ -158,32 +158,188 @@ get_model_id.bbi_model <- function(.mod) {
...
+        c(
+          "x" = "Input data file does not exist or cannot be opened",
+          "i" = glue("Referenced input data path: {data_path_rel}"),
+          msg_extra
+        )
+      )
+    }
+  }
+
+  if(isTRUE(pull_from_config)){
```
kyleam, line 296: I don't understand the rationale for doing both of these and comparing, rather than just choosing one code path based on whether `pull_from_config = TRUE`.

R/get-path-from-object.R
```diff
@@ -158,32 +158,188 @@ get_model_id.bbi_model <- function(.mod) {
...
+      return(data_path_config)
+    }
+  }
+
+  return(data_path)
+}
+
+
+
+#' Get the relative data path from a control stream file
```
kyleam, line 316: It's not necessarily relative.

R/get-path-from-object.R
```diff
@@ -158,32 +158,188 @@ get_model_id.bbi_model <- function(.mod) {
...
+#' @noRd
+get_data_path_from_ctl <- function(.mod){
+  check_model_object(.mod, "bbi_nonmem_model")
+  mod_path <- get_model_path(.mod)
+  ctl <- nmrec::read_ctl(mod_path)
+
+  # Get data record
+  data_rec <- nmrec::select_records(ctl, "data")[[1]]
```
kyleam, line 324: This should verify that there is one (and only one) data record.

R/get-path-from-object.R
```diff
@@ -158,32 +158,188 @@ get_model_id.bbi_model <- function(.mod) {
...
+  data_path <- nmrec::get_record_option(data_rec, "filename")$value
```
kyleam, line 325: The data file in the NONMEM control stream may be quoted. This should strip those.

kyleam (changes requested):
Initial round of comments, focusing on the `get-path-from-object.R` changes


## Overview of breaking changes

Here is the potential breakage I see with this. For this comment, I'm trying to catalog and give an initial assessment of how likely I think those changes are to matter in practice; I'm leaving any opinions about what, if anything, we should do about these to a follow-up comment.

-   now fails if data file doesn't exist
    
    To align with other functions in `get-path-from-object.R`, this function now how as a `.check_exists` argument that defaults to `TRUE`.
    
    In practice, this change is probably unlikely to be noticed because 1) it's probably pretty rare for a data file that was already used as input to a finished model to be moved or deleted, and 2) users are likely calling `get_data_path()` to then read in the path, so they're going to hit into a failure then anyway.

-   argument name change
    
    The first argument of `get_data_path()` has been renamed from `.mod` to `.bbi_object` (I assume for consistency with other function in `get-path-from-object.R`).
    
    This change will break any callers existing callers that happen to call `get_data_path()` with a named argument.
    
    The old signature was of the form `fn(x, ...)`. In my estimation, it's pretty rare to see `fn(x = "blah")` calls in such a case.

-   any unforeseen nmrec breakage
    
    This function now relies heavily on nmrec, meaning that it inherits any unsupported features and any bugs in parsing. Bugs are of course an unknown.  For unsupported features/specifications, the spots that matter in this case are at the level of extracting *unparsed* records from the control stream or parsing the data record.
    
    The main unsupported feature I can think of here is multiple problem records. Given that bbi itself [doesn't handle that](https://github.com/metrumresearchgroup/bbi/issues/187), I think that isn't a concern here.

-   narrowing of method
    
    This narrowed `get_data_path.bbi_model()` to `get_data_path.bbi_nonmem_model()`. Conceptually that makes sense because you're now depending on the existence of a NONMEM control stream.
    
    However, that breaks any model type that does not inherit from `bbi_nonmem_model` but that does write a bbi-aligned config file after the model runs. So the old method used to just work for these models, and now they need to implement their own.
    
    In practice, it's unlikely that this matters outside the context of `bbr.bayes`, and in particular Stan models at this point.
    
    From a future-looking standpoint of other model types getting added, it's worth thinking through the fact that many of them are likely to write a `bbi_config.json`, and that would have just worked with the old method.

R/check-up-to-date.R
barrettk: With this change, if you were to change the relative path in the control stream file (but not move the data), you would _now_ receive two warnings: one for the the control stream, and one for the data (cant be found since the relative path is now wrong). Previously, you would only see that the control stream file had changed. 

I dont think this added warning is _that_ necessary, but I do think it would be more informative.

R/get-path-from-object.R
barrettk: Thanks for the added details. Will update the docs here to provide more context. However cc'ing @seth127 for the comment about where the documentation should live. My vote would actually be to include the info in both helper docs, but maybe that's a bit overkill?

R/get-path-from-object.R
barrettk: This was suggested by @seth127 [here](https://github.com/metrumresearchgroup/bbr/issues/651#issuecomment-1942423377):
>  If a bbi_config.json does exist, check the path and warn if they don't match

R/get-path-from-object.R
barrettk: I didn't realize it could be an absolute path as well. Will update variable/function names and any relevant docs

R/get-path-from-object.R
kyleam: > This was suggested by @seth127 [here](https://github.com/metrumresearchgroup/bbr/issues/651#issuecomment-1942423377):

That's fine, but "@seth127 said to do it" isn't the rationale that, as a reviewer, I'm trying to understand and get you to document/expand on.

---

My current view is that the approach here is overcomplicating things.  If the caller asked to pull from the config, get it from there and error if you can't.   If they said don't get or from the config, then get it from the model.

(That's taking for granted the current design of having an argument.  I'll expand on it later if I don't talk myself out of it while writing the comment, but one thing I wonder is whether we should just have the `bbi_model` method retain the bbi config handling and then have a custom `bbi_nonmem_model` method that falls back to extracting it from the control stream if the config file doesn't exist.  The `bbi_nonmem_model` method should of course document that custom handling on top.  Anyway, as I said, haven't though that through yet, and there are of course other variations.)

R/get-path-from-object.R
kyleam: > My vote would actually be to include the info in both helper docs, but maybe that's a bit overkill?

In my view, yes, or rather it's an invitation for docs about something complicated to get out of sync with each other.  (If you felt strongly that you didn't want a reference here, you could use `@template` to avoid that.)

R/check-up-to-date.R
kyleam: > one for the the control stream, and one for the data (cant be found since the relative path is now wrong). Previously, you would only see that the control stream file had changed.

Yes, but, in that case, you're subtlety changing the question, losing the answer to "did that data file that was used to run this model change?"

kyleam:
I said:

> I'm leaving any opinions about what, if anything, we should do about these [breaking changes] to a follow-up comment.

For this particular case, my understanding is that the current consensus is that we're okay accepting breakage here, especially to improve consistency, as long we think it's unlikely to matter in practice. (Of course unstated in that are the questions of what is considered public API for bbr and how closely this project tries to adheres to semantic versioning.) Anyway, I think all of the breaking changes I outlined above are unlikely to have any fallout outside of bbr.bayes, and I'm happy to make the needed adjustments there.

So I don't have any objections to making these changes.

Below I'm going to outline an alternative design for consideration. It was prompted by thinking about the last breakage above ("narrowing of method").


### Alternative design

The older "get data path from bbi\_config.json behavior" was under `get_data_path.bbi_model()`. How it works isn't tied to the model being NONMEM or (despite the confusing `bbi_model` class name) even being executed by bbi. The model run just needs to produce a json file that shares important features with the data file that bbi produces (e.g., has "data\_path" and "data\_md5" keys).

The switch of this PR to `get_data_path.bbi_nonmem_model()` and the control stream is coupling things more closely to NONMEM. And that feels fine for a custom method, but what about keeping things working for the more general case?

-   leave `get_data_path.bbi_model()` pulling from the `bbi_config.json`

-   add a `.check_exists` that defaults to `TRUE` for consistency with other `get-path-from-object.R` helpers

-   add a `get_data_path.bbi_nonmem_model()` method that first tries to return the data path from `bbi_config.json` (i.e. behaves the same as the other models) but, if that comes up empty, has extra handling on top to get it from the control stream
    
    (It doesn't really apply to Stan, but other model types could follow a similar pattern of having a tailored method that knows how to fall back to other spots for finding the data path.)

R/get-path-from-object.R
barrettk: I used the internal `unquote_filename` function we have for this. Marking as resolved

R/get-path-from-object.R
barrettk: I dont really have any opinion on this, which is why I pointed you to Seth. Im fine with either direction. That being said, in my most recent refactor I did remove this check and followed your recommendation about exclusively pulling from one location or the other.

barrettk:
>leave get_data_path.bbi_model() pulling from the bbi_config.json

>add a get_data_path.bbi_nonmem_model() method that first tries to return the data path from bbi_config.json (i.e. behaves the same as the other models) but, if that comes up empty, has extra handling on top to get it from the control stream

I feel pretty strongly about defaulting to pulling the path from the control stream file as long as there aren't notable reasons to avoid this (which it seems like there aren't). I think the default procedure should be to build the path _without_ attempting to check if the model was previously run. That just seems like a bunch of unnecessary checks, given that we shouldnt need to have submitted the model to extract the path. My main thing is that `nm_join` shouldnt need `bbi` to run. Your suggested workaround (checking the ctl if no json is found) would still allow for that, but I still think the default behavior should be to pull from the `ctl` file. Im ok with switching the internal calls (such as within `check_up_to_date`) to default to pulling from the config though.

I am curious how we would support a `get_data_path.bbi_nonmem_model()` and `get_data_path.bbi_model()` method, but I also see that being more of a benefit downstream if/when we add more model types to `bbr`.

barrettk:
Hopefully drone doesnt fail again, but FYI it failed twice in a row **at various stages** (different R versions and/or CI pipelines) when trying to download bbi:

```
══ Warnings ════════════════════════════════════════════════════════════════════
--
624 | ── Warning ('test-bbr.R:47:3'): check_bbi_exe() errors on too low version [BBR-BBR-004] ──
625 | URL 'https://api.github.com/rate_limit': Timeout of 60 seconds was reached
626 | Backtrace:
627 | ▆
628 | 1. └─bbr:::skip_if_over_rate_limit() at test-bbr.R:47:3
629 | 2.   └─utils::download.file(...)
630 |  
631 | ══ Failed tests ════════════════════════════════════════════════════════════════
632 | ── Error ('test-bbr.R:47:3'): check_bbi_exe() errors on too low version [BBR-BBR-004] ──
633 | Error in `download.file("https://api.github.com/rate_limit", destfile = tmp,
634 | quiet = TRUE)`: cannot open URL 'https://api.github.com/rate_limit'
635 | Backtrace:
636 | ▆
637 | 1. └─bbr:::skip_if_over_rate_limit() at test-bbr.R:47:3
638 | 2.   └─utils::download.file(...)
```

R/get-path-from-object.R
kyleam: It looks like that point in the [specification](https://github.com/metrumresearchgroup/bbr/issues/651#issuecomment-1942423377) was probably prompted by this part of your initial post in gh-651:

> I propose we attempt to fetch _both_ data paths. i.e. we extract the data path from the `ctl` file, and _if_ the model has been run, we also parse it from the `bbr` model object. We could compare the two, and warn the user if there is a mismatch (i.e. if the model file has been moved since it was run). We could have an argument for which one to return (with a default), to ensure a single string is usually returned.

But part of the reason why I'm confused about the code I was commenting on is that, as implemented, the check is done only with the non-default `pull_from_config = TRUE`.  That doesn't seem to align with the motivation described above.

In particular, the "control stream data path no longer matches with previous run" scenarios I see are:

 1. the caller goes with the default `pull_from_config = FALSE`. They are not warned about the mismatch.

 2. the caller passes an explicit `pull_from_config = FALSE`. They are not warned about mismatch.

 3. the caller passes a non-default `pull_from_config = TRUE`. They are warned about the mismatch.

Of all of these, the first seems like the one that it'd be most where it'd be most important to warn (because the caller isn't explicitly asking for source), but the check isn't done in that case.

I know the warning is now removed entirely. To me, whether that is a good direction to go depends how much we're worried about a scenario like this:

 * caller runs model

 * caller then edits control stream and points it to a different data path

 * caller later calls `nm_join()`. In this case, the appropriate data to join is the path in the config. That is was what was fed to the model that produced the result.

If we pull from the control stream by default, the above seems like an important mistake to guard against even if it will be very rare. (And again, the code I was commenting would not have warned here.)

Stepping back a bit, shouldn't `nm_join()` (or `nm_tables()` underneath, I suppose) explicitly request the config data here? It's concerned with an output run, which is tied to the data in the config regardless of what the current value in the control stream is.

kyleam:
@barrettk Thanks for the updates.  I'll focus on the higher level things right now rather than inline comments so that we can get consensus on a path forward.

> I feel pretty strongly about defaulting to pulling the path from the control stream file as long as there aren't notable reasons to avoid this (which it seems like there aren't). I think the default procedure should be to build the path _without_ attempting to check if the model was previously run. That just seems like a bunch of unnecessary checks, given that we shouldnt need to have submitted the model to extract the path.

Does this check boil down to something more than whether the config file exists?  To me, checking whether a file exists and getting the value from JSON (that we're responsible for writing) is a much simpler code path than parsing the value from the control stream, so it makes sense to go with that if it's available.

And in either case, you have the possibility of a mismatch (assuming no bugs, this is likely from the user changing the specified data path in some way).  But in that case, it's not at all clear to me that the control stream path should be preferred if an output directory exists.  At least in the common case of joining output tables to the data file, I think it'd be incorrect to not use the data that generated that path.

The reasons I see to prefer the config file is that 1) it's the simpler code path if we have it, 2) it sticks with what has been done in the past, so less chance of unforeseen surprises, and 3) I think it gives the correct precedence for `nm_join()` (or more generally of combining data with results, which should use the data that was input for those results).

> My main thing is that `nm_join` shouldnt need `bbi` to run.

I don't understand this comment for the following reasons:

 1. `nm_join()` is about joining output tables to the data, so that depends on there being a previous run

 2. `nm_join()` calls `bbi nonmem summary` on the output, but that's unrelated to getting the data path.

> I am curious how we would support a `get_data_path.bbi_nonmem_model()` and `get_data_path.bbi_model()` method, but I also see that being more of a benefit downstream if/when we add more model types to `bbr`.

I'm not sure I understand.  Regardless of the question about what source `.bbi_nonmem_model()` should give precedence to, why not add back a `.bbi_model()` method that returns the path from the config if it exists?  That'd avoid breaking Stan models in `bbr.bayes`, and would work for any other model types in the future that generate a `bbi_config.json` with a data path.

---

At a high level, here are the open questions I see at this point:

 * Should `nm_join()` prefer the config file?  (I think so, for the reasons given above and [here](https://github.com/metrumresearchgroup/bbr/pull/664#discussion_r1502655285))

 * Should `check_up_to_date()` prefer the data path from the config file?  ([I think so](https://github.com/metrumresearchgroup/bbr/pull/664#discussion_r1499290676).)

 * Is it worth adding back a `.bbi_model()` method that returns the data path from the config?  (I think so, for the reasons given above and [here](https://github.com/metrumresearchgroup/bbr/pull/664#issuecomment-1959853113))

kyleam:
> Hopefully drone doesnt fail again, but FYI it failed three times in a row [...]

Thanks for posting the details here.  Based on the output, it was a network issue.  We try to skip tests that hit the GitHub API if we're over the rate limit, but it looks like the URL that we use to check that was down/unresponsive.

So, as you say, unrelated to the changes here.

R/get-path-from-object.R
```diff
@@ -158,45 +158,221 @@ get_model_id.bbi_model <- function(.mod) {
...
-#' `bbi_{.model_type}_summary` object.
+#' Helper to extract the path to the input data from a bbi object.
+#'
+#' @param .bbi_object The object to query. Could be
+#' a `bbi_{.model_type}_model` object,
+#' `bbi_{.model_type}_summary` object,
+#' or a tibble of class `bbi_log_df`.
+#' @param .check_exists If `TRUE`, the default, will throw an error if the file
+#'  does not exist
+#' @param pull_from_config If `TRUE`, pull the data path from `bbi.json` created
```
seth127, line 169: Should be `bbi_config.json`

seth127:
@kyleam @barrettk thanks for all of the discussion here. This issue turned out to be a bit more complex than I expected...

I looked through the comments and code, and had some offline discussion with @barrettk, and I want to document my current thinking here. Hopefully we can agree on a path forward based on this.

# Keep current `get_data_path.bbi_model` behavior

I like [this suggestion of yours](https://github.com/metrumresearchgroup/bbr/pull/664#issuecomment-1959853113) above. Essentially, `get_data_path.bbi_model` gains a `.check_exists = TRUE` argument, but otherwise stays the same (continues to use the `bbi_config.json`).

This will hopefully make extending this method (e.g. in `bbr.bayes`) simpler and more reliable going forward.

# NONMEM-specific method

This leaves open the question of what the more specific NONMEM method(s) (`get_data_path.bbi_nonmem_model` and `get_data_path.bbi_nonmem_summary`) should do.

## Checking paths in ctl vs json

After going through this a bit, I've come back around to [my original thought](https://github.com/metrumresearchgroup/bbr/issues/651#issuecomment-1942423377) that, for NONMEM models _that have been run_, we should check for a data path in both the control stream and the `bbi_config.json` and _warn the user if they don't match._

Expanding on this a bit, these are the possibile scenarios I see:

1. model hasn't been run _--> I propose we pull the path from the ctl_
2. model has been run, and data path is the same in both ctl and config _--> In this case it doesn't matter which we prefer_
3. model has been run, and data paths don't match _--> we could either..._
  a. pull from config (and don't check ctl)
  b. pull from the ctl (and don't check config)
  c. check whether they match

The key cases are under number 3, and I can see a few different non-ideal things happening in either 3a or 3b. I can give more detail on these, if you'd like. They mostly boil down to a user loading a data set that isn't what they think it is (and receiving no friction or warning in doing so).

This leads me to prefer 3c, and I don't really see a down-side, other than some mild complexity in the code path.

<details><summary>**check_up_to_date() sidebar**</summary>

We had discussed that perhaps case 3 should just be handled by `check_up_to_date()`, but after testing a bit, I don't love this option. For one, it relies on a user knowing to call this. I think most bad cases in 3a and 3b would be unintentional (e.g. a user changing the data path in the control stream and _intending_ to re-run the model, but not doing it yet) and it seems likely they wouldn't be thinking to call `check_up_to_date()`.

Furthermore, even if they did, the messaging isn't particularly clear about the problem.  In various iterations of changing the control stream, renaming the file, etc. the most informative message I get is something like this. Often, I just get the first message, which (understandably) doesn't tell us anything about the changed data path being the issue.

```r
check_up_to_date(MOD1)
# The following files have changed in 1
# * /data/home/sethg/metrumresearchgroup/bbr/inst/model/nonmem/basic/1.ctl
# The following files in 1 ARE NO LONGER PRESENT
# * /data/home/sethg/metrumresearchgroup/bbr/inst/extdata/acop_v2.csv
```

I considered wanting to call `check_up_to_date()` under-the-hood in `get_data_path()`, but that now feels like a bad idea for several reasons. And, as I just wrote, I don't think the message is particularly helpful anyway.

</details>

## If the ctl and json paths don't match

This is essentially 3c: we check both the control stream and the `bbi_config.json` and we get two different paths. In this case, I would propose to 

* give the user a tailored warning
* return the path from the `bbi_config.json`

I was swayed by the [points above](https://github.com/metrumresearchgroup/bbr/pull/664#issuecomment-1964221266) that, in most cases when the model has been run, the user would want the path to the data _that the model was run with_.

As an **aside on `nm_join()`**, I think some of the discussion above is related to my comment in #595, which I tried to clarify [here](https://github.com/metrumresearchgroup/bbr/issues/595#issuecomment-1930533218). Continuing to try to clarify: **`nm_join()` should not expect to work on a not-run model, but `nm_data()` should.** Sorry for the confusion there.

## Proposed pseudo-code

There has been a lot of writing here, so I will distill my proposed path forward into some pseudo-code that is hopefully unambiguous. I am certainly open to further discussion, but I feel pretty strongly that, in a NONMEM-specific method, it is worth checking both places and warning the user if they don't match. I don't see it as much additional burden for us, and I think it is more likely to avoid actual undesirable behavior.

<details><summary>get_data_path Seth pseudo-code</summary>

_Note: the code below uses the `get_data_path_from_json()` and `get_data_path_from_ctl()` helpers that have been implemented in this PR. Those can change if needed, but nicely encapsulate the two different logistical options._

```r

get_data_path.bbi_model <- function(.bbi_object, .check_exists = TRUE, ...) {
  data_path <- get_data_path_from_json(.bbi_object)
  if (.check_exists = TRUE) { ... }
  return(data_path)
}

get_data_path.bbi_nonmem_model <- function(.bbi_object, .check_exists = TRUE, ...) { 
  get_data_path_nonmem_impl(.bbi_object, .check_exists = .check_exists, ...)
}

# note: we need this method too, otherwise it will fall back to bbi_model
#  (which would technically work, but skips the check vs the control stream)
get_data_path.bbi_nonmem_summary <- function(.bbi_object, .check_exists = TRUE, ...) { 
  get_data_path_nonmem_impl(.bbi_object, .check_exists = .check_exists, ...)
}

get_data_path_nonmem_impl <- function(.bbi_object, .check_exists = TRUE, ...) {
  path_from_ctl <- get_data_path_from_ctl(.bbi_object)
  
  path_from_config <- tryCatch(
    get_data_path_from_json(.bbi_object),
    error = function(.e) {
      if (str_contains(.e$message, "Cannot extract data path")) {
      	  return NULL
      	} else { stop(.e) } # this syntax may not be quite right... basically raise the error if it's anything else
    }
  )
  
  # if no config (model hasn't been run) then pull from ctl, otherwise use config path
  data_path <- if (is.null(path_from_config)) {
    path_from_ctl  
  } else {
    # if they're not the same, warn
    if (path_from_config != path_from_ctl) {
      warn("they aren't equal, preferring config path")
    }
    path_from_config
  }
  
  if (.check_exists = TRUE) { ... }
  return(data_path)  
  
}

```
</details>

kyleam:
@barrettk @seth127 Thanks for documenting your discussion.  Reading through that and your pseudo-code, that sounds like a good direction to me.

> in a NONMEM-specific method, it is worth checking both places and warning the user if they don't match. I don't see it as much additional burden for us, and I think it is more likely to avoid actual undesirable behavior.

This is probably clear from my last comment in the thread [here](https://github.com/metrumresearchgroup/bbr/pull/664#discussion_r1502655285), but I agree with the value of warning here on the mismatch.  I very much appreciate having the rationale and scenarios laid out.  Thanks.

R/get-path-from-object.R
```diff
@@ -158,45 +158,221 @@ get_model_id.bbi_model <- function(.mod) {
...
+#' @details
+#'
+#' ### Notes on model extensions:
+#' `bbi` has different handling for `.mod` vs `.ctl` extensions for `NONMEM`
+#' models. `bbi` was designed to mimic `PsN` behavior. As such, the control
+#' stream file is first copied to a subdirectory before executing the model. The
+#' difference is that on copy `bbi` adjusts a relative data path for `.mod`
+#' files, adding on a `"../"`. Or, said from the user perspective, writers of
+#' `.ctl` files are expected to specify a relative path for the data file
+#' **one level deeper** than the control stream they're editing.
```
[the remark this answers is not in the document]
seth127: I like @kyleam 's wording here (and I agree that we don't need the link to bbi code or docs).

I also think putting this in `?submit_model` makes more sense, and that the `?get_data_path` docs should have something like `See "Notes on NONMEM model extensions" in [submit_model()] documentation for more information on handling of data paths in control stream files.`

What do you think about that @barrettk ?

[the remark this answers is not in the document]
barrettk: I think it would be good to have this information pop up sooner rather than later, which definitely points to `submit_model`. However given the relevance of the information, I think it belongs more with `get_data_path()`. I haven't looked into that `@template` thing @kyleam mentioned yet, but I would lean towards being more verbose in the `get_data_path` helper docs, while mentioning it/linking to the `get_data_path` docs within `submit_model`.

Dont have a strong opinion here if you guys disagree with that logic. I definitely think this information should get exposed *before* `get_data_path`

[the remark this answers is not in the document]
barrettk: FYI I ended up opting for the `@template` option. I think that's a good practice to start enforcing for documentation we may want to include in multiple places. I followed what `plyr` did [here](https://github.com/hadley/plyr/tree/main/man-roxygen) and [here](https://github.com/hadley/plyr/blob/main/R/aaply.r#L16-L18).

R/check-up-to-date.R
barrettk: Marking this resolved for the same reasons highlighted [here](https://github.com/metrumresearchgroup/bbr/pull/664#discussion_r1504776552) (i.e. moot point after the refactor).

kyleam: > (i.e. moot point after the refactor).

I agree it's resolved.   I wouldn't classify it as a moot point, any more than any other review comment that is raised and then addressed.

R/get-path-from-object.R
```diff
@@ -158,45 +158,234 @@ get_model_id.bbi_model <- function(.mod) {
...
+#' @rdname get_data_path
 #' @export
-get_data_path.bbi_model <- function(.mod, ...) {
+get_data_path.bbi_model <- function(
+    .bbi_object,
+    .check_exists = TRUE,
+    ...
+){
+  data_path <- get_data_path_from_json(.bbi_object)
+  if(.check_exists){
```
kyleam, line 193: Minor, but best for this to use `isTRUE` because this value is coming from the caller as is. (Same comment may apply in some other spots here.)

R/get-path-from-object.R
```diff
@@ -158,45 +158,234 @@ get_model_id.bbi_model <- function(.mod) {
...
+
+  # Get data record
+  data_recs <- nmrec::select_records(ctl, "data")
+  n_data <- length(data_recs)
+  if(n_data !=1){
+    recs_fmt <- purrr::map_chr(data_recs, function(rec) rec$format())
+    rlang::abort(
+      c(
+        glue::glue("Expected a single data record, but found {n_data}:\n\n"),
+        glue("{recs_fmt}")
```
kyleam, line 313: nit-pick, unnecessary glue: `glue("{recs_fmt}")` is just `rec_fmt`

R/get-path-from-object.R
```diff
@@ -158,45 +158,234 @@ get_model_id.bbi_model <- function(.mod) {
...
+  data_path_ctl <- nmrec::get_record_option(data_recs[[1]], "filename")$value
+  data_path_ctl <- unquote_filename(data_path_ctl)
+
+  if(isTRUE(normalize)){
+    # Normalize to current model directory
+    model_dir <- dirname(get_model_path(.mod, .check_exists = FALSE))
+    # Adjust for model extension
+    data_path_ctl_adj <- adjust_data_path_ext(data_path_ctl, mod_path)
+    data_path <- fs::path_norm(file.path(model_dir, data_path_ctl_adj)) %>%
+      as.character()
```
kyleam, line 328: Won't this mishandle absolute paths?

For example, for an absolute path "/baz/zzz" this logic would incorrectly append the model directory:

```R
fs::path_norm(file.path("/foo/bar", "/baz/zzz"))
#> /foo/bar/baz/zzz
```

(`get_data_path_from_json` may have a similar pre-existing bug here; it depends on whether bbi remaps a data path to a relative path before writing `bbi_config.json` and I haven't confirmed that yet.)

R/get-path-from-object.R
```diff
@@ -158,45 +158,234 @@ get_model_id.bbi_model <- function(.mod) {
...
+#' Adjust the data path based on the model extension
+#'
+#' @param data_path a file path to an input data set
+#' @param mod_path a relative or absolute model path. Must include the extension.
+#'
+#' @noRd
+adjust_data_path_ext <- function(data_path, mod_path){
+  checkmate::assert_true(is_valid_nonmem_extension(mod_path))
+
+  if(grepl("ctl", fs::path_ext(mod_path))){
```
kyleam, line 375: nit-pick: I think the `grepl` use here is confusing because a reader needs to look at the callers to verify that this can only be "mod" or "ctl", so there's no need to worry about this pattern matching some mis-specified extension like "ctlo" or "actl".

In my opinion the `identical(fs::path_ext("mod_path"), "ctl")` I suggested <https://github.com/metrumresearchgroup/bbr/pull/664#discussion_r1499293489> would make things clearer.

tests/testthat/test-get-path-from-object.R
```diff
@@ -149,6 +149,106 @@ test_that("get_data_path parses summary object [BBR-GPFO-017]", {
   expect_identical(readLines(res_data_path, n = 1), DATA_TEST_FIRST_LINE)
 })
 
+test_that("get_data_path parses bbi_*_log_df object", {
```
kyleam, line 152: Nice, thanks adding all these cases.

tests/testthat/test-get-path-from-object.R
```diff
@@ -149,6 +149,106 @@ test_that("get_data_path parses summary object [BBR-GPFO-017]", {
...
+  # Expect mismatch now that the relative/defined paths are different
+  expect_false(
+    isTRUE(
+      all.equal(get_data_path_from_ctl(mod), get_data_path_from_json(mod))
+    )
+  )
+  expect_warning(
+    get_data_path(mod, .check_exists = FALSE),
+    "does not match the one defined in the control stream"
+  )
```
kyleam, line 236: Could you also assert that the returned result matches the config rather than the model value? I'd like to have that behavior recorded it in the tests somewhere.

(Unless you're already doing that somewhere, and I'm missing it.)

tests/testthat/test-get-path-from-object.R
```diff
@@ -149,6 +149,106 @@ test_that("get_data_path parses summary object [BBR-GPFO-017]", {
...
+
+  # Copy over data and modify json
+  fs::file_copy(
+    get_data_path(MOD1),
+    file.path(temp_dir, basename(data_path_real))
+  )
+  json <- jsonlite::read_json(file.path(temp_dir, "1", "bbi_config.json"))
+  json$data_path <- "../acop.csv"
+  jsonlite::write_json(json, file.path(temp_dir, "1", "bbi_config.json"))
```
kyleam, line 245: Okay as is, but should you use your `modify_data_path_json` test helper here for consistency?

kyleam (changes requested):
@barrettk Thanks for the updates. Left a few comments, but the only thing I consider blockers are

-   testing config precedence on mismatch

-   fixing handling of absolute paths in `get_data_path_from_ctl` and adding a test case. (edit: as I mention inline, bbi also mishandles this, so I consider this more a nice to have than a blocker.  The fix is straightforward.)

---

I've confirmed that bbr.bayes tests (f70941c) pass when running bbr from scratch merge of this PR (0113aec3) and main (88e96a6f)

Minor doc note: building with latest roxygen, I get docs changes on top of what you committed (looks like just a change in order of an entry). Not a big of a deal, but could verify that the docs are up to date on your end?

R/get-path-from-object.R
kyleam: > it depends on whether bbi remaps a data path to a relative path before writing `bbi_config.json` and I haven't confirmed that yet

I just checked and it looks like bbi mishandles it and errors:

```
mdev:bbi-cli-testing$ grep DATA abs.ctl
$DATA /data/home/kylem/scratch/bbi-cli-testing/acop.csv IGNORE=@
mdev:bbi-cli-testing$ bbi nonmem run local --overwrite abs.ctl
INFO[0000] Successfully loaded default configuration from /data/home/kylem/scratch/bbi-cli-testing/bbi.yaml 
INFO[0000] Beginning Local Path                         
INFO[0000] A total of 1 models have completed the initial preparation phase 
INFO[0000] [abs] Beginning local work phase             
INFO[0007] [abs] Beginning cleanup phase                
ERRO[0007] abs error reading data to hash: open /data/home/kylem/scratch/bbi-cli-testing/abs/data/home/kylem/scratch/bbi-cli-testing/acop.csv: no such file or directory 
ERRO[0007] abs error hashing data: invalid argument     
INFO[0007] [abs] Cleanup completed                      
INFO[0007] Waiting for any post execution hooks to finish 
1 models completed in 7.64932725s       
```

Which of course makes a fix here less important, but I'd still appreciate if this was adjusted to not append the model's directory to an absolute path.

kyleam: It at least ends up with the correct absolute path in the `bbi_config.json`, but there's no hash:

```
$  jq '{data_path, data_md5}' <abs/bbi_config.json 
{
  "data_path": "/data/home/kylem/scratch/bbi-cli-testing/acop.csv",
  "data_md5": ""
}
```

Summary: it would be great if both the `from_ctl` and `from_json` avoided tacking on an invalid leading path to an absolute file, but there are already existing issues here, so I don't consider that a blocker.

Given the above issue,

barrettk: How do you envision knowing if the path is absolute or not? I tried something like the following, but it caused a number of tests to fail:
```r
    # Adjust for model extension
    data_path_ctl_adj <- adjust_data_path_ext(data_path_ctl, mod_path)
    # Normalize to current model directory if not an absolute path
    model_dir <- dirname(get_model_path(.mod, .check_exists = FALSE))
    if(data_path_ctl_adj != normalizePath(data_path_ctl_adj)){
      data_path <- fs::path_norm(file.path(model_dir, data_path_ctl_adj)) %>%
        as.character()
    }else{
      data_path <- data_path_ctl_adj
    }
```

kyleam: > How do you envision knowing if the path is absolute or not?

`fs::is_absolute_path()`?   If that's true, avoid doing `file.path(model_dir, ...)`.

barrettk: That seemed to work more appropriately. Added a test for this as well

tests/testthat/test-get-path-from-object.R
barrettk: Added

tests/testthat/test-get-path-from-object.R
barrettk: Good catch; yes that was my intention

kyleam (approved):
@barrettk Thanks for the updates.  Ready to go from my POV assuming CI passes.

R/get-path-from-object.R
kyleam: I said:

> I just checked and it looks like bbi mishandles it and errors when hashing the file

Issue filed here: https://github.com/metrumresearchgroup/bbi/issues/316

PR #665, opened 2024-02-23
DESCRIPTION: bump dev version

kyleam:
I made a bbr.bayes dev tag for @timwaterhouse to use in the expo.  I'm pushing a dev tag here too so that the required bbr can be pulled in.


kyleam:
This isn't a code change and is just making a dev release for testing purposes, so I'll merge without review.

PR #667, opened 2024-02-28
Prepare for 1.10.0 Release

barrettk:
Changeset: https://github.com/metrumresearchgroup/bbr/compare/1.9.0...release/1.10.0

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,23 @@
+# bbr 1.10.0
+
+## New features and changes
+
+* When passed an `nmbayes` model (defined by the `bbr.bayes` package),
+  `nm_join()` now gives a more informative error that points to
+  `bbr.bayes::nm_join_bayes()`. (#662)
+
+* `check_nonmem_finished()` has been converted to a method so that tailored
+  logic can be implemented for derived model types, such as the `nmbayes` model
+  type in `bbr.bayes`. (#663)
+
+* The character methods for `tail_lst()` and `tail_output()` have been adjusted
+  to work with custom model types where `*.lst` and `OUTPUT` files do not reside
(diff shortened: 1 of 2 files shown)


barrettk:
### Scorecard Details

image

NEWS.md
```diff
@@ -1,3 +1,23 @@
...
+  logic can be implemented for derived model types, such as the `nmbayes` model
+  type in `bbr.bayes`. (#663)
+
+* The character methods for `tail_lst()` and `tail_output()` have been adjusted
+  to work with custom model types where `*.lst` and `OUTPUT` files do not reside
+  in the top-level output directory. (#661)
+  
+* `get_data_path()` can now pull from the control stream file if the model
+  has not yet been submitted. This allows functions like `nm_data()` to be used
+  pre-model submission. (#664)
```
kyleam, line 19: Do you think it's worth mentioning

 * that there's no a `.check_exists` argument that defaults to `FALSE`
 * support for log (via new log methods)
?

kyleam (approved):
Looks good.  I've left a couple of suggestions for other things to mention in the NEWS, if you agree.

I've also verified that `check_pkgdown()` is happy.

NEWS.md
barrettk: Yeah thought about this. Was worried the bullet would be too long, but will add those elements

kyleam: > Was worried the bullet would be too long

Up to you, but you can put them as separate bullets.

barrettk: @kyleam let me know what you think of the new bullet point before I merge

kyleam: Works for me.  Thanks for adding.

PR #671, opened 2024-03-15
Support bootstrap runs

barrettk:
## New Bootstrap functionality

The functions below do not capture all new functions or supported S3 methods, but rather highlight key functions that will be helpful to users and/or future developers.

### New exported functions:
 - **`new_bootstrap_run()`**: Create a boostrap model object from an existing model
    - Example call:
        ```r
       .boot_run <- new_bootstrap_run(.mod) %>% inherit_param_estimates() %>%
          tweak_initial_estimates()
        ```
 - **`setup_bootstrap_run()`**: Creates a new model object and re-sampled dataset per model run
    - Example call:
        ```r
       setup_bootstrap_run(.boot_run, n = 100, strat_cols = c("SEX", "ETN"), seed = 1234)
        ```
 - **`get_boot_models()`**: Read in all bootstrap run model objects
    - Example call:
        ```r
       boot_models <- get_boot_models(.boot_run)
        ```
 - **`summarize_bootstrap_run()`**:  Summarize a bootstrap run and store results
    - Example call:
        ```r
       .boot_run <- read_model(file.path(MODEL_DIR, "1-boot")) # Just to show reading in a model later
       boot_sum <- summarize_bootstrap_run(.boot_run)

       > boot_sum
       
       ── Based on ───────────────────────────────────────────────
       Model: /data/Projects/package_dev/bbr/inst/model/nonmem/basic/1.ctl
       Dataset: /data/Projects/package_dev/bbr/inst/extdata/acop.csv
       
       ── Run Specifications ───────────────────────────────────────────
       • Stratification Columns: SEX, ETN
       • Seed: 1234
       
       ── Bootstrap Run Summary ────────────────────────────────────────
       Estimation Method(s):
       
       – First Order Conditional Estimation with Interaction 
       
       Run Status:
       
       – Number of runs: 100


       |parameter_names |original |p50     |p2.5   |p97.5   |
       |:---------------|:--------|:-------|:------|:-------|
       |THETA1          |2.32     |2.30    |2.17   |2.50    |
       |THETA2          |54.6     |54.2    |50.2   |58.0    |
       |THETA3          |463      |462     |416    |522     |
       |THETA4          |-0.0820  |-0.0807 |-0.149 |-0.0573 |
       |THETA5          |4.18     |4.14    |-1.25  |5.84    |
       |OMEGA(1,1)      |0.0985   |0.0988  |0.0677 |0.138   |
       |OMEGA(2,1)      |0.00     |0.00    |0.00   |0.00    |
       |OMEGA(2,2)      |0.157    |0.148   |0.101  |0.203   |
       |SIGMA(1,1)      |1.00     |1.00    |1.00   |1.00    |
       ... 0 more rows
        ```
        ```r
       # Optionally compare to original estimates (uses bootstrap summary object)
       > param_estimates_compare(boot_sum)
       # A tibble: 9 × 5                                                                                                                                                                                                    
         parameter_names original      p50     p2.5    p97.5
         <chr>              <dbl>    <dbl>    <dbl>    <dbl>
       1 THETA1            2.32     2.30     2.17     2.50  
       2 THETA2           54.6     54.2     50.2     58.0   
       3 THETA3          463.     462.     416.     522.    
       4 THETA4           -0.0820  -0.0807  -0.149   -0.0573
       5 THETA5            4.18     4.14    -1.25     5.84  
       6 OMEGA(1,1)        0.0985   0.0988   0.0677   0.138 
       7 OMEGA(2,1)        0        0        0        0     
       8 OMEGA(2,2)        0.157    0.148    0.101    0.203 
       9 SIGMA(1,1)        1        1        1        1    
        ```
 - **`bootstrap_estimates()`**:  Tabulate parameter estimates for each model submission in a bootstrap run
    - Example call:
        ```r
       # Long format is helpful for plotting estimates:
       bootstrap_estimates(.boot_run, format_long = TRUE) %>%
       dplyr::filter(grepl("THETA", parameter_names)) %>%
         ggplot(aes(x = estimate)) + facet_wrap(~parameter_names, scales = "free") +
         geom_histogram(color = "white", alpha = 0.7) + theme_bw()
        ```
        ![image](https://github.com/metrumresearchgroup/bbr/assets/18685116/74ab6754-807e-4ded-a5b2-7cc826eacd06)

 - **`get_model_status()`**: Returns messages indicating which model(s) have finished executing and which are still running.
    - Example call:
        ```r
       > get_model_status(.boot_run)
        77 model(s) have finished
        23 model(s) are incomplete
        ```
 - **`cleanup_bootstrap_run()`**: Cleanup bootstrap run directory. This will delete all child models, and only keep the information you need to read in estimates or summary information.
    - Example call:
        ```r
       cleanup_bootstrap_run(.boot_run)
        ```
### Internal Helpers
 - **`get_boot_spec_path()`**:  Get the bootstrap specification file path
 - **`get_boot_spec()`**:  Reads in and formats the bootstrap specification file. Tabulates all relevant `bbi_nonmem_model` model files from a bootstrap control stream file.


closes https://github.com/metrumresearchgroup/bbr/issues/682


barrettk:
See message in https://github.com/metrumresearchgroup/bbr/pull/671/commits/4e4d6afa0ab3ca91c6c2e86ab1ee1f09edaf499a for context, but just wanted to add some more thoughts:

 - While I think a bbi patch to summarize model runs in some format/cleanup the individual run files would ideal, I think it could be quite challenging, and take a decent amount of time (both in terms of development and defining how we would even want to go about doing that). This could be more of a long term vision; similar to how we refactored model_summary() in bbi 3.2.0.
 - I think we could have a temporary post-processing bootstrap function that would do a few things. A) check that all model runs are complete, and only proceed if so. B) Summarize all model runs so that the bootstrap model can be read in at a later time without needing the individual run data. It would optionally cleanup the other files too. The final thing it would do would be to create a new bbi json file that would include any relevant info about the model run. If this file is present, checking the model status (potentially other helper functions) would read that file instead of iterating through each of the models. This would speed up printing the model to the console, and likely have other benefits down the line too.
   - this idea could be nice, as it would allow each user to choose whether or not they want to keep all the individual model files (if they wanted to read one individual model in for whatever reason).

barrettk:
FYI Im refactoring the `wait_for_nonmem` function a bit. Im making the underlying `check_nonmem_finished` use S3 (was set up to, but only had one method), which negates the need for `wait_for_nonmem` to. Given that `bbr.bayes` plans to use this (but hasnt yet from what I can tell - only [mention](https://github.com/search?q=repo%3Ametrumresearchgroup%2Fbbr.bayes%20wait_for_nonmem&type=issues) I saw), I opted to make `wait_for_nonmem` just have a default class, which supports `bbi` models, a list of models, and now bootstrap models. These changes should not impact `bbr.bayes`, but just mentioning it ahead of time.

cc @seth127 @kyleam

kyleam:
Thanks for the heads up @barrettk.  I read your comment a few times and am not really following it, but I'll leave some comments/clarifications in case they're helpful.

> FYI Im refactoring the `wait_for_nonmem` function a bit. Im making the underlying `check_nonmem_finished` use S3 (was set up to, but only had one method), which negates the need for `wait_for_nonmem` to.

I'm confused what "use S3" means given, as you mention, it's already an S3 method.  It was converted to an S3 method with 84ee34b9 (check_nonmem_finished: convert to S3 method, 2024-02-14) so that bbr.bayes could implement a custom method.

> Given that `bbr.bayes` plans to use this (but hasnt yet from what I can tell - only [mention](https://github.com/search?q=repo%3Ametrumresearchgroup%2Fbbr.bayes%20wait_for_nonmem&type=issues) I saw), I opted to make `wait_for_nonmem` just have a default class, which supports `bbi` models, a list of models, and now bootstrap models.

I'm not aware of any plans to implement a custom `wait_for_nonmem` method for bbr.bayes.  As I mention in [this comment](https://github.com/metrumresearchgroup/bbr.bayes/issues/36#issuecomment-1865444729), adding the custom `check_nonmem_finished()` method was all that was needed to make `wait_for_nonmem()` compatible with nmbayes models.

> The code reads a lot better with these (uncommitted) changes in my opinion as well.

I'll try to take a look once you push them, which will probably clear up my confusion above.

> These changes should not impact `bbr.bayes`, but just mentioning it ahead of time.

I think so, as bbr.bayes can implement its own `check_nonmem_finished` method (and it sounds like that's staying as is).

(No need to spend time clearing up my confusion above, as long as you think bbr.bayes will be unaffected.)

barrettk:
For anyone looking through this, here are some example calls I have in my scratch pad that should work as of the latest commit:

### Starting model
```r
.mod <- MOD1
```

### New bootstrap run, or read in a previous one
```r
.boot_run <- new_bootstrap_run(.mod)
.boot_run <- read_model(file.path(MODEL_DIR, "1_boot_run_1"))
```

### Sample data and create bootstrap runs ahead of time
```r
.boot_run <- setup_bootrap_run(.boot_run, n = 10, .overwrite = TRUE)
```
### Helpers
```r
get_boot_spec_path(.boot_run) # Get the file path of the spec file
boot_spec <- get_boot_spec(.boot_run) # table of bootstrap run model & data paths
boot_models <- get_boot_models(.boot_run) # list of all bootstrap run models
```

### Submission
```r
.p <- submit_model(.boot_run, .overwrite = TRUE)

# Check status
check_nonmem_finished(.boot_run)

> get_model_status(.boot_run, max_print = 3)
10 model(s) have finished                                                                                                                                                                                
0 model(s) are still running
```

### Summarize
```r
summarize_bootstrap_run(.boot_run) # Default
summarize_bootstrap_run(.boot_run, estimates_only = TRUE) # quicker - no `model_summaries` call
summarize_bootstrap_run(.boot_run, include_based_on = TRUE) # include based on model as part of the results
```

#### Example summary call:
```r
> summarize_bootstrap_run(.boot_run, include_based_on = TRUE)
# A tibble: 11 × 16                                                                                                                                                       
   run      absolute_model_path     ofv_no_constant constant_to_ofv ofv_with_constant THETA1 THETA2 THETA3   THETA4 THETA5 `OMEGA(1,1)`
   <chr>    <chr>                   <chr>           <chr>           <chr>              <dbl>  <dbl>  <dbl>    <dbl>  <dbl>        <dbl>
 1 based_on /data/Projects/package… 2583.311        1361.866906209… 3945.17780715107    2.32  54.6   463.  -0.0820    4.18       0.0985
 2 01       /data/Projects/package… 2566.297        1361.866906209… 3928.16388710692    2.33  52.4   496.  -0.0856    4.19       0.0721
 3 02       /data/Projects/package… 1923.499        1361.866906209… 3285.36557304593    2.13  58.4   464.   0.0593    1.13       0.0824
 4 03       /data/Projects/package… 4215.585        1361.866906209… 5577.45165426658    2.97   3.84   13.0 -0.0110    6.98       0.0656
 5 04       /data/Projects/package… 2654.071        1361.866906209… 4015.93741478159    2.36  54.4   434.  -0.0737    4.23       0.101 
 6 05       /data/Projects/package… 2734.369        1361.866906209… 4096.23611699424    2.31  56.2   436.  -0.127     5.23       0.0805
 7 06       /data/Projects/package… 3837.497        1361.866906209… 5199.3639870871     5.16   4.11   13.1 -0.00920   6.09       0.0547
 8 07       /data/Projects/package… 2630.112        1361.866906209… 3991.9792510367     2.32  55.8   419.  -0.0815    4.32       0.114 
 9 08       /data/Projects/package… 2602.424        1361.866906209… 3964.29135050524    2.27  54.9   422.  -0.0830    4.29       0.104 
10 09       /data/Projects/package… 3018.395        1361.866906209… 4380.26178365223    2.20  56.2    15.5 -0.0234    5.20       0.0675
11 10       /data/Projects/package… 3538.885        1361.866906209… 4900.75177245378    3.62   3.88   13.0 -0.0104    6.70       5.43  
# ℹ 5 more variables: `OMEGA(2,1)` <dbl>, `OMEGA(2,2)` <dbl>, `SIGMA(1,1)` <dbl>, error_msg <lgl>, termination_code <dbl>
```

kylebaron (commented):
First round of comments.

R/bootstrap-model.R
```diff
@@ -0,0 +1,628 @@
...
+#'
+#' @seealso setup_bootstrap_run
+#' @return S3 object of class `bbi_nmboot_model`.
+#' @export
+new_bootstrap_run <- function(
+    .mod,
+    .suffix = glue("{get_model_id(.mod)}_boot"),
+    .inherit_tags = TRUE,
+    .overwrite = FALSE,
+    increment = FALSE,
```
kylebaron, line 23: I think this increment is going to result in just a lot of runs getting triggered. Instead, could we have the user supply a sub-run number?

R/bootstrap-model.R
```diff
@@ -0,0 +1,628 @@
...
+
+  checkmate::assert_class(.mod, NM_MOD_CLASS)
+
+  model_dir <- get_model_working_directory(.mod)
+
+  # Increments the bootstrap run
+  # (i.e. impossible to overwrite an existing run via this function)
+  if(isTRUE(increment)){
+    new_run_id <- get_next_boot_run(model_dir, .suffix)
+    boot_run_id <- paste0(.suffix, "_run_", new_run_id)
```
kylebaron, line 36: IMO this creates some inconsistency in the file naming.

R/bootstrap-model.R
```diff
@@ -0,0 +1,628 @@
...
+#' submit_model(.boot_run)
+#'
+#' # Check status of runs during submission
+#' get_model_status(.boot_run)
+#' }
+#' @export
+setup_bootstrap_run <- function(
+    .boot_run,
+    n = 100,
+    strat_cols = NULL,
```
kylebaron, line 115: Save `strat_cols` somewhere; it could / should show up in boot run summaries, printing the model object, etc.

R/print.R
```diff
@@ -281,10 +281,188 @@ print.bbi_nonmem_summary <- function(x, .digits = 3, .fixed = FALSE, .off_diag =
...
+#' @importFrom purrr map_chr
+#' @importFrom cli cat_line
+#' @importFrom dplyr mutate_if
+#' @importFrom checkmate assert_number
+#'
+#' @param .digits Number of significant digits to use for parameter table. Defaults to 3.
+#' @param .nrow If `NULL`, the default, print all rows of the parameter table.
+#'   Otherwise, prints only `.nrow` rows.
+#' @export
+print.bbi_nmboot_summary <- function(x, .digits = 3, .nrow = 10, ...) {
```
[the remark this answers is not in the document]
barrettk: FWIW it's a bit different now after the latest refactor:
```r
> boot_sum
Based on Dataset: /data/Projects/package_dev/bbr/inst/extdata/analysis3.csv

Median values --> Records: 4299	 Observations: 4299	 Subjects: 4299

Objective Function Summary (final est. method):

                               ┌──────|────────────┐                              
      |------------------------|      |            |------------------------------|
                               └──────|────────────┘                              
      ↓                        ↓      ↓            ↓                              ↓
   29002.02                30497.29 31065.65    31892.17                      33679.53

Estimation Method(s):

– First Order Conditional Estimation with Interaction 

No Heuristic Problems Detected


|run |parameter_names |estimate |error_msg |termination_code |ofv      |any_heuristics |
|:---|:---------------|:--------|:---------|:----------------|:--------|:--------------|
|001 |THETA1          |0.396    |NA        |0.00             |3.22e+04 |FALSE          |
|001 |THETA2          |4.12     |NA        |0.00             |3.22e+04 |FALSE          |
|001 |THETA3          |1.08     |NA        |0.00             |3.22e+04 |FALSE          |
|001 |THETA4          |4.22     |NA        |0.00             |3.22e+04 |FALSE          |
|001 |THETA5          |1.27     |NA        |0.00             |3.22e+04 |FALSE          |
|001 |SIGMA(1,1)      |0.0415   |NA        |0.00             |3.22e+04 |FALSE          |
|001 |OMEGA(1,1)      |0.181    |NA        |0.00             |3.22e+04 |FALSE          |
|001 |OMEGA(2,1)      |0.0617   |NA        |0.00             |3.22e+04 |FALSE          |
|001 |OMEGA(2,2)      |0.0769   |NA        |0.00             |3.22e+04 |FALSE          |
|001 |OMEGA(3,1)      |0.127    |NA        |0.00             |3.22e+04 |FALSE          |
... 1190 more rows
```

Can you clarify what you mean by "it's almost never going to be readable?"

NAMESPACE
```diff
@@ -161,8 +176,10 @@ export(replace_model_field)
 export(replace_note)
 export(replace_tag)
 export(run_log)
+export(setup_bootstrap_run)
```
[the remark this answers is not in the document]
barrettk: Yes if you take a look at an example `bbr_boot_spec.json` file in the run directory it's stored in that file. It's stored in the summary object too, but can definitely add it to the print method. Would you want to add `replace`, `seed`, and `n` there too?

R/submit-model.R
```diff
@@ -85,6 +85,47 @@ submit_model.bbi_nonmem_model <- function(
   return(res)
 }
 
+#' @describeIn submit_model Takes a `bbi_nmboot_model` object.
+#' @export
+submit_model.bbi_nmboot_model <- function(
+    .mod,
+    .bbi_args = NULL,
+    .mode = getOption("bbr.bbi_exe_mode"),
```
kylebaron, line 93: Probably should be `.mode = "sge"` or maybe a boot exe mode option?

NAMESPACE
```diff
@@ -125,6 +139,7 @@ export(model_diff)
 export(model_summaries)
 export(model_summary)
 export(modify_model_field)
+export(new_bootstrap_run)
```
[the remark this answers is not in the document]
barrettk: This cant happen now that I got rid of the `increment` argument. I do think we'll want to rename the `.suffix` argument now that that that's the full name though

NAMESPACE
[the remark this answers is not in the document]
barrettk: There was an `overwrite` bug in `setup_bootstrap_run` in the version installed in the repo. Unsure if the latest fix addresses this (think it does), but will look into it

R/bootstrap-model.R
barrettk: Mentioned above, but I opted to just remove the `increment` argument. So now the only way to make a new run would be to change the `.suffix` arg (may be a different name later)

R/bootstrap-model.R
```diff
@@ -0,0 +1,628 @@
...
+    replace = TRUE,
```
kylebaron, line 116: Drop the replace functionality; for bootstrap applications, it should always be true.

R/bootstrap-model.R
```diff
@@ -0,0 +1,628 @@
...
+    mod_names <- purrr::map_chr(n_seq, max_char = nchar(n), pad_left)
+    mod_paths <- file.path(boot_dir, mod_names)
+    orig_mod <- read_model(get_based_on(.boot_run))
+
+    boot_args <- list(
+      boot_run = .boot_run,
+      all_mod_names = mod_names,
+      orig_mod_path = get_model_path(orig_mod),
+      orig_mod_id = get_model_id(orig_mod),
+      orig_data = nm_data(orig_mod) %>% suppressMessages(),
```
kylebaron, line 148: Should only include data that entered the original problem.

R/bootstrap-model.R
```diff
@@ -0,0 +1,668 @@
...
+    verbose_msg(glue("Sampling run {new_mod_index}/{length(boot_args$all_mod_names)}"))
+  }
+
+  # Copy over control stream
+  #  - Cant use copy_model_from, as we want these individual model runs to be
+  #    regular `bbi_base_model` objects
+  orig_mod_path <- boot_args$orig_mod_path
+  mod_path_ext <- paste0(mod_path, ".", fs::path_ext(orig_mod_path))
+  fs::file_copy(
+    orig_mod_path, mod_path_ext, overwrite = boot_args$overwrite
```
kylebaron, line 169: `orig_mod_path` points to the original model (e.g., 106.ctl or 102.ctl). This should point to the base boot model (106-boot.ctl or 102-boot.ctl`). Only the `-boot` versions will have `$TABLE` and `$COV` removed if requested.

R/bootstrap-model.R
```diff
@@ -0,0 +1,668 @@
...
+    based_on_model_md5 = tools::md5sum(boot_args$orig_mod_path),
+    based_on_data_md5 = tools::md5sum(get_data_path(boot_args$boot_run)),
+    output_dir = boot_args$boot_run[[ABS_MOD_PATH]]
+  )
+
+  boot_run_ids <- purrr::map_chr(boot_models, function(boot_run){
+    paste0("run_", basename(boot_run[[ABS_MOD_PATH]]))
+  })
+
+  boot_run_spec <- purrr::map(boot_models, function(boot_run){
```
kylebaron, line 252: This is a bottle neck; for the analysis data set, it takes 1 min to resample and write the control streams and 35 sec to generate this boot spec. I wonder if there's a simpler way to do this; I think the extra time is taken by the overhead around these functions.

R/bootstrap-model.R
```diff
@@ -0,0 +1,668 @@
...
+    # New model setup
+    n_seq <- seq(n)
+    mod_names <- purrr::map_chr(n_seq, max_char = nchar(n), pad_left)
+    mod_paths <- file.path(boot_dir, mod_names)
+    orig_mod <- read_model(get_based_on(.boot_run))
+
+    # TODO: only include subjects that entered the original problem
+    starting_data <- nm_data(orig_mod) %>% suppressMessages()
+    if(!is.null(strat_cols)){
+      checkmate::assert_true(strat_cols %in% names(starting_data))
```
kylebaron, line 111: This currently fails if there are more than one stratification column 

```suggestion
      checkmate::assert_true(all(strat_cols %in% names(starting_data)))
```

R/bootstrap-model.R
```diff
@@ -0,0 +1,668 @@
...
+    strat_cols = boot_args$strat_cols,
+    replace = TRUE
+  ) %>% dplyr::rename("OID" = "ID", "ID" = "KEY") %>% # TODO: should this be a user arg?
+    dplyr::select(all_of(unique(c(names(boot_args$orig_data), "OID")))) # TODO: should this be a user arg?
+
+  mod_name <- basename(mod_path)
+  orig_mod_id <- boot_args$orig_mod_id
+
+  # Write out new dataset
+  data_boot_name <- glue("boot_{orig_mod_id}_{mod_name}.csv")
```
kylebaron, line 186: Use hyphen separator rather than underscore; but maybe this should just be `{mod_name}.csv` to keep things simple? that way the data name matches the control stream and yaml stems.

kylebaron:
When tables are requested, they have the name of the original model. We might think about updating run numbers, but _only_ if tables are requested. `.msf` files are also affected, but hate to spend time re-naming them; if we could do it quickly, I'd drop. Otherwise leave as is for now. 

```sh
106-boot/0823$ ll
total 2144
drwxr-x---    2 kyleb kyleb   4096 Apr 29 16:57 ./
drwxr-xr-x 1003 kyleb kyleb  69632 Apr 29 16:15 ../
-rw-r--r--    1 kyleb kyleb    703 Apr 29 16:57 .gitignore
-rw-r--r--    1 kyleb kyleb   1920 Apr 29 16:54 0823.clt
-rw-r--r--    1 kyleb kyleb   3493 Apr 29 16:54 0823.coi
-rw-r--r--    1 kyleb kyleb   3493 Apr 29 16:54 0823.cor
-rw-r--r--    1 kyleb kyleb   3493 Apr 29 16:54 0823.cov
-rw-r--r--    1 kyleb kyleb     13 Apr 29 16:54 0823.cpu
-rw-r--r--    1 kyleb kyleb   1291 Apr 29 16:15 0823.ctl
-rwxr-x---    1 kyleb kyleb  32307 Apr 29 16:56 0823.ctl.out*
-rw-r--r--    1 kyleb kyleb  12196 Apr 29 16:54 0823.ext
-rw-r--r--    1 kyleb kyleb   9133 Apr 29 16:53 0823.grd
-rw-r--r--    1 kyleb kyleb  54510 Apr 29 16:56 0823.lst
-rw-r--r--    1 kyleb kyleb  26859 Apr 29 16:53 0823.phi
-rwxr-x---    1 kyleb kyleb    143 Apr 29 16:19 0823.sh*
-rw-r--r--    1 kyleb kyleb    937 Apr 29 16:53 0823.shk
-rw-r--r--    1 kyleb kyleb  13511 Apr 29 16:53 0823.shm
-rw-r--r--    1 kyleb kyleb  44624 Apr 29 16:54 0823.xml
-rw-r--r--    1 kyleb kyleb   6840 Apr 29 16:54 106.msf
-rw-r--r--    1 kyleb kyleb 851908 Apr 29 16:54 106.tab
-rw-r--r--    1 kyleb kyleb  51876 Apr 29 16:54 106_ETAS.msf
-rw-r--r--    1 kyleb kyleb  27520 Apr 29 16:54 106_RMAT.msf
-rw-r--r--    1 kyleb kyleb 430248 Apr 29 16:54 106_SMAT.msf
-rw-r--r--    1 kyleb kyleb 481132 Apr 29 16:54 106par.tab
-rw-r--r--    1 kyleb kyleb    784 Apr 29 16:57 Run_0823.o1823
-rwxr-xr-x    1 kyleb kyleb   1084 Apr 29 16:15 bbi.yaml*
-rwxr-x---    1 kyleb kyleb   1695 Apr 29 16:57 bbi_config.json*
-rwxr-x---    1 kyleb kyleb    159 Apr 29 16:15 grid.sh*
```

R/bootstrap-model.R
```diff
@@ -0,0 +1,670 @@
...
+    writeLines("/data\n*.ctl\n*.mod\n*.yaml", file.path(boot_dir, ".gitignore"))
+
+    # New model setup
+    n_seq <- seq(n)
+    mod_names <- purrr::map_chr(n_seq, max_char = nchar(n), pad_left)
+    mod_paths <- file.path(boot_dir, mod_names)
+    orig_mod <- read_model(get_based_on(.boot_run))
+
+    # TODO: only include subjects that entered the original problem
+    starting_data <- nm_data(orig_mod) %>% suppressMessages()
```
kylebaron, line 110: Some possibilities: 

Could call `nm_join()` under the hood as the default action; override by user-passed data frame. 

Otherwise - the main thing we need here is a vector of the IDs that entered into the main problem; that's what we re-sample to generate the bootstrap data sets. So, if there is a table record with ID in it, we could just read in that table, grab the unique IDs, and then use that vector to filter down the `nm_data()` output and resample that data frame.

R/print.R
```diff
@@ -281,10 +281,207 @@ print.bbi_nonmem_summary <- function(x, .digits = 3, .fixed = FALSE, .off_diag =
...
+  heuristics_cols <- names(.h)[!grepl(ABS_MOD_PATH, names(.h))]
+  heuristics <- purrr::map_dfr(heuristics_cols, function(col){
+    tibble(heuristic = col, any_found = any(.h[[col]]), n_found = sum(.h[[col]]))
+  })
+
+  if (any(heuristics$any_found)) {
+    cli::cat_line("**Heuristic Problem(s) Detected:**\n", col = "red")
+    heuristics_found <- heuristics$heuristic[which(heuristics$any_found)]
+    heuristics_n <- heuristics$n_found[which(heuristics$any_found)]
+    heuristics_perc <- (heuristics_n/n_samples) * 100
```
kylebaron, line 339: I'd like to see the numbers rather than percents in the heuristic summary.  Also, I'd reframe this as a summary of the bootstrap runs rather than *alert* problems detected. 

Current output:
```r
**Heuristic Problem(s) Detected:**

– parameter_near_boundary (0.2%)
– hessian_reset (1.3%)
– has_final_zero_gradient (0.2%)
– prderr (0.1%)
```

Proposed: 

```r
Bootstrap run summary: 

- Number of runs: 1000
– parameter_near_boundary: 2 (0.2%)
– hessian_reset: 13 (1.3%)
– has_final_zero_gradient: 2 (0.2%)
– prderr: 1 (0.1%)

```

R/param-estimates-batch.R
```diff
@@ -110,51 +110,117 @@ param_estimates_batch <- function(.path,
...
+        )
+      )
+    }
+  }
+
+  # Dont pass .compare_cols here, as we can only use columns in parameter_names,
+  # which could only be the default columns.
+  param_estimates_compare(
+    .boot_sum$boot_summary, .orig_mod = .orig_mod, probs = probs, na.rm = na.rm
+  )
```
kylebaron, line 209: It would be nice to get more concise names in the output; possibly:

```r
> param_estimates_compare(boot_sum)
# A tibble: 12 × 5                                                                                                                                     
   parameter     original     p50   p2.5    p97.5
   <chr>            <dbl>   <dbl>   <dbl>   <dbl>
 1 THETA1           0.434   0.430   0.313   0.557 
 2 THETA2            4.12    4.12    4.06    4.17  
 3 THETA3            1.12    1.11    1.05    1.18  
 4 THETA4            4.21    4.21    4.18    4.25  
 5 THETA5            1.29    1.29    1.22    1.36  
 6 OMEGA(1,1)       0.221   0.215   0.117   0.329 
 7 OMEGA(2,1)      0.0690  0.0689  0.0286   0.110 
 8 OMEGA(2,2)      0.0827  0.0826  0.0631   0.103 
 9 OMEGA(3,1)       0.134   0.132  0.0881   0.181 
10 OMEGA(3,2)      0.0735  0.0731  0.0515  0.0933
11 OMEGA(3,3)       0.169   0.166   0.128   0.208 
12 SIGMA(1,1)      0.0399  0.0399  0.0375  0.0423
```

Currently has a large amount of white space between `parameter_names` and `original_estimate` due to long names:
```r
> param_estimates_compare(boot_sum)
# A tibble: 12 × 5                                                                                                                                     
   parameter_names original_estimate    p50   p2.5  p97.5
   <chr>                       <dbl>  <dbl>  <dbl>  <dbl>
 1 THETA1                     0.434  0.430  0.313  0.557 
 2 THETA2                     4.12   4.12   4.06   4.17  
 3 THETA3                     1.12   1.11   1.05   1.18  
 4 THETA4                     4.21   4.21   4.18   4.25  
 5 THETA5                     1.29   1.29   1.22   1.36  
 6 OMEGA(1,1)                 0.221  0.215  0.117  0.329 
 7 OMEGA(2,1)                 0.0690 0.0689 0.0286 0.110 
 8 OMEGA(2,2)                 0.0827 0.0826 0.0631 0.103 
 9 OMEGA(3,1)                 0.134  0.132  0.0881 0.181 
10 OMEGA(3,2)                 0.0735 0.0731 0.0515 0.0933
11 OMEGA(3,3)                 0.169  0.166  0.128  0.208 
12 SIGMA(1,1)                 0.0399 0.0399 0.0375 0.0423
```

R/bootstrap-model.R
barrettk: I have this implementation now:
```r
    # Only include subjects that entered the original problem by default
    can_be_joined <- can_be_nm_joined(orig_mod)
    if(isTRUE(can_be_joined)){
      starting_data <- nm_join(orig_mod) %>% suppressMessages()
    }else{
      rlang::warn(
        paste(
          "Defaulting to input data, which may include data that doesn't enter",
          "the final problem (i.e. ignored subjects)"
        )
      )
      starting_data <- nm_data(orig_mod) %>% suppressMessages()
    }
```

A user-provided data frame is still a possibility, but would likely come at a later stage/separate PR

R/print.R
barrettk: Id like @seth127's approval here, just because a potential consistency concern (might want to remove both or neither)

R/param-estimates-batch.R
barrettk: @seth127 I think im safe to change `original_estimate` --> `original`, but just wanted to make sure you're good with `parameter_names` --> `parameter` too (given how often we use that column name in other places). Given that this is solely an output, we definitely could, but just wanted to make sure consistency wasnt a concern here too.

R/bootstrap-model.R
seth127: I think we can definitely optimize this at some point. Thanks for pointing that out.

R/bootstrap-model.R
seth127: @barrettk making sure you don't miss this one

R/print.R
seth127: I would defer to @kylebaron on this display. I think these bootstraps are different enough from a regular model that a little inconsistency between the print output isn't a problem.

R/param-estimates-batch.R
seth127: I don't have a problem with `original_estimate` --> `original`, but I want to keep `parameter_names` because, as @barrettk mentions, that's what we call that column in the output from a few other functions. 1) I'd like to keep this consistent and 2) this facilitates easier joining, when necessary.

R/bootstrap-model.R
barrettk: FWIW I decided to address this ahead of the next dev tag. Was initially worried about the refactor messing with too many functions to risk impacting the timeline, but I saved 95% of the time by just dropping the individual problem statements and data paths from the spec file.

R/print.R
barrettk: @kylebaron let me know what you think of this structure. I moved the "number of runs" to the run summary instead of the specification section like you suggested. It does seem like a specification, but I could see it being more useful/digestible next to the heuristics. I changed the highlighting to just the values too since they are seemingly more of an expectation (to a degree) rather than a warning.

image

kylebaron: This looks great, @barrettk. Very clean summary of what was going on in the run. This is great for a first release. 

One cosmetic tweak that I might make is add some white space padding around `hessian_reset:` to get the summary numbers to line up; and might add some padding around the counts in the summary numbers. That's optional cosmetics but shouldn't be too hard to do. I think overall looks great. 

I guess one other small thing: the words `Run Summary` show up in two places pretty close together. Maybe change the second one to `Run Status` or just `Status`?

Thanks for making all the changes. 

Screenshot 2024-05-01 at 14 04 41

barrettk:
All tests passing (including new bootstrap ones) as of the latest commit ([b70f320](https://github.com/metrumresearchgroup/bbr/pull/671/commits/b70f320dc096efc3400604bae377705adb73a817)):

```r
> devtools::load_all()
ℹ Loading bbr
> devtools::test(filter = "bootstrap")
ℹ Testing bbr
✔ | F W  S  OK | Context
✔ |         71 | testing bootstrap functionality and running bbi [109.6s]                                                                                                         

══ Results ═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Duration: 109.6 s

[ FAIL 0 | WARN 0 | SKIP 0 | PASS 71 ]
> devtools::test()
ℹ Testing bbr
✔ | F W  S  OK | Context
✔ |         21 | bbr exec functions                                                                                                                                               
✔ |         71 | testing bootstrap functionality and running bbi [115.6s]                                                                                                         
✔ |         13 | checking if models are up to date [1.3s]                                                                                                                         
✔ |          9 | Collapse columns to string representation                                                                                                                        
✔ |         73 | Constructing config log from bbi_config.json [4.7s]                                                                                                              
✔ |         43 | Copying model objects [1.4s]                                                                                                                                     
✔ |          8 | copy-model-helpers                                                                                                                                               
✔ |         34 | cov-cor                                                                                                                                                          
✔ |         35 | Extract model paths from based_on fields [3.0s]                                                                                                                  
✔ |         34 | Test get_omega, get_sigma, and get_theta functions [2.9s]                                                                                                        
✔ |         67 | Build paths from model object [1.8s]                                                                                                                             
✔ |         22 | inherit-param-estimates [5.3s]                                                                                                                                   
✔ |         26 | initial-estimates [1.2s]                                                                                                                                         
✔ |          7 | model_diff() comparing models                                                                                                                                    
✔ |         19 | Test bbi summary on multiple models [1.3s]                                                                                                                       
✔ |        108 | Test bbi summary functions [4.8s]                                                                                                                                
✔ |         73 | Modify attributes of model object                                                                                                                                
✔ |         25 | Testing function to create or read in model object                                                                                                               
✔ |         35 | nm-file                                                                                                                                                          
✔ |         38 | nm-join [3.5s]                                                                                                                                                   
✔ |         26 | nm-tables                                                                                                                                                        
✔ |         30 | Test bbi batch parameter estimate functions [9.7s]                                                                                                               
✔ |         11 | Test param_estimates functions [1.3s]                                                                                                                            
✔ |         43 | test parsing labels for parameter table [2.1s]                                                                                                                   
✔ |         34 | testing print methods for bbi objects [11.1s]                                                                                                                    
✔ |          5 | read_bbi_path() helper function                                                                                                                                  
✔ |         50 | Reading NONMEM output files into R                                                                                                                               
✔ |         66 | Constructing run log from model yaml [3.0s]                                                                                                                      
✔ |         10 | submit_model(.dry_run=T)                                                                                                                                         
✔ |         14 | submit_models(.dry_run=T)                                                                                                                                        
✔ |         86 | Test creating summary logs [21.5s]                                                                                                                               
✔ |          9 | Comparing tags between models                                                                                                                                    
✔ |         47 | test_threads(.dry_run=T) [12.6s]                                                                                                                                 
✔ |         21 | tweak-initial-estimates [1.3s]                                                                                                                                   
✔ |         16 | test-use-bbi [4.8s]                                                                                                                                              
✔ |         34 | Utility functions for building args, etc.                                                                                                                        
✔ |         43 | testing a composable workflow and running bbi [248.8s]                                                                                                           

══ Results ═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Duration: 470.9 s

[ FAIL 0 | WARN 0 | SKIP 0 | PASS 1306 ]
```

R/print.R
```diff
@@ -281,10 +281,209 @@ print.bbi_nonmem_summary <- function(x, .digits = 3, .fixed = FALSE, .off_diag =
...
+  cli_h1("Bootstrap Run Summary")
+
+  # TODO: confirm this is appropriate for only_sim (unsure where this comes from)
+  only_sim <- isTRUE("only_sim" %in% names(.d))
+  if (only_sim) {
+    cat_line("No Estimation Methods (ONLYSIM)\n")
+  } else {
+    cli::cat_line("Estimation Method(s):\n")
+    purrr::walk(paste(x$estimation_method, "\n"), cli::cat_bullet, bullet = "en_dash")
+  }
```
barrettk, line 327: @seth127 this is one section I wasnt sure of and didn't _really_ attempt to address. The rest of the `TODO`s I left for now (about exposing `resample_df` arguments to users relating to ID/Key columns, your `.overwrite` argument `TODO`, and an existing one from outside of this PR):
image

barrettk:
Tests still passing after latest refactors FWIW. Bootstrap tests now happen at the end since they take so long

<details closed>
<summary>Local Test Results</summary>

```r
> devtools::test()
ℹ Testing bbr
✔ | F W  S  OK | Context
✔ |         21 | bbr exec functions                                                                                                                                               
✔ |         13 | checking if models are up to date [1.3s]                                                                                                                         
✔ |          9 | Collapse columns to string representation                                                                                                                        
✔ |         73 | Constructing config log from bbi_config.json [4.9s]                                                                                                              
✔ |         43 | Copying model objects [1.3s]                                                                                                                                     
✔ |          8 | copy-model-helpers                                                                                                                                               
✔ |         34 | cov-cor                                                                                                                                                          
✔ |         35 | Extract model paths from based_on fields [3.9s]                                                                                                                  
✔ |         34 | Test get_omega, get_sigma, and get_theta functions [3.5s]                                                                                                        
✔ |         67 | Build paths from model object [1.9s]                                                                                                                             
✔ |         22 | inherit-param-estimates [5.8s]                                                                                                                                   
✔ |         26 | initial-estimates [1.2s]                                                                                                                                         
✔ |          7 | model_diff() comparing models                                                                                                                                    
✔ |         19 | Test bbi summary on multiple models [1.8s]                                                                                                                       
✔ |        108 | Test bbi summary functions [5.3s]                                                                                                                                
✔ |         73 | Modify attributes of model object                                                                                                                                
✔ |         25 | Testing function to create or read in model object                                                                                                               
✔ |         35 | nm-file                                                                                                                                                          
✔ |         38 | nm-join [3.9s]                                                                                                                                                   
✔ |         26 | nm-tables [1.1s]                                                                                                                                                 
✔ |         30 | Test bbi batch parameter estimate functions [10.2s]                                                                                                              
✔ |         11 | Test param_estimates functions [1.4s]                                                                                                                            
✔ |         43 | test parsing labels for parameter table [2.1s]                                                                                                                   
✔ |         34 | testing print methods for bbi objects [10.0s]                                                                                                                    
✔ |          5 | read_bbi_path() helper function                                                                                                                                  
✔ |         50 | Reading NONMEM output files into R                                                                                                                               
✔ |         66 | Constructing run log from model yaml [2.3s]                                                                                                                      
✔ |         10 | submit_model(.dry_run=T)                                                                                                                                         
✔ |         14 | submit_models(.dry_run=T)                                                                                                                                        
✔ |         86 | Test creating summary logs [16.2s]                                                                                                                               
✔ |          9 | Comparing tags between models                                                                                                                                    
✔ |         47 | test_threads(.dry_run=T) [11.6s]                                                                                                                                 
✔ |         21 | tweak-initial-estimates [1.3s]                                                                                                                                   
✔ |         16 | test-use-bbi [4.6s]                                                                                                                                              
✔ |         34 | Utility functions for building args, etc.                                                                                                                        
✔ |         43 | testing a composable workflow and running bbi [253.2s]                                                                                                           
✔ |         71 | testing bootstrap functionality and running bbi [105.1s]                                                                                                         

══ Results ═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Duration: 461.3 s

[ FAIL 0 | WARN 0 | SKIP 0 | PASS 1306 ]
```
</details>

R/bootstrap-model.R
```diff
@@ -0,0 +1,681 @@
+#' Create a boostrap model object from an existing model
+#'
+#' @param .mod a `bbr` model object
+#' @param .suffix a suffix for the bootstrap run directory. Will be prefixed by
+#'  the model id of `.mod`.
+#' @inheritParams copy_model_from
```
seth127, line 6: We need to define `.inherit_tags` here too, because we inherit from `copy_model_from()`, which says "If FALSE, the default, new model will..."

That is correct for `copy_model_from()` but for `new_bootstrap_run()` this defaults to `TRUE`.

Related: I do like this defaulting to `TRUE`, even though `copy_model_from()` doesn't. We just need the docs to reflect that.

R/bootstrap-model.R
```diff
@@ -0,0 +1,681 @@
...
+#' @param remove_cov,remove_tables Logical (T/F). Optionally remove `$COVARIANCE`
```
seth127, line 7: Note in this description that both of these default to `TRUE`.

R/bootstrap-model.R
```diff
@@ -0,0 +1,681 @@
...
+  }
+
+  # Return read-in model to get the updated class as part of the model object
+  return(read_model(file.path(model_dir, boot_dir)))
+}
+
+
+#' Set up a bootstrap model run.
+#'
+#' Creates a new model object and re-sampled dataset per model run.
```
seth127, line 59: `dataset` should be `datasets`

R/bootstrap-model.R
```diff
@@ -0,0 +1,681 @@
...
+#' # Submit
+#' submit_model(.boot_run)
+#'
+#' # Check status of runs during submission
+#' get_model_status(.boot_run)
+#' }
+#' @export
+setup_bootstrap_run <- function(
+    .boot_run,
+    n = 100,
```
seth127, line 86: Per user feedback, change this default to `200`

barrettk: I thought that was just for the specific project template script? Do we want to confirm that value internally too?

R/bootstrap-model.R
barrettk: I think dataset is right actually, though I am now realizing my use of "run" is a bit ambiguous in a few places. It's meant to suggest that one new model and one new dataset are created per "model run". Where a "bootstrap run" is a collection of [x] "model runs". 

Let me know how I should change it based on that^

R/param-estimates-batch.R
```diff
@@ -110,51 +110,117 @@ param_estimates_batch <- function(.path,
...
+      )
+      if (fs::file_exists(orig_config_path)) {
+        res <- check_up_to_date(.orig_mod)
+        if(any(res[!res])){
+          rlang::warn("The original model is not up to date")
+        }
+      }else{
+        # Set the original model to NULL if the run directory/config file didnt
+        # exist
+        rlang::warn("The original model has not been run yet. Cannot compare estiamtes.")
```
seth127, line 192: typo: `estiamtes`

R/param-estimates-batch.R
```diff
@@ -110,51 +110,117 @@ param_estimates_batch <- function(.path,
...
+  # which could only be the default columns.
+  param_estimates_compare(
+    .boot_sum$boot_summary, .orig_mod = .orig_mod, probs = probs, na.rm = na.rm
+  )
+}
+
+
+#' @rdname param_estimates_compare
+#' @export
+param_estimates_compare.default <- function(
```
seth127, line 215: Not totally necessary, since this isn't changing current behavior, but it would be nice to do some input checking on `.boot_sum` in this `default` method. Maybe just that it's something that inherits `data.frame` and that it has the columns we expect?

R/print.R
seth127: Hmm, I'm not sure about that either. I assume it comes from copying [this section from `bbi_nonmem_summary`](https://github.com/metrumresearchgroup/bbr/blob/e502d1a44fe4d326909fa9fafebfdde8bde5c034/R/print.R#L207-L214) above.

That section was added in https://github.com/metrumresearchgroup/bbr/commit/bbe68ed0fbb3a5bb299cf769d795c8208b1c2226 and the commit message explains the rationale pretty well.

In terms of what to do here... You may want to check, but I don't think it's possible to have gotten to this point (i.e. a `bbi_nmboot_summary` object that would trigger this print method) with an `ONLYSIM` model. I feel like it would fail somewhere earlier in the bootstrap pipeline. It's probably worth testing what happens when you feed a model with `$SIM` and no `$EST` to `new_bootstrap_model()`...

If I'm right though, you can probably delete this check here.

Related: you probably want to revisit those [those `bbi_nonmem_summary` lines](https://github.com/metrumresearchgroup/bbr/blob/e502d1a44fe4d326909fa9fafebfdde8bde5c034/R/print.R#L207-L214) as part of #687

R/submit-model.R
seth127: I would consider setting up an `"bbr.bbi_nmboot_exe_mode"` option at some point, but I don't think it's needed for now.

In terms of the default... I guess maybe `"sge"` is safest. If someone has this set to `"local"` and they submit models in batches of 100... I guess it's likely to freeze up their machine.

And if they _want_ to submit locally, they can set this in their call (and probably reset `batch_size` too).

@barrettk if you agree, can you change this default to `.mode = "sge"`?

seth127 (changes requested):
I haven't gotten to review the tests yet, or all of the setup code, so there may be some more changes. That said, there are a few things here that I thought I should put in now.

R/modify-records.R
```diff
@@ -0,0 +1,151 @@
...
+    }
+  }else{
+    if(!is.null(prob_text)){
+      prob_str$value <- prob_text
+    }
+    prob_str_return <- prob_str$value
+  }
+
+  # write control stream and return problem statement
+  nmrec::write_ctl(ctl, mod_path)
```
kyleam, line 74: Why does this write to disk when `prob_text = NULL`? That seems unnecessary and based on the docstring ("If \`NULL\` return the current \`$PROBLEM\` statement"), I wouldn't expect it.

R/modify-records.R
```diff
@@ -0,0 +1,151 @@
...
+  data_rec <- nmrec::select_records(ctl, "data")[[1]]
+  data_rec$parse()
+
+  # Overwrite 'filename' option
+  data_rec$values <- purrr::map(data_rec$values, function(data_opt){
+    if(inherits(data_opt, "nmrec_option_pos") && data_opt$name == "filename"){
+      data_opt$value <- data_path
+    }
+    data_opt
+  })
```
kyleam, line 98: Rather than the `map`, I think it'd be simpler to use something like (untested)

```R
opt <- get_record_option(..., "filename")
if (is.null(opt)) {
  dev_error(...)
}

opt$value <- data_path
```

R/modify-records.R
```diff
@@ -0,0 +1,151 @@
...
+  ctl <- safe_read_ctl(.mod)
+
+  # Get data record
+  data_rec <- nmrec::select_records(ctl, "data")[[1]]
+  data_rec$parse()
+
+  # Overwrite 'filename' option
+  data_rec$values <- purrr::map(data_rec$values, function(data_opt){
+    if(inherits(data_opt, "nmrec_option_pos") && data_opt$name == "filename"){
+      data_opt$value <- data_path
```
kyleam, line 95: I haven't traced the calling code path, but do we need to ever worry about quoting the file path here (e.g., spaces in the name)?

R/modify-records.R
```diff
@@ -0,0 +1,151 @@
...
+#' @note
+#' To add support for more record types to `remove_records()`, run the following
+#' command to see the record types of a given model supported by `nmrec`:
+#'  ```
+#'  purrr::map_chr(ctl$records, "name")
+#'  ```
+#'
+#' @keywords internal
+remove_records <- function(.mod, type = c("covariance", "table")){
+  type <- match.arg(type)
```
kyleam, line 122: Unrelated to this feature and I know we use `match.arg()` elsewhere in the code base, but my suggestion would be to use `rlang::arg_match()` to disallow partial matches.

R/modify-records.R
```diff
@@ -0,0 +1,151 @@
...
+
+  rec_indices <- seq_along(ctl$records)
+  indices_remove <- purrr::keep(rec_indices, function(index){
+    ctl$records[[index]]$name == type
+  })
+
+  # Remove records
+  if(length(indices_remove) >= 1){
+    ctl$records[indices_remove] <- NULL
+  }
```
kyleam, line 133: The removal logic seems a bit more complicated than it needs to be. Can't lines 125-133 be reduced to something like this?

```R
todrop <- purrr::map_lgl(ctl$records, function(r) identical(r[["name"]], type))
ctl$records[todrop] <- NULL
```

R/modify-records.R
```diff
@@ -0,0 +1,151 @@
...
+
+  # Write out modified ctl
+  mod_path <- get_model_path(.mod)
+  nmrec::write_ctl(ctl, mod_path)
+  return(invisible(TRUE))
```
kyleam, line 138: It doesn't cause any issues, but I don't understand why `TRUE` is returned here.

kyleam (dismissed):
This is an intial review of `R/modify-records.R` at 1a888ad1 ([requested](https://github.com/metrumresearchgroup/bbr/pull/671#discussion_r1608942559) by @seth127).

I've made various comments, but here are the main requests for changes:

-   don't let `safe_read_ctl` send error objects to downstream code that expects an `nmrec_ctl_records` object

-   avoid invalid operator error in `modify_prob_statement`

R/modify-records.R
```diff
@@ -0,0 +1,151 @@
+#' Modify options and records from a `NONMEM` control stream file
```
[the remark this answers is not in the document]
kyleam: @seth127 sure, review of that file is here: https://github.com/metrumresearchgroup/bbr/pull/671#pullrequestreview-2071647847

barrettk:
@kyleam still reading through your comments, but just wanted to mention that I had spoke to @seth127 previously about making any changes to `modify-records.R` in the vpc PR since I made  changes on top of this there. From what I've read so far your feedback is still relevant, but wanted to let you know that I would be addressing the feedback in the VPC PR.

R/model-diff.R
```diff
@@ -44,7 +44,7 @@ model_diff <- function(.mod, .mod2 = NULL, .file = "model", ..., .viewer = FALSE
 
 #' @rdname model_diff
 #' @export
-model_diff.bbi_nonmem_model <- function(
+model_diff.default <- function(
```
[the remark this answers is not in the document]
barrettk: I actually made this change in the vpc PR...The latest dev tag will include those changes, but probably doesnt make sense to fix this here as well

R/param-estimates-batch.R
barrettk: I added a data.frame check, but I dont think I can check the columns. It seems like I cant interact with `.compare_cols` if it's not a character vector (I thought it would be an expression, but it just errors out):

In the function:
```r
param_estimates_compare.default <- function(
    .boot_sum,
    .orig_mod = NULL,
    .compare_cols = starts_with(c("THETA", "SIGMA", "OMEGA")),
    probs = c(.5, 0.025, 0.975),
    na.rm = FALSE
){
```

```r
Browse[2]> .compare_cols
Error: `starts_with()` must be used within a *selecting* function.
ℹ See <https://tidyselect.r-lib.org/reference/faq-selection-context.html> for details.

Browse[2]> inherits(.compare_cols, "character")
Error: `starts_with()` must be used within a *selecting* function.
ℹ See <https://tidyselect.r-lib.org/reference/faq-selection-context.html> for details.
In addition: Warning messages:
1: In get(object, envir = currentEnv, inherits = TRUE) :
  restarting interrupted promise evaluation
2: In inherits(.compare_cols, "character") :
  restarting interrupted promise evaluation
```

tests/testthat/test-workflow-bootstrap.R
```diff
@@ -0,0 +1,302 @@
...
+
+    # copy model file into new model dir
+    fs::file_copy(CTL_TEST_FILE, MODEL_DIR_BBI)
+
+    # Create base model to bootstrap with (not yet submitted)
+    mod1 <- new_model(
+      file.path(MODEL_DIR_BBI, "1"),
+      .description = "original test-bootstrap model",
+      .tags = ORIG_TAGS,
+      .bbi_args = list(threads = 4, parallel = TRUE)
```
seth127, line 63: These simple models sometimes fail when parallelized. Unless there's a reason to keep them, I would remove these `.bbi_args`, just to avoid flakiness.

kylebaron:
Not sure how this is happening / where this is coming from:

```r
>   boot_sum <- summarize_bootstrap_run(boot_run)  
Error in `dplyr::bind_rows()` at purrr/R/superseded-map-df.R:69:3:                                               
! Can't combine `..1$version` <character> and `..552$version` <double>.
Run `rlang::last_trace()` to see where the error occurred.
```

kylebaron:
Requesting better error message when we try to stratify on a nonexistent column: 
```r
> boot_run <- setup_bootstrap_run(boot_run, n = 1000, strat_cols = 
+                                   c("STUDY", "RF2"))
Error in setup_bootstrap_run(boot_run, n = 1000, strat_cols = c("STUDY", :
Assertion on 'all(strat_cols %in% names(starting_data))' failed: Must be TRUE.
```

barrettk:
@kylebaron as of the last commit:
```r
> setup_bootstrap_run(.boot_run, n = 10, strat_cols = c("SEXf", "ETNf"), .overwrite = T)
Error in `setup_bootstrap_run()`:                                                                                                                                                               
! The following `strat_cols` are missing from the input data: SEXf, ETNf
Run `rlang::last_trace()` to see where the error occurred.
```

seth127 (approved):
This is all looking good. A few of the requested changes have been addressed on the VPC branch (https://github.com/metrumresearchgroup/bbr/pull/687) because of refactors that had already been made there. Because of that, I think we should go ahead and merge this to `main`. 

@barrettk may make a few more comments on here, to point to changes that were mentioned here but made on the other branch.

R/modify-records.R
```diff
@@ -0,0 +1,151 @@
...
+    idx <- purrr::detect_index(
+      prob_rec$values,
+      function(x) inherits(x, "nmrec_option_record_name")
+    )
+    if(idx == 0){
+      dev_error("Unsupported $PROBLEM record format")
+    }
+    if(!is.null(prob_text)){
+      prob_rec$values <- append(prob_rec$values, paste0(" ", prob_text), idx)
+      prob_str_return <- prob_rec$values[[idx + 1]]$format()
```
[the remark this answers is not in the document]
barrettk: Addressed in [b7fc76e](https://github.com/metrumresearchgroup/bbr/pull/687/commits/b7fc76e173d8433094039f4a28ee90e1d4b5a084)

R/modify-records.R
```diff
@@ -0,0 +1,151 @@
...
+#' @keywords internal
+safe_read_ctl <- function(.mod){
+  check_model_object(.mod, c(NM_MOD_CLASS, NMBOOT_MOD_CLASS))
+  mod_path <- get_model_path(.mod)
+  ctl <- tryCatch(nmrec::read_ctl(mod_path), error = identity)
+  if(inherits(ctl, "error")){
+    warning(
+      glue::glue("Could not read control stream file. Reason: {ctl$message}")
+    )
+  }
```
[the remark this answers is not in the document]
barrettk: Addressed in [4f51be1](https://github.com/metrumresearchgroup/bbr/pull/687/commits/4f51be1c76d8260170288b29f9fa9b90cc082987)

 - It's now just a convenience wrapper for getting the control stream from a model object (rather than the model path)

R/modify-records.R
```diff
@@ -0,0 +1,151 @@
...
+      function(x) inherits(x, "nmrec_option_record_name")
+    )
+    if(idx == 0){
+      dev_error("Unsupported $PROBLEM record format")
+    }
+    if(!is.null(prob_text)){
+      prob_rec$values <- append(prob_rec$values, paste0(" ", prob_text), idx)
+      prob_str_return <- prob_rec$values[[idx + 1]]$format()
+    }else{
+      prob_str_return <- prob_rec$values[[idx]]$format()
```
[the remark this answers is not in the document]
barrettk: Could have done it a different way (rather than `NULL`), but I was ultimately trying to reduce the number of times we do certain checks (such as a newline problem statement or checking for multiple statements), and therefore opted to contain the reading and modification within the same function. When it comes to the purpose of the reading: the intention was to pull each of these statements out for storing in the bootstrap spec, including in a `run_log`, etc.

R/modify-records.R
barrettk: Good point - will address

R/modify-records.R
barrettk: I thought about this. I dont think so. We have that handling in [`get_data_path_from_ctl()`](https://github.com/metrumresearchgroup/bbr/blob/cdc6433ab2f714a1e4dca7b663a889143cbb0eaf/R/get-path-from-object.R#L361-L362), but here we always specify the string exactly how we want to, and the current path should be irrelevant.

R/modify-records.R
```diff
@@ -0,0 +1,151 @@
...
+#' control stream file
+#'
+#' @param data_path Data path to set in a `$DATA` record.
+#'
+#' @keywords internal
+modify_data_path_ctl <- function(.mod, data_path){
+  ctl <- safe_read_ctl(.mod)
+
+  # Get data record
+  data_rec <- nmrec::select_records(ctl, "data")[[1]]
```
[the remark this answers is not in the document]
barrettk: Addressed in [528a552](https://github.com/metrumresearchgroup/bbr/pull/687/commits/528a5528f838b85a539498757044d8ebb8c48d94)

R/modify-records.R
barrettk: Addressed in [528a552](https://github.com/metrumresearchgroup/bbr/pull/687/commits/528a5528f838b85a539498757044d8ebb8c48d94)

R/modify-records.R
barrettk: I actually got rid of the `match.arg` method, in place of supporting all spelling variants `nmrec` does. I did so via [this function](https://github.com/metrumresearchgroup/bbr/blob/528a5528f838b85a539498757044d8ebb8c48d94/R/modify-records.R#L309-L335).

If you take a look at [this example](https://github.com/metrumresearchgroup/bbr/blob/528a5528f838b85a539498757044d8ebb8c48d94/R/modify-records.R#L149-L152), you can see that I use the name `nmrec` stores as the identifier when making any modifications. Was originally envisioning a helper function to extract that name given a supported spelling, but this convention worked well without overcomplicating things

R/modify-records.R
barrettk: addressed in [dde8c7d](https://github.com/metrumresearchgroup/bbr/pull/687/commits/dde8c7df37f320e755367ee24ce258d602cba16b)

tests/testthat/test-workflow-bootstrap.R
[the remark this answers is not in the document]
barrettk: Addressed in [84b1e2d](https://github.com/metrumresearchgroup/bbr/pull/687/commits/84b1e2db33b72e06a3ad391deac6b057f723a200)

Also addressed most of the other `modify-records.R` changes requested by @kyleam in https://github.com/metrumresearchgroup/bbr/pull/687. Relevant commits were attached to each comment, though I know the changes in the other PR could lead to additional feedback

Issue #550, opened 2022-08-19
drone: pkgdown check


kyleam:
Nearly every release we end up needing do some combination of adding entries to `_pkgdown.yml` and marking functions with `@keywords internal`.  To catch these before a PR is merged, I suggest we add a [`check_pkgdown`](https://pkgdown.r-lib.org/reference/check_pkgdown.html) step to Drone.

Issue #629, opened 2023-12-13
Remove `mrgvalidate` and `mrgvalprep` from DESCRIPTION


seth127:
We are no longer using either of these, so we should remove them from `Suggests`.

I noticed that they were still there when filing #628

kyleam:
Give the switch in validation approach, I think (didn't verify latest state) a good number of our packages have this (unsurpsingly).  We've got an internal ticket to clean up mrgval* related things across packages (275).  (Of course a cleanup just for bbr is fine, but seems worth mentioning in case whoever takes this on is up for doing a wider sweep.)

PR #681, opened 2024-04-29
ci: switch to GitHub Actions

kyleam:
This series switches the CI from Drone to GitHub Actions.  The goal is to preserve important features, including coverage measurement and the "oldest" job.

The main switch happens in the third commit.  The fourth commits takes care of gh-550 by adding a `check_pkgdown()` step.

---

- [x] Drop tip commit that tests dry-run s3 sync.
- [x] After approval and right before merge, adjust the required status checks in the branch protection settings.

- [ ] Coordinate with open pull requests.  Once this is merged, these PRs should bring this in (by merge or rebase) to move away from Drone.


kyleam:
Test s3 auth and dry-run upload look good:

 * https://github.com/metrumresearchgroup/bbr/actions/runs/8886837649/job/24401107031?pr=681
 * 3aa36d8cf4520400c96e87abf7d397457788a5cc

I'll drop the tip commit.

.github/workflows/main.yaml
```diff
@@ -0,0 +1,127 @@
...
+      - 'main'
+      - 'scratch/**'
+    tags:
+      - '[0-9]+.[0-9]+.[0-9]+'
+      - '[0-9]+.[0-9]+.[0-9]+.[0-9]+'
+  pull_request:
+
+env:
+  BBI_VERSION: v3.3.0
+  MPN: 2024-03-01
```
kyleam, line 14: This hard coding of the latest MPN version is unfortunate.  Drone, instead, just uses the `:latest` and `:cran-latest` Docker tags, and those get updated behind the scenes.  Each package doesn't need to handle the update.

In the near future (ideally the next MPN release), I think as part of publish the MPN site, we should update something like `https://mpn.metworx.com/latest` to provide an easy way to programmatically get the latest MPN version.  (There is a way we've used internally for one script, but it's really convoluted, and I'd rather not propagate it to the GitHub Actions config for each of our R packages.)

barrettk:
Drive by comment: We have an open issue about dropping those two packages that can be closed once this is merged:
https://github.com/metrumresearchgroup/bbr/issues/629

kyleam:
> We have an open issue about dropping those two packages that can be closed once this is merged:
> #629

Thanks for the reminder.  Updated that commit to auto-close that issue.

.github/workflows/main.yaml
kyleam: Update to query new `https://mpn.metworx.com` path.

kyleam:
Pushed update to pull shared actions from new repo at <https://github.com/metrumresearchgroup/actions>.

kyleam:
I'll return this to a draft because 1) the plan is to hold off on this until after the upcoming release and 2) I'll need to resolve the conflict and make a tweak to the oldest build for the addition of mrgmisc dependency.

For reference, here's a scratch actions run that brings in the current tip of `feat/vpc` (f81cf2d4): https://github.com/metrumresearchgroup/bbr/actions/runs/9504620205

kyleam:
I've rebased and resolved the conflicts.

Notable changes:

 * The 4.0.5 and 4.1.3 builds have been updated to 4.2.3 and 4.3.1 to match the two most recent R versions on the latest AMI.  R 4.0.5 is still tested in the "oldest" job.

 * The **oldest** build now uses the **2023-01-25** snapshot.  An increase is required due to functionality that bbr v1.11.0 started using, and this particular snapshot is required to get purrr v1.0.0.

   Relevant commits:
   * e6708271 (DESCRIPTION: specify minimum version for cli)
   * 34c469ef (DESCRIPTION: specify minimum version for withr)
   * 4df2765d (DESCRIPTION: specify minimum version for purrr)
   * c6cee200 (ci: increase 'oldest' build to MPN 2023-01-25)

kyleam:
I said:
> The **oldest** build now uses the **2023-01-25** snapshot

I think everyone is agreement that we are long overdue for a bump here.  There may be some disagreement about whether 2023-01-25 is too recent.  However, at this point, this _is_ the effective minimum snapshot since the 1.11.0 release.

So my vote would be for discussion of that to not hold up merging this PR.  We can create a dedicated issue to discuss whether we want to make changes to push back the minimum required snapshot.

barrettk:
@kyleam I went through all your recent commits/refactors and agree with your perspective on opening a dedicated PR if we wanted to dial back the oldest snapshot/revisit version requirements. A shame we missed some of these things by deleting the `mpn-oldest` checks, but im also glad we're finally bumping the required `cli` version (`purrr` bump is nice too). Fingers crossed those two dont get dialed back, though I agree the oldest snapshot does feel a bit recent (I personally have no objections given how MPN works though).

.github/workflows/main.yaml
```diff
@@ -0,0 +1,114 @@
...
+        run: |
+          # Even on oldest, install new enough nmrec for vignettes to
+          # build.
+          version=0.4.0
+          tdir=$(mktemp -d)
+          git clone --branch "$version" --depth 1 \
+            https://github.com/metrumresearchgroup/nmrec.git "$tdir"
+          # Delete testthat version constraint.  It doesn't matter in
+          # this context and causes pak's solver to abort.
+          sed -i 's/testthat (.*/testthat,/' "$tdir/DESCRIPTION"
```
barrettk: Can you explain this chunk? Why it was needed and what it's doing? Are you removing the testthat constraint from `nmrec`?

.github/workflows/main.yaml
```diff
@@ -0,0 +1,114 @@
...
+          echo "NMREC_PKG=local::$tdir" >>$GITHUB_ENV
+      - uses: r-lib/actions/setup-r-dependencies@v2
+        with:
+          extra-packages: |
+            any::pkgdown
+            any::rcmdcheck
+            ${{ env.MRGMISC_PKG }}
+            ${{ env.NMREC_PKG }}
+          upgrade: ${{ matrix.config.label == 'oldest' && 'FALSE' || 'TRUE' }}
+      - uses: r-lib/actions/check-r-package@v2
```
barrettk, line 82: Can you explain the `&& 'FALSE' || 'TRUE'` part of this? Im not sure what you're actually evaluating to be `TRUE` or `FALSE`. Have a few thoughts on what it could be but just wanted to clarify

.github/workflows/main.yaml
```diff
@@ -0,0 +1,114 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - os: ubuntu-20.04
+            r: 4.2.3
+          - os: ubuntu-20.04
+            r: 4.3.1
+          - os: ubuntu-latest
+            r: release
```
barrettk, line 28: How does `r: release` work? Is this set somewhere else?

.github/workflows/main.yaml
```diff
@@ -0,0 +1,114 @@
...
+            any::rcmdcheck
+            ${{ env.MRGMISC_PKG }}
+            ${{ env.NMREC_PKG }}
+          upgrade: ${{ matrix.config.label == 'oldest' && 'FALSE' || 'TRUE' }}
+      - uses: r-lib/actions/check-r-package@v2
+      - name: Check pkgdown
+        # The pkgdown on MPN 2020-06-08 does not have check_pkgdown().
+        if: matrix.config.label != 'oldest'
+        shell: Rscript {0}
+        run: pkgdown::check_pkgdown()
```
barrettk, line 87: Is this accurate still since you updated oldest to be a 2023 snapshot?

.github/workflows/main.yaml
kyleam, line 75: @barrettk was asking for more details here, which I agree would be good to provide..  I pushed a scratch branch to document the failure, but it now passes: https://github.com/metrumresearchgroup/bbr/actions/runs/9700059054/job/26770553048

And that makes sense because we bumped the oldest snapshot from 2020-06-08 to 2023-01-25.  I'll drop this kludge.

barrettk (commented):
Left some comments

.github/workflows/main.yaml
kyleam: Good catch, another piece that should have been updated with the MPN update.

.github/workflows/main.yaml
barrettk: (Saw this is outdated from your previous commit)

.github/workflows/main.yaml
kyleam: Yes, it's fed to as the `r-version` input to `r-lib/actions/setup-r@v2` below.

https://github.com/metrumresearchgroup/bbr/blob/72b1d4a12aa0b31d5cc3217040d62981dd6aa188/.github/workflows/main.yaml#L46-L48

https://github.com/r-lib/actions/blob/fbafc3bc4ba114e72680c71e835c59b022606c46/setup-r/action.yml#L1-L10

.github/workflows/main.yaml
kyleam: Sure.  If the first condition is true (the label is "oldest"), `'FALSE'` is the result.  Otherwise it falls back to `'TRUE'`.

Here's an example from the Actions docs that covers this: https://docs.github.com/en/actions/learn-github-actions/expressions#example

.github/workflows/main.yaml
kyleam: > Can you explain this chunk? Why it was needed and what it's doing?

pak's dependency solver was aborting given nmrec's `testthat (>= 3.0.0)` constraint.  In other words, the `r-lib/actions/setup-r-dependencies@v2` step would fail.  I'll see if I can dig up the exact failure, though we may be past the retention period for the logs.

> Are you removing the testthat constraint from `nmrec`?

Yes.  As the comment mentions, it doesn't matter in this context.  That's a constraint for nmrec's tests, and we aren't running those.

kyleam: > I'll see if I can dig up the exact failure, though we may be past the retention period for the logs.

Yes, logs from those periods are expired already.

But I think it essentially just comes down to:

 * we were setting repos to be from the 2020-06 period (via MPN and RSPM)
 * testthat 3.0.0 isn't available until 2020-10

Given the above, pak was rightly failing saying it couldn't fulfill nmrec's requirement for testthat 3.0.0 or above.

.github/workflows/main.yaml
barrettk: Thanks for the link. I understand the ternary operator now, though am still bewildered why `&&` is used for that, rather than a `?` or something I saw in other examples (outside of github actions). Guess that's just something I have to accept lol (not saying there isnt a good reason, just wasnt intuitive to me).

.github/workflows/main.yaml
```diff
@@ -0,0 +1,109 @@
...
+      - uses: metrumresearchgroup/actions/mpn-latest@v1
+      - uses: r-lib/actions/setup-r@v2
+        with:
+          r-version: release
+          use-public-rspm: true
+          extra-repositories: 'https://mpn.metworx.com/snapshots/stable/${{ env.MPN_LATEST }}'
+      - uses: r-lib/actions/setup-r-dependencies@v2
+        with:
+          extra-packages: any::pkgpub
+      - uses: metrumresearchgroup/actions/publish-r-release@v1
```
barrettk, line 109: I just wanted to get clarity here:

I see that we run `r-lib/actions/setup-r-dependencies@v2` for the release, though it seems like we only explicitly check out `mrgmisc` and `nmrec` if `matrix.config.label == 'oldest'`. I realize those packages _are on_ `mpn-latest`, but just wanted to confirm if that's what was happening here.

.github/workflows/main.yaml
kyleam: > just wasnt intuitive to me

Fair enough.  Think it's just a matter of what languages you're familiar with.

Similar behavior in python:

```python
>>> True and "x" or "y"
'x'
>>> False and "x" or "y"
'y'
```

barrettk (approved):
Looks good, thanks for the changes and replies. Left one more point of clarity

.github/workflows/main.yaml
kyleam: Yes, that's correct.

kyleam:
@barrettk Thanks for the careful review.

PR #700, opened 2024-06-07
Fixes from testing on project model

kylebaron:
- Convert `NA` to `.` when writing user-passed data
- Always simulate with `ONLYSIMULATION`


barrettk (approved):
Changes look good - want to mention the addition of the `ONLYSIMULATION` flag in the docs, but will do so in the parent PR

PR #708, opened 2024-06-28
add bootstrap and simulation vignettes

barrettk:


seth127:
@barrettk this looks good to me overall.

@kylebaron would you be up for reviewing this, specifically considering how it compliments and/or overlaps upcoming updates to the [Expo 1 bootstrap article](https://merge.metrumrg.com/expo/expo1-nonmem-foce/posts/bootstrap.html)?

My general thinking is:

* This vignette should be more "technical" in terms of showing examples of the function syntax and describing what it does and how the functions fit together (i.e. `new_` -> `setup_` -> `submit_` -> `summarize_` -> `cleanup_`)
* The Expo article should be demonstrating how to do an end-to-end bootstrap, using this functionality, in the context of a (semi-)realistic analysis.

That said, if there is a lot of overlap, I am fine with getting rid of this vignette and have the Expo article stand alone to compliment the [function reference docs](https://metrumresearchgroup.github.io/bbr/reference/index.html#bootstrap-and-simulation). As we've all discussed before, I prefer to _not_ have to maintain the same information in multiple places.

Thanks to both of you for all of the work on this feature.

_(By the way, I have similar feeling about a vignette for the new simulation feature: depending on what ends up in the Expo, I'm not sure a dedicated vignette is necessary.)_

PR #707, opened 2024-06-26
`data` argument to `setup_bootstrap_run` and remove `nm_join` requirement

barrettk:
We use `nm_join()` under the hood to create the _filtered_ starting dataset we resample from. There was previously no method for passing a `.join_col`, so the default `NUM` would always be used under the hood.

The workaround was to add a `NUM` column to the original model and resubmit _before_ attempting to bootstrap it, though this is obviously inconvenient and shouldn't be the behavior. Note that a support ticket has already been filed for this issue, so this PR was initially in response to that.

Rather than providing a new `.join_col` argument, this PR removes the `nm_join()` dependency by parsing the `IGNORE` and `ACCEPT` options in a `$DATA` record, returning a filtered dataset.

Since we had to touch this function anyways, I reached out to @kylebaron about adding a `data` argument as well. There have been [previously discussions](https://github.com/metrumresearchgroup/bbr/issues/678) about this, though mostly offline. I followed the same approach used for `nmsim` models (save new data to the output directory after some checks) with some minor differences.

PR #711, opened 2024-07-02
Parse $DATA record for the purpose of filtering the input data

barrettk:
Parse `$DATA` record for the purpose of filtering the input data
 - Extracts and formats `IGNORE` and `ACCEPT` record options, transforming them into `dplyr::filter()` expressions
 - adds `filter` argument to `nm_data()`
 - adjusts `setup_bootstrap_run` to use this method instead of `nm_join()`. Additional details will be documented in https://github.com/metrumresearchgroup/bbr/pull/707


R/modify-records.R
```diff
@@ -323,6 +323,131 @@ get_table_columns <- function(.mod, from_data = TRUE){
...
+  return(expr)
+}
+
+
+#' Filter `NONMEM` input data based on `IGNORE` and `ACCEPT` record options
+#'
+#' @param .mod a `bbi_nonmem_model` object
+#' @param data a starting dataset
+#' @keywords internal
+filter_nm_data <- function(.mod, data = nm_data(.mod)){
```
kylebaron, line 358: Suggest to break this function down a bit more. Especially the code that processes `$DATA` block. If you isolate that into its own function, it will be _much_ easier to test.

R/modify-records.R
```diff
@@ -323,6 +323,131 @@ get_table_columns <- function(.mod, from_data = TRUE){
...
+    rlang::abort(
+      c(
+        glue::glue("Expected a single data record, but found {n_data}:\n\n"),
+        recs_fmt
+      )
+    )
+  }
+  data_recs[[1]]$parse()
+
+  # Extract & format IGNORE options
```
kylebaron, line 373: Probably starting here?

R/modify-records.R
```diff
@@ -323,6 +323,131 @@ get_table_columns <- function(.mod, from_data = TRUE){
...
+  # Only supports `ACCEPT=(list)` form
+  accept_filters <- purrr::map_chr(accept_exprs, function(expr) {
+    return(paste0(expr))
+  })
+
+  # Combine all filter expressions
+  all_filters <- c(ignore_filters, accept_filters)
+
+  # Create the final dplyr::filter expression
+  filter_expression <- paste(all_filters, collapse = " & ")
```
kylebaron, line 435: Then return this filter expression

R/modify-records.R
```diff
@@ -323,6 +323,131 @@ get_table_columns <- function(.mod, from_data = TRUE){
...
+
+  # Translate the ignore and accept expressions
+  ignore_exprs <- lapply(ignore_vals, translate_operator) %>% gsub(",", " &", .)
+  accept_exprs <- lapply(accept_vals, translate_operator) %>% gsub(",", " &", .)
+
+  ## Combine the expressions into filter expressions ##
+  data_cols <- get_input_columns(.mod)
+
+  # `IGNORE=#`, `IGNORE=@`, `IGNORE=c1`, `IGNORE=(list)`
+  ignore_filters <- purrr::map_chr(ignore_exprs, function(expr) {
```
kylebaron, line 398: Maybe this is the function you want to split out? Just processing a single expr.

R/modify-records.R
```diff
@@ -323,6 +323,131 @@ get_table_columns <- function(.mod, from_data = TRUE){
...
+    if(expr == "#"){
+      # IGNORE=# is the default.  That is, in the absence of IGNORE option, any
+      # record whose first character is '#' is treated as a comment record.
+      col_filters <- purrr::map_chr(data_cols, function(col) {
```
kylebaron, line 402: I think you only want to check `data_cols[1]`; this should only look at the first character of each record (row).

R/modify-records.R
```diff
@@ -323,6 +323,131 @@ get_table_columns <- function(.mod, from_data = TRUE){
...
+        paste0("!grepl('^#', ", col, ")")
+      })
+      return(paste(col_filters, collapse = " & "))
+    }else if(expr == "@"){
+      # IGNORE=@ signifies that any data record having an alphabetic character
+      # or `@` as its first non-blank character (not just in column 1)
+      # should be ignored. This permits a table file having header lines to be
+      # used as an NM-TRAN data set.
+      col_filters <- purrr::map_chr(data_cols, function(col) {
```
kylebaron, line 411: The documentation says "not just in column 1", but for this application I think we are still looking at column 1; I don't know how looking at all data cols will work here.

kylebaron: The filters for `inst/models/nonmem/basic/1.ctl` come back as: 

```
[1] "!grepl('^[A-Za-z@]', ID) & !grepl('^[A-Za-z@]', TIME) & !grepl('^[A-Za-z@]', MDV) & !grepl('^[A-Za-z@]', EVID) & !grepl('^[A-Za-z@]', DV) & !grepl('^[A-Za-z@]', AMT) & !grepl('^[A-Za-z@]', SEX) & !grepl('^[A-Za-z@]', WT) & !grepl('^[A-Za-z@]', ETN) & !grepl('^[A-Za-z@]', NUM)"
[2] "ID!=2"     
```

The first filter will drop any row that contains a column with alpha or `@`.

R/modify-records.R
barrettk: Just seeing this but coincidentally did just that

R/modify-records.R
barrettk: Oh I think you're right. I thought the `@` and `#` were treated the same, but it does only mention the difference for `@`. For reference:
```
Also permitted: IGNORE='c' or IGNORE="c",  where  c  may  be  any
      character  except  space.   IGNORE=# is the default.  That is, in
      the absence of IGNORE option, any record whose first character is
      # is treated as a comment record.

      IGNORE=@  signifies  that  any  data  record having an alphabetic
      character or @ as its first non-blank character (not just in col-
      umn  1) should be ignored.  Alphabetic characters are the letters
      A-Z and a-z.  This permits a table file having header lines to be
      used as an NM-TRAN data set.
```

R/modify-records.R
barrettk: Yeah agree that's probably a good idea

R/modify-records.R
barrettk: Not sure if you'll still hold the same opinion after the latest commit, but can revisit this if you do. Can now do:
```r
# these will be internal functions
nm_data_filter(.mod, data = data) # used by bootstrap function
nm_data_null_map(.mod, data = data) # needs better function name
nm_data_drop_records(.mod, data = data)

# or do all of them, including column renaming - this would be exported, and maybe as an `nm_data()` argument
nm_data_as_nmtran(.mod, data = data)
```

R/modify-records.R
barrettk: I think this behavior is intended if you apply the `@` filter though right? As you affirm, the `NONMEM` documentation does say the `@` applies to all columns

barrettk: @kylebaron I actually just discovered something that appears to be a `NONMEM` bug. I processed the `1.ctl` example using the defaults, and compared it to `FDATA` and was surprised that the first record was filtered out _in addition_ to the header row:

```r
# FDATA - captured using the NM-TRAN PR (not merged yet)
> nmtran_p <- run_nmtran(.mod, delete_on_exit = F)
> fdata <- nm_file_impl(file.path(nmtran_p$run_dir, "FDATA")) %>% stats::setNames(names(nm_data(.mod)))

# R-processed NM-TRAN data
> nmtran_data <- nm_data_as_nmtran(.mod, data = data)


> fdata
# A tibble: 778 × 10
      ID  TIME   MDV  EVID    DV   AMT   SEX    WT   ETN   NUM
   <int> <dbl> <int> <int> <dbl> <int> <int> <dbl> <int> <int>
 1     1  0.25     0     0  12.6     0     1  51.6     1     2
 2     1  0.5      0     0  11.2     0     1  51.6     1     3
 3     1  0.75     0     0  17.7     0     1  51.6     1     4
 4     1  1        0     0  21.9     0     1  51.6     1     5
 5     1  1.25     0     0  16.2     0     1  51.6     1     6
 6     1  1.5      0     0  18.7     0     1  51.6     1     7
 7     1  1.75     0     0  23.8     0     1  51.6     1     8
 8     1  2        0     0  16.5     0     1  51.6     1     9
 9     1  2.25     0     0  17.3     0     1  51.6     1    10
10     1  2.5      0     0  18.4     0     1  51.6     1    11
# ℹ 768 more rows
# ℹ Use `print(n = ...)` to see more rows

> nmtran_data
# A tibble: 779 × 10
      ID  TIME   MDV  EVID    DV   AMT   SEX    WT   ETN   NUM
   <int> <dbl> <int> <int> <dbl> <int> <int> <dbl> <int> <int>
 1     1  0        0     0  1.22     0     1  51.6     1     1
 2     1  0.25     0     0 12.6      0     1  51.6     1     2
 3     1  0.5      0     0 11.2      0     1  51.6     1     3
 4     1  0.75     0     0 17.7      0     1  51.6     1     4
 5     1  1        0     0 21.9      0     1  51.6     1     5
 6     1  1.25     0     0 16.2      0     1  51.6     1     6
 7     1  1.5      0     0 18.7      0     1  51.6     1     7
 8     1  1.75     0     0 23.8      0     1  51.6     1     8
 9     1  2        0     0 16.5      0     1  51.6     1     9
10     1  2.25     0     0 17.3      0     1  51.6     1    10
# ℹ 769 more rows
# ℹ Use `print(n = ...)` to see more rows
```
You'll notice that the `TIME=0` record for `ID==1` was filtered out in `fdata`. This is _only_ true for the first ID too. This feels like a `NONMEM` bug, as I cant imagine filtering out that first record is an expectation.

kylebaron: @barrettk  - I think records and columns refer to rows and columns _in the data file_, not in the realized data frame.  As currently implemented, finding `IGNORE=@` will drop all data rows where any data frame column contains alpha. I posted an example demonstrating this.

kylebaron: > The filters for `inst/models/nonmem/basic/1.ctl` come back as:
> 
> ```
> [1] "!grepl('^[A-Za-z@]', ID) & !grepl('^[A-Za-z@]', TIME) & !grepl('^[A-Za-z@]', MDV) & !grepl('^[A-Za-z@]', EVID) & !grepl('^[A-Za-z@]', DV) & !grepl('^[A-Za-z@]', AMT) & !grepl('^[A-Za-z@]', SEX) & !grepl('^[A-Za-z@]', WT) & !grepl('^[A-Za-z@]', ETN) & !grepl('^[A-Za-z@]', NUM)"
> [2] "ID!=2"     
> ```
> 
> The first filter will drop any row that contains a column with alpha or `@`.
> 
> If you run this on `basic/1`, changing SEX from 0,1 to M,F, it will return no rows.

If you look at that set of filter statements, if you have _any_ data frame column that is character (`[A-Za-z]`) you won't get any of the rows back. This is not what is intended by `IGNORE=@`. The intended behavior is dropping any rows in the file that start with `[A-Za-z]`.

R/nm-tran-data.R
```diff
@@ -0,0 +1,292 @@
...
+#'  filtering the expression corresponds to.
+#' @param data_cols column names associated with the input data. Used for
+#'  `'ignore'` expressions.
+#'
+#' @examples
+#' \dontrun{
+#' test_exprs <- "SEX==1, ID.EQ.2, WT/=70, AGE.NE.30, A=1, WT.GT.40"
+#'
+#' translate_nm_expr(test_exprs, type = 'ignore')
+#' #> [1] "SEX!=1 & ID!=2 & WT==70 & AGE==30 & A!=1 & WT<40"
```
kylebaron, line 69: This looks like it's on the right track; will let you test the important parts much more easily.

R/modify-records.R
kylebaron: The translate filter exprs function (?) is what I was thinking about. Much easier to test.

R/modify-records.R
barrettk: I get NMTRAN errors when I try to reproduce your example though (truncated). The resulting FDATA is 19 records, all for `ID = 1`
```r
 WARNINGS AND ERRORS (IF ANY) FOR PROBLEM    1
             
 (WARNING  2) NM-TRAN INFERS THAT THE DATA ARE POPULATION.

 (DATA ERROR)      RECORD         1, DATA ITEM   7, CONTENTS: F
 ITEM IS NOT A NUMBER.

 (DATA ERROR)      RECORD         2, DATA ITEM   7, CONTENTS: F
 ITEM IS NOT A NUMBER.

 (DATA ERROR)      RECORD         3, DATA ITEM   7, CONTENTS: F
```

```r
> fdata
# A tibble: 18 × 10
      ID  TIME   MDV  EVID    DV   AMT SEX      WT   ETN   NUM
   <int> <dbl> <int> <int> <dbl> <int> <chr> <dbl> <int> <int>
 1     1  0.25     0     0 12.6      0 F      51.6     1     2
 2     1  0.5      0     0 11.2      0 F      51.6     1     3
 3     1  0.75     0     0 17.7      0 F      51.6     1     4
 4     1  1        0     0 21.9      0 F      51.6     1     5
 5     1  1.25     0     0 16.2      0 F      51.6     1     6
 6     1  1.5      0     0 18.7      0 F      51.6     1     7
 7     1  1.75     0     0 23.8      0 F      51.6     1     8
 8     1  2        0     0 16.5      0 F      51.6     1     9
 9     1  2.25     0     0 17.3      0 F      51.6     1    10
10     1  2.5      0     0 18.4      0 F      51.6     1    11
11     1  2.75     0     0 18.3      0 F      51.6     1    12
12     1  3        0     0 18.4      0 F      51.6     1    13
13     1  4        0     0 13.1      0 F      51.6     1    14
14     1  7        0     0 15.1      0 F      51.6     1    15
15     1 10        0     0  8.45     0 F      51.6     1    16
16     1 13        0     0 11.0      0 F      51.6     1    17
17     1 16        0     0  5.74     0 F      51.6     1    18
18     1 19        0     0  8.42     0 F      51.6     1    19
```

Input data for reference
```r
> data
# A tibble: 799 × 10
      ID  TIME   MDV  EVID    DV   AMT SEX      WT   ETN   NUM
   <int> <dbl> <int> <int> <dbl> <int> <chr> <dbl> <int> <int>
 1     1  0        0     0  1.22     0 F      51.6     1     1
 2     1  0.25     0     0 12.6      0 F      51.6     1     2
 3     1  0.5      0     0 11.2      0 F      51.6     1     3
 4     1  0.75     0     0 17.7      0 F      51.6     1     4
 5     1  1        0     0 21.9      0 F      51.6     1     5
 6     1  1.25     0     0 16.2      0 F      51.6     1     6
 7     1  1.5      0     0 18.7      0 F      51.6     1     7
 8     1  1.75     0     0 23.8      0 F      51.6     1     8
 9     1  2        0     0 16.5      0 F      51.6     1     9
10     1  2.25     0     0 17.3      0 F      51.6     1    10
# ℹ 789 more rows
```

kylebaron: @barrettk  - please revisit `inst/model/nonmem/basic/2`; this has been updated to actually run this model (was just filtering last time).

kylebaron: Again, you have to check this on `basic/2`; you won't see anything on `basic/1` because there is no non-numeric data in _the csv file_. 


```
4c4
< $DATA ../../../../extdata/acop.csv IGNORE=@
---
> $DATA ../../../../extdata/acop2.csv IGNORE=@
42,47c42,47
< $TABLE NUM IPRED NPDE CWRES NOPRINT ONEHEADER FILE=1.tab
< $TABLE NUM CL V KA ETAS(1:LAST) NOAPPEND NOPRINT ONEHEADER FILE=1par.tab
< $TABLE ID CL NOAPPEND NOPRINT ONEHEADER FIRSTONLY FILE=1first1.tab ; for testing firstonly
< $TABLE NUM CL NOAPPEND NOPRINT ONEHEADER FIRSTONLY FILE=1first2.tab ; for testing firstonly
< $TABLE NUM ID CL NOAPPEND NOPRINT ONEHEADER FIRSTONLY FILE=1first3.tab ; for testing firstonly
< $TABLE NUM CL KA FAKE NOAPPEND NOPRINT ONEHEADER FILE=1dups.tab ; for testing dropping duplicate columns
---
> $TABLE NUM IPRED NPDE CWRES NOPRINT ONEHEADER FILE=2.tab
> $TABLE NUM CL V KA ETAS(1:LAST) NOAPPEND NOPRINT ONEHEADER FILE=2par.tab
> $TABLE ID CL NOAPPEND NOPRINT ONEHEADER FIRSTONLY FILE=2first1.tab ; for testing firstonly
> $TABLE NUM CL NOAPPEND NOPRINT ONEHEADER FIRSTONLY FILE=2first2.tab ; for testing firstonly
> $TABLE NUM ID CL NOAPPEND NOPRINT ONEHEADER FIRSTONLY FILE=2first3.tab ; for testing firstonly
> $TABLE NUM CL KA FAKE NOAPPEND NOPRINT ONEHEADER FILE=2dups.tab ; for testing dropping duplicate columns
```

barrettk:
Hey @kylebaron I was wondering if you could take a look at `translate_nm_operator` and `translate_nm_expr` after the `@` adjustment and latest commit (added 2 more `NONMEM 7.3` supported operators). Additionally, I was wondering if the prior would have to be revisted to account for the `TIME` paragraph in the section below:


Part of the ` IGNORE=(list)` documentation in the `$DATA` docs:
> With  "=",  "==",  "/=', ".EQ." and ".NE.", the value in the data record and
the value  in  the  list are  compared as character strings. Otherwise, they
are converted to numeric and compared numerically.   (This  is  the  case
with | .NEN.  and .EQN.)  This comparison is made prior to time translation.
**Hence, the TIME item cannot be compared numerically  if  it contains
non-numeric characters such as ":".**

kylebaron:
Hey @barrettk - looking at this this AM. KTB

R/filter-nm-data.R
```diff
@@ -0,0 +1,193 @@
...
+#' @noRd
+get_data_filter_exprs <- function(data_rec, type = c("ignore", "accept")){
+  type <- match.arg(type)
+
+  # Extract & format IGNORE/ACCEPT options
+  opts <- purrr::keep(data_rec$values, function(val){
+    inherits(val, "nmrec_option_value") && identical(val[["name"]], type)
+  })
+
+  exprs <- purrr::map_chr(opts, function(val){
```
kylebaron, line 35: MIght be worth considering splitting by `,`  and unlisting here. Then you're guaranteed to have single expressions the rest of the way. I think you can avoid one `map()` call 
in `translate_nm_expr()`.

kylebaron (commented):
This looks great, @barrettk. Left some comments / suggestions. cc @seth127 .

R/nm-file.R
```diff
@@ -101,6 +103,8 @@ nm_data <- function(.mod) {
   verbose_msg(glue("  rows: {nrow(.d)}"))
   verbose_msg(glue("  cols: {ncol(.d)}"))
   verbose_msg("") # for newline
+
+  if(isTRUE(filter)) .d <- filter_nm_data(.mod, data = .d)
```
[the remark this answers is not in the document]
barrettk: I can add a message about filtering yeah. In terms of the `toupper` aspect; I noticed that too and am unsure. I'd be weary about removing it in this PR (cc @seth127), but I think it should move to below `filter_nm_data()` at the minimum given that that function uses the names as they are specified in the control stream.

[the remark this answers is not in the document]
barrettk: Edit: I take that back. `toupper` must be called ahead of time, at least for our example dataset:

```r
> filter_nm_data(.mod, data = .d)
ignore and/or accept statements could not be converted to filters
• The following errors occurred:
• object 'ID' not found
NULL
```

```r
> .d
# A tibble: 799 × 10
      id  time   mdv  evid    dv   amt   sex    wt   etn   num
   <int> <dbl> <int> <int> <dbl> <int> <int> <dbl> <int> <int>
 1     1  0        0     0  1.22     0     1  51.6     1     1
 2     1  0.25     0     0 12.6      0     1  51.6     1     2
 3     1  0.5      0     0 11.2      0     1  51.6     1     3
 4     1  0.75     0     0 17.7      0     1  51.6     1     4
 5     1  1        0     0 21.9      0     1  51.6     1     5
 6     1  1.25     0     0 16.2      0     1  51.6     1     6
 7     1  1.5      0     0 18.7      0     1  51.6     1     7
 8     1  1.75     0     0 23.8      0     1  51.6     1     8
 9     1  2        0     0 16.5      0     1  51.6     1     9
10     1  2.25     0     0 17.3      0     1  51.6     1    10
# ℹ 789 more rows
# ℹ Use `print(n = ...)` to see more rows
```

R/filter-nm-data.R
```diff
@@ -0,0 +1,193 @@
...
+      if(expr == "#"){
+        # IGNORE=# is the default.  That is, in the absence of IGNORE option, any
+        # record whose first character is '#' is treated as a comment record.
+        paste0("!grepl('^#', ", data_cols[1], ")")
+      }else if(expr == "@"){
+        # IGNORE=@ signifies that any data record having an alphabetic character
+        # or `@` as its first non-blank character in column one should be ignored.
+        #  - This permits a table file having header lines to be used as an
+        #    NM-TRAN data set.
+        paste0("!grepl('^[A-Za-z@]', ", data_cols[1], ")")
```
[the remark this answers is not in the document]
barrettk: I thought about it since you had that, but didnt see anything in the `NONMEM` docs that would indicate that behavior. It specifically says:
 > IGNORE=@  signifies  that  any  data  record having an alphabetic character or @ as its first non-blank character (not just in column  1) should be ignored.

R/filter-nm-data.R
```diff
@@ -0,0 +1,193 @@
...
+      }else if(grepl('^[a-zA-Z0-9]{1,}$', expr)){
```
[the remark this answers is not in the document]
barrettk: I dont think so, _I think_ this can be any number of characters as long as it's _at least 1_:
> IGNORE=c1 [Note: The following two options, IGNORE  and  ACCEPT  allow  the user  to drop records from the NM-TRAN data set prior to the run. With NONMEM 7.5, it is also possible to  drop  records  from  the NONMEM  data  set  using abbreviated code, which is more flexible than the criteria that can be specified in the $DATA record. See PRED_IGNORE_DATA, below.] 

> Specifies that any data record having the character c1 in  column 1  should be ignored, i.e., these records are not included in the NONMEM data set.  This allows comment records to be  included  in the  NM-TRAN  data set.  In general, records having the character c1 in column 1 will be called "comment records". Also permitted: IGNORE='c' or IGNORE="c",  where  c  may  be  any character  except  space.   IGNORE=# is the default.  That is, in the absence of IGNORE option, any record whose first character is # is treated as a comment record.

Will do some testing to confirm this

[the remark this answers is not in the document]
kylebaron: It looks like it should be a single character: 

```
^[a-zA-Z]$
```

R/nm-file.R
```diff
@@ -87,10 +87,12 @@ nm_par_tab <- function(.mod) {
 
 #' @describeIn nm_file Reads the input data file from a `bbi_nonmem_model` or
 #'   `bbi_nonmem_summary` object
+#' @param filter Logical (`T`/`F`). If `TRUE`, filter data based on `IGNORE` and
```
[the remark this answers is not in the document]
barrettk: I opted to adjust the `.rename` documentation, since we have `"Logical ('T'/'F')"` in a number of other places too. Let me know if you disagree with this move.

R/filter-nm-data.R
```diff
@@ -0,0 +1,193 @@
...
+  expr <- gsub("(?<=[^=!<>])=(?=[^=!<>])", "==", expr, perl = TRUE)
+  return(expr)
+}
+
+#' Function to invert ignore expressions
+#' @noRd
+#' @keywords internal
+invert_operator <- function(expr) {
+  expr <- dplyr::case_when(
+    grepl("==", expr) ~ gsub("==", "!=", expr),
```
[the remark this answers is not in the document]
barrettk: I added it to the location you mentioned for the `gsub` expression. It broke things when I attempted to add it to `translate_nm_operator`, but let me know if you were thinking of including this in the `grepl` expression instead

R/filter-nm-data.R
```diff
@@ -0,0 +1,193 @@
...
+  # Extract & format IGNORE/ACCEPT options into expressions
+  ignore_exprs <- get_data_filter_exprs(data_rec, type = "ignore")
+  accept_exprs <- get_data_filter_exprs(data_rec, type = "accept")
+
+  # Translate NONMEM syntax into `dplyr::filter` logic
+  ignore_filters <- purrr::map_chr(ignore_exprs, translate_nm_expr, data_cols = names(data))
+  accept_filters <- purrr::map_chr(accept_exprs, translate_nm_expr, type = "accept")
+
+  # Combine all filter expressions
+  all_filters <- c(ignore_filters, accept_filters)
```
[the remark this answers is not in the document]
barrettk: Did my own testing and confirmed this to be the case. This wasn't obvious from the docs, so thanks for pointing it out! Added a related test for this too.

R/filter-nm-data.R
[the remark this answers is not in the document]
barrettk: Sounds good. Yeah the digits definitely shouldnt have been in there either

R/filter-nm-data.R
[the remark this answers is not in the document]
kylebaron: Ok to just leave it. But look at the `pattern` there .. it's just looking for a literal string. So you can use `fixed = TRUE` in that case can be faster and safer. 

In `translate_nm_operator()`, there's all sorts of back ticks in there, but it's not needed in the last 4. You do need them in the first two, but you could render all this using `fixed = TRUE` and it would run the same (even faster). 

```r
translate_nm_operator <- function(expr) {
  # EQN and NEN are available after NONMEM 7.3
  # expr <- gsub("\\.EQ\\.|\\.EQN\\.", "==", expr) # regular expression used here, but you don't have to 
  # expr <- gsub("\\.NE\\.|\\.NEN\\.|/=", "!=", expr) # regular expression used here, but not required
  expr <- gsub("\\.LT\\.", "<", expr)
  expr <- gsub("\\.LE\\.", "<=", expr)
  expr <- gsub("\\.GT\\.", ">", expr)
  expr <- gsub("\\.GE\\.", ">=", expr)
```

R/filter-nm-data.R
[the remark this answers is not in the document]
kylebaron: I was suggesting it because of what you shared in the docs ... first _non-blank_ character.  If `data_cols[1]` had a row that was ` C`, it should be ignored, right?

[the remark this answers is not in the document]
barrettk: Yeah that sounds right; just made the adjustment

R/bootstrap-model.R
```diff
@@ -173,26 +165,27 @@ setup_bootstrap_run <- function(
...
+      # NULL if IGNORE/ACCEPT expressions cant be turned into dplyr expressions
+      if(is.null(starting_data)){
         rlang::inform(
           paste(
             "Defaulting to input data, which may include data that doesn't enter",
             "the final problem (i.e. ignored subjects)"
           )
         )
         starting_data <- nm_data(orig_mod) %>% suppressMessages()
+      }else{
```
kylebaron, line 179: If you got this far, it seems like everything is working as planned by default, no? I wouldn't send a message here.

R/filter-nm-data.R
```diff
@@ -0,0 +1,221 @@
...
+  expr <- gsub("(?<=[^=!<>])=(?=[^=!<>])", "==", expr, perl = TRUE)
+  return(expr)
+}
+
+#' Function to invert ignore expressions
+#' @noRd
+#' @keywords internal
+invert_operator <- function(expr) {
+  expr <- dplyr::case_when(
+    grepl("==", expr) ~ gsub("==", "!=", expr, fixed = TRUE),
```
kylebaron, line 93:
```suggestion
    grepl("==", expr, fixed = TRUE) ~ gsub("==", "!=", expr, fixed = TRUE),
```

kylebaron: If you do it on one side, do it on both.

R/filter-nm-data.R
```diff
@@ -0,0 +1,221 @@
...
+  # Create the final dplyr::filter expression
+  filter_expression <- paste(r_filters, collapse = " & ")
+
+  # Apply filters
+  filtered_data <- tryCatch({
+    data %>% dplyr::filter(eval(parse(text = filter_expression)))
+  }, error = function(cond){
+    rlang::inform(
+      c(
+        "ignore and/or accept statements could not be converted to filters",
```
kylebaron, line 208:
```suggestion
        "ignore/accept list could not be converted to filters",
```

kylebaron: Just for consistency?

R/bootstrap-model.R
```diff
@@ -173,26 +165,27 @@ setup_bootstrap_run <- function(
...
+        n_recs_dropped <- attributes(starting_data)$n_records_dropped
+        if(n_recs_dropped > 0){
+          rlang::inform(
+            c(
+              glue("Parsed IGNORE/ACCEPT options in {basename(get_model_path(.boot_run))}"),
+              glue("Dropped {n_recs_dropped} records from starting data")
```
kylebaron, line 185: I don't think this message is needed; but if printing the status to the console, it's probably more informative to say the number of records retained; more likely that the user will be able to connect with that number; we don't usually track the number dropped.

R/nm-file.R
```diff
@@ -87,20 +88,31 @@ nm_par_tab <- function(.mod) {
...
   .d <- fread(.path, na.strings = ".", verbose = FALSE)
   .d <- remove_dup_cols(.d)
   .d <- as_tibble(.d)
+
   names(.d) <- toupper(names(.d))
+
+  if(isTRUE(filter)){
+    .d <- filter_nm_data(.mod, data = .d)
+    recs_dropped <- attributes(.d)$n_records_dropped
+    verbose_msg(paste("Number of records dropped:", cli::col_blue(recs_dropped)))
```
kylebaron, line 109: Probably saying the number retained will be more helpful here, if anything is printed at all.

R/filter-nm-data.R
```diff
@@ -0,0 +1,221 @@
...
+}
+
+#' Function to translate NONMEM operators to R operators
+#' @noRd
+#' @keywords internal
+translate_nm_operator <- function(expr) {
+  # EQN and NEN are available after NONMEM 7.3
+  expr <- gsub("\\.EQ\\.|\\.EQN\\.", "==", expr)
+  expr <- gsub("\\.NE\\.|\\.NEN\\.|/=", "!=", expr)
+  expr <- gsub("\\.LT\\.", "<", expr)
```
kylebaron, line 78: More literal strings; you can can leave them as is, but no need to do matching here. Just commenting for posterity. You can convert the first two (lines 76 and 77) to 5 literal `gsubs` and it will read clearer and run faster (just a touch faster, but it'll be faster).

R/filter-nm-data.R
```diff
@@ -0,0 +1,221 @@
...
+  return(expr)
+}
+
+#' Translate `NONMEM` `IGNORE` and `ACCEPT` expressions into [dplyr::filter()]
+#'  expressions.
+#'
+#' @param nm_expr a `NONMEM` filter expression. e.g., `'ID.EQ.2, BLQ=1'`.
+#' @param type Either `'ignore'` or `'accept'`. Denotes which type of `NONMEM`
+#'  filtering the expression corresponds to.
+#' @param data_cols column names associated with the input data. Used for
```
kylebaron, line 110: Do you start with capital letter in your arg documentation?

R/bootstrap-model.R
```diff
@@ -86,15 +86,8 @@ new_bootstrap_run <- function(
 #' @param strat_cols columns to maintain proportion for stratification
 #' @param seed a seed for sampling the data. Set to `NULL` to avoid setting.
 #' @param data A dataset to resample from. Defaults to `NULL`, which will use
```
kylebaron, line 88: Maybe check the help topic for consistency in capitalization for first letter in arg description.

kylebaron (commented):
Noticed at the end of the review some inconsistency in capitalization for the first letter of your function argument descriptions; might be worth a look at all the functions here to get it consistent.

R/bootstrap-model.R
barrettk: I would think knowing the proportion that was dropped (or proportion that remained) would be useful? Like knowing "I lose ~ 15% of the records" type of thing?

R/bootstrap-model.R
```diff
@@ -173,26 +165,27 @@ setup_bootstrap_run <- function(
...
-        # select only columns from original data set
-        starting_data <- starting_data %>%
-          dplyr::select(attr(starting_data, "nm_join_origin")$data)
-
-        if ("DV.DATA" %in% names(starting_data)) {
-          starting_data <- dplyr::rename(starting_data, "DV" = "DV.DATA")
-        }
-      }else{
+      # NULL if IGNORE/ACCEPT expressions cant be turned into dplyr expressions
+      if(is.null(starting_data)){
```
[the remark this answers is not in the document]
barrettk: The purpose of this PR was to move away from `nm_join` as the default via parsing the data record. I dont think we want to start requiring a dataset since this was not required in the previous release. That could be a change down the line, but I would consider that out of scope for this and the parent PR.

R/bootstrap-model.R
barrettk: Yeah this has been a `bbr` issue for a while now. We seem to go back and forth across a lot of functions and dont really seem to have a standard. Ill update the bootstrap ones in this PR, but there are a couple inherited from other functions (like `delete_models` in `cleanup_bootstrap_run`) that I feel like would be out of scope for this PR

R/bootstrap-model.R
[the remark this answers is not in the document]
barrettk: Ahh I misinterpreted that - im good with that change. I just implemented your suggestion in the latest commit.

R/filter-nm-data.R
```diff
@@ -0,0 +1,249 @@
...
+#' }
+#' @keywords internal
+#' @seealso [translate_nm_operator()], [translate_nm_expr()]
+invert_operator <- function(expr) {
+  expr <- dplyr::case_when(
+    grepl("==", expr, fixed = TRUE) ~ gsub("==", "!=", expr, fixed = TRUE),
+    grepl("!=", expr, fixed = TRUE) ~ gsub("!=", "==", expr, fixed = TRUE),
+    grepl("<=", expr, fixed = TRUE) ~ gsub("<=", ">=", expr, fixed = TRUE),
+    grepl(">=", expr, fixed = TRUE) ~ gsub(">=", "<=", expr, fixed = TRUE),
+    grepl("<", expr, fixed = TRUE) ~ gsub("<", ">", expr, fixed = TRUE),
```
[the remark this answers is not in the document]
barrettk: Could you add a bit of clarity here? I confirmed that `<=` gets correctly inverted:

```r
> invert_operator(c('A==2', 'B <= 4', 'C >= 2', 'D > 1', 'E < 5'))
[1] "A!=2"   "B >= 4" "C <= 2" "D < 1"  "E > 5"
```

R/filter-nm-data.R
```diff
@@ -0,0 +1,249 @@
...
+#'
+#' @examples
+#' \dontrun{
+#'
+#' translate_nm_operator(c("AGE.NE.30", "ID.EQ.2", "WT/=70"))
+#' #> [1] "AGE!=30" "ID==2" "WT!=70"
+#' }
+#' @keywords internal
+#' @seealso [invert_operator()], [translate_nm_expr()]
+translate_nm_operator <- function(expr) {
```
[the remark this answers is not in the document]
barrettk: The `dplyr` filter expression should still fail but I agree a more helpful error message would be nice

R/bootstrap-model.R
```diff
@@ -173,20 +165,11 @@ setup_bootstrap_run <- function(
...
-        # select only columns from original data set
-        starting_data <- starting_data %>%
-          dplyr::select(attr(starting_data, "nm_join_origin")$data)
-
-        if ("DV.DATA" %in% names(starting_data)) {
-          starting_data <- dplyr::rename(starting_data, "DV" = "DV.DATA")
-        }
-      }else{
-        rlang::inform(
+      starting_data <- nm_data(orig_mod, filter = TRUE) %>% suppressMessages()
```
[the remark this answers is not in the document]
barrettk: One point and two questions, though id like @seth127's opinion here too:

 point: 
 - It's my understanding that we would prefer to keep this away from `nm_data()`. I think `setup_bootstrap_run` would have to be the landing spot for this, but this is what i'd like Seth's opinion on.

questions:
 - I started doing this, but we'd need to add a `.bbi_args` argument like we do for `nm_join()` if we were to do that to avoid `model_summary()` issues. Does doing this seem out of place for `setup_bootstrap_run`?
 - Im unsure if this should be a `dev_error()` versus a regular abort call or even a warning. Are we adamant that if there's a mismatch, the filtering _definitely_ went wrong, or is it possible a mismatch could arise for a different reason?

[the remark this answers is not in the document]
barrettk: @kylebaron Im proposing a message like the one below, though let me know if it's too verbose.

 Im **assuming we'll want to error out here**, but wanted to pose the option of prompting the user if this is ok. My understanding is that diverging from the number of expected records would mean the bootstrap run is invalid. However since we allow users to manually adjust the bootstrap template control stream (after `new_bootstrap_run` is called), I was wondering if there could ever be an intentional deviation from the model the bootstrap is based on? If that's a possibility, I would propose something like this:
image

kylebaron (approved):

PR #712, opened 2024-07-08
reproducible example that IGNORE=@ isn't working as intented

kylebaron:
@barrettk Here's a reproducible example where `IGNORE=@` doesn't work in the existing implementation. 

I dummied up run 2 and added a character column at the end of the `acop.csv` data set. The result is that no rows are retained in the output, as expected with the implementation as-is. 

I'll pick up on the conversation back on the `parse-data-record` PR. 

https://github.com/metrumresearchgroup/bbr/blob/5ccbc80d010e73dd7a27fb9a22f80b72daf0ce42/test.R

Update: Tuesday, July 9 2024
https://github.com/metrumresearchgroup/bbr/pull/712/files#diff-fe7ee37e07a69e9f8db8aef5ebfe8ef28bf5b7e2617c51d832107c04408c4ec7


kylebaron:
@barrettk - previous commit didn't actually run the model; just filtered using new implementation. This was updated just not to run the model. Please give it a try and let me know what you find.

seth127:
@barrettk can you look at this and make sure that this has been fixed as of `bbr 1.12.0`? Thanks.

barrettk:
@seth127 confirmed this is working using the reproducible example from this PR - will close

PR #719, opened 2024-09-27
ci: update to latest bbi

kyleam:
Routine bump to use recent bbi release.

(Eventually I'd like to update the [setup-bbi action](https://github.com/metrumresearchgroup/actions/tree/main/setup-bbi) to make it default to the latest version.)


barrettk (approved):
Took a look at the latest bbi release too. Looks good

PR #720, opened 2024-09-30
Add `size_by` argument to `model_tree()`

barrettk:
### Core new feature: `size_by`
 - Allows users to control the node sizing by a particular column present in `.log_df`

 - Important to note that `ofv` cant be used without first calling `add_summary()`, which is consistent with the previous behavior of `color_by` and `include_info`.

 - This argument *does* support logical columns, however I noticed that the normalized sizing can differ here depending when the first `TRUE` value originates. If it's the first value, all the nodes appear smaller. Anywhere else and the "base size" is larger. This observation is likely true for numeric columns too, though is less apparent when the values are all closer to each other.
    -  We may want to revisit NA values for numeric columns

image

### Other changes

 - Fixed a bug when coloring numeric columns. See details in [6d11e25](https://github.com/metrumresearchgroup/bbr/pull/720/commits/6d11e25bb779282ff79c040037bf413790384b21)
 - Added `font_size` argument. While updating the vignette I foresaw a future request to adjust this, and it was easy enough to pass to `collapsibleTree::collapsibleTreeNetwork` (with one other change)
    - Was later requested by @callistosp 
 - Updated vignette with new example
 - Formatting fix: dont display tooltips for 'start' node ([fc65532](https://github.com/metrumresearchgroup/bbr/pull/720/commits/fc65532169cb5c31a13e11c291cd8e0f63bb0bb5))
 - Css simplifications and other minor adjustments

### DIFF
--- a/R/model-tree.R
+++ b/R/model-tree.R
@@ -173,7 +173,7 @@ model_tree.bbi_log_df <- function(
 
   # Format coloring
   tree_data <- color_tree_by(tree_data, color_by = color_by)
-  tree_attr <- ifelse(is.null(color_by), "leafCount", color_by)
+  tree_attr <- attributes(tree_data)$color_by
 
   # Format sizing
   tree_data <- size_tree_by(tree_data, size_by = size_by)
@@ -721,11 +721,15 @@ color_tree_by <- function(tree_data, color_by = "run"){
       tree_data$col <- factor(tree_data$col)
       levels(tree_data$col) <- c(node_colors, pal_bbr)
     }
+    # Set color_by attribute to color_by argument
+    attr(tree_data, "color_by") <- color_by
   }else{
     # all bbr red color
     pal_bbr <- scales::pal_gradient_n(bbr_cols)(1)
     tree_data$col <- factor(tree_data$col)
     levels(tree_data$col) <- c(node_colors, pal_bbr)
+    # Set color_by attribute to leafCount (default)
+    attr(tree_data, "color_by") <- "leafCount"
   }
 
   return(tree_data)
@@ -745,6 +749,7 @@ size_tree_by <- function(tree_data, size_by = NULL){
       # Set node sizes with NA values (including start node) to mean value
       mean_val <- mean(tree_data$node_size, na.rm = TRUE)
       tree_data$node_size[is.na(tree_data$node_size)] <- mean_val
+      # Set size_by attribute to node_size column
       attr(tree_data, "size_by") <- "node_size"
     }else{
       col_class <- class(tree_data[[size_by]])
@@ -757,6 +762,7 @@ size_tree_by <- function(tree_data, size_by = NULL){
       attr(tree_data, "size_by") <- "leafCount"
     }
   }else{
+    # Set size_by attribute to leafCount (default)
     attr(tree_data, "size_by") <- "leafCount"
   }
 
--- a/R/model-tree.R
+++ b/R/model-tree.R
@@ -691,7 +691,7 @@ color_tree_by <- function(tree_data, color_by = "run"){
     n_levels <- dplyr::n_distinct(vals)
 
     # To preview color palette: scales::show_col(pal_bbr)
-    if(inherits(vals, "numeric")){
+    if(inherits(vals, c("numeric", "integer"))){
       # Gradient coloring (sorted): get colors for unique values (excluding NA)
       sorted_vals <- sort(unique(vals))
       pal_bbr <- scales::pal_gradient_n(bbr_cols)(seq(0, 1, length.out = n_levels))
--- a/R/model-tree.R
+++ b/R/model-tree.R
@@ -739,16 +739,41 @@ color_tree_by <- function(tree_data, color_by = "run"){
 #' another column.
 #' @inheritParams make_tree_tooltip
 #' @inheritParams model_tree
+#' @param rescale_to A numeric vector of length 2 specifying the range to rescale
+#'  `size_by` values to. Defaults to `c(1, 3)`, where `1` is the smallest node
+#'  size and `3` is the largest.
 #' @noRd
-size_tree_by <- function(tree_data, size_by = NULL){
+size_tree_by <- function(tree_data, size_by = NULL, rescale_to = c(1, 3)){
   if(!is.null(size_by)){
     checkmate::assert_true(size_by %in% names(tree_data))
+    checkmate::assert_numeric(rescale_to, len = 2, lower = 1)
+
     tree_data$node_size <- tree_data[[size_by]]
+
     # Scale size with numeric value
     if(inherits(tree_data$node_size, c("numeric", "integer"))){
-      # Set node sizes with NA values (including start node) to mean value
+      # Rescale values based on standard deviation
+      # - a large SD can lead to very large nodes
       mean_val <- mean(tree_data$node_size, na.rm = TRUE)
-      tree_data$node_size[is.na(tree_data$node_size)] <- mean_val
+      sd_val <- sd(tree_data$node_size, na.rm = TRUE)
+
+      if(sd_val != 0){
+        tree_data$node_size <- (tree_data$node_size - mean_val) / sd_val
+      }else{
+        # Would only happen if all values are the same
+        # - avoids dividing by 0
+        tree_data$node_size <- rep(1, length(tree_data$node_size))
+      }
+
+      # Rescale to specified range
+      # Note: node sizes cannot be negative
+      tree_data <- tree_data %>% dplyr::mutate(
+        node_size = scales::rescale(.data$node_size, to = rescale_to)
+      )
+
+      # Set node sizes with NA values (including start node) to mean rescale_to
+      tree_data$node_size[is.na(tree_data$node_size)] <- mean(rescale_to)
+
       # Set size_by attribute to node_size column
       attr(tree_data, "size_by") <- "node_size"
     }else{
--- a/R/model-tree.R
+++ b/R/model-tree.R
@@ -648,6 +648,7 @@ make_tree_tooltip <- function(tree_data, digits = 3, font_size = 10){
       })
       paste0(other_html, collapse = "")
     })
+    other_tooltip[1] <- "" # Skip start node for other tooltips
     tooltip <- paste0(tooltip, other_tooltip, sum_tooltip)
   }else{
     tooltip <- paste0(tooltip, sum_tooltip)
--- a/R/model-tree.R
+++ b/R/model-tree.R
@@ -680,7 +680,19 @@ color_tree_by <- function(tree_data, color_by = "run"){
     n_levels <- dplyr::n_distinct(vals)
 
     # To preview color palette: scales::show_col(pal_bbr)
-    if(inherits(vals, "logical")){
+    if(inherits(vals, "numeric")){
+      # Gradient coloring (sorted): get colors for unique values (excluding NA)
+      sorted_vals <- sort(unique(vals))
+      pal_bbr <- scales::pal_gradient_n(bbr_cols)(seq(0, 1, length.out = n_levels))
+
+      # Assign colors based on sorted values
+      color_mapping <- setNames(pal_bbr, sorted_vals)
+      tree_data$col <- ifelse(
+        tree_data$col %in% sorted_vals,
+        color_mapping[as.character(tree_data$col)],
+        tree_data$col
+      )
+    }else if(inherits(vals, "logical")){
       # Ensure both TRUE and FALSE colors are extracted, even if only one occurs
       pal_bbr <- scales::pal_gradient_n(bbr_cols)(c(0,1))
       # Explicitly set FALSE to white and TRUE to red
@@ -693,7 +705,7 @@ color_tree_by <- function(tree_data, color_by = "run"){
         )
       )
     }else{
-      # Gradient coloring: get colors for unique values (excluding NA)
+      # Gradient coloring; doesn't need to be sorted
       pal_bbr <- scales::pal_gradient_n(bbr_cols)(seq(0, 1, length.out = n_levels))
       tree_data$col <- factor(tree_data$col)
       levels(tree_data$col) <- c(node_colors, pal_bbr)
--- a/R/model-tree.R
+++ b/R/model-tree.R
@@ -21,6 +21,7 @@
 #'        be saved as a PNG and loaded into the viewer.
 #' @param width Width in pixels (optional, defaults to automatic sizing)
 #' @param height Height in pixels (optional, defaults to automatic sizing)
+#' @param font_size Font size of the label text in pixels
 #' @param ... Additional arguments passed to [run_log()]. Only used if `.log_df`
 #'        is a modeling directory.
 #'
@@ -121,6 +122,7 @@ model_tree <- function(
     static = FALSE,
     width = NULL,
     height = NULL,
+    font_size = 10,
     ...
 ){
   UseMethod("model_tree")
@@ -139,6 +141,7 @@ model_tree.character <- function(
     static = FALSE,
     width = NULL,
     height = NULL,
+    font_size = 10,
     ...
 ){
   checkmate::assert_directory_exists(.log_df)
@@ -149,7 +152,8 @@ model_tree.character <- function(
     color_by = color_by, size_by = size_by,
     add_summary = add_summary, digits = digits,
     zoomable = zoomable, static = static,
-    width = width, height = height
+    width = width, height = height,
+    font_size = font_size
   )
 }
 
@@ -165,6 +169,7 @@ model_tree.bbi_log_df <- function(
     static = FALSE,
     width = NULL,
     height = NULL,
+    font_size = 10,
     ...
 ){
   # Make sure required dependencies are installed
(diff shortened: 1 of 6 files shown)


R/model-tree.R
```diff
@@ -1,12 +1,15 @@
...
 #'        a base directory to look in for models. See details for more options.
-#' @param include_info vector of columns present in `.log_df` to include in the
+#' @param include_info A vector of columns present in `.log_df` to include in the
 #'        tooltip.
-#' @param color_by a run log column to color the nodes by. Can be helpful for
+#' @param color_by A run log column to color the nodes by. Can be helpful for
 #'        identifying which models are starred, have heuristics, etc. See details
 #'        for more information.
+#' @param size_by A **numeric** (or integer) run log column to size the nodes by.
+#'        If not specified, the default sizing is `'leafCount'`, which sizes the
```
kyleam, line 11: nit-pick and okay to leave as is: I don't think adding the specific name of the collapsibleTree value here ("leaftCount") adds anything on top of the description of the behavior.

R/model-tree.R
```diff
@@ -50,8 +54,8 @@
 #'     - Note that the above summary columns will only receive the special
 #'     formatting if added via `add_summary = TRUE`.
 #'     - i.e. if `.log_df = run_log() %>% add_summary()` and
-#'     `include_info = 'ofv'`, The `'OFV'` parameter will be formatted as any
-#'     other additional column.
+#'     `include_info = 'ofv'`, The `'OFV'` parameter will display the same as if
```
kyleam, line 57: `, The` -> `, the` ?

R/model-tree.R
```diff
@@ -95,74 +108,90 @@ model_tree <- function(
...
+    font_size = 10,
     ...
 ){
   # Make sure required dependencies are installed
   stop_if_tree_missing_deps(static = static)
 
   # Make tree data
-  tree_data <- make_tree_data(.log_df, include_info, color_by, add_summary)
+  # TODO: decide if we want to append color_by and size_by to tooltip automatically
+  # include_info <- unique(c(include_info, color_by, size_by))
```
kyleam, line 173: Decide on this and remove this comment before merging?

(I'll defer to your opinion on which way to go. My leaning would be to _not_ add these automatically given that `color_by` already existed before this PR and wasn't added to the tooltip automatically.)

R/model-tree.R
```diff
@@ -95,74 +108,90 @@ model_tree <- function(
...
+  tree_data <- make_tree_data(.log_df, include_info, color_by, size_by, add_summary)
 
   # Format coloring
   tree_data <- color_tree_by(tree_data, color_by = color_by)
+  tree_attr <- ifelse(is.null(color_by), "leafCount", color_by)
+
+  # Format sizing
+  tree_data <- size_tree_by(tree_data, size_by = size_by)
```
kyleam, line 181: `ifelse -> if/else` because the condition is always a scalar. Same for line 178 above (although I know that one is not new to this PR).

R/model-tree.R
```diff
@@ -95,74 +108,90 @@ model_tree <- function(
...
 
   # Create model tree
-  tree_attr <- ifelse(is.null(color_by), "leafCount", color_by)
   pl_tree <- collapsibleTree::collapsibleTreeNetwork(
-    tree_data, zoomable = zoomable, attribute = tree_attr,
-    fill="col", collapsed = FALSE, nodeSize = "leafCount",
-    tooltipHtml = "tooltip", width = width, height = height
+    tree_data, zoomable = zoomable, collapsed = FALSE,
+    # Coloring and sizing
+    attribute = tree_attr, fill="col", nodeSize = node_size, aggFun = identity,
```
kyleam, line 191: Based on dropping this `aggfun` and testing, I think setting this to `identity` (or at least something other than the default value of `sum`) is necessary to get the desired sizing, but it's not obvious to me what's going on here. An aggregation function is a `[]T -> T` function, reducing many values to one (which aligns with the `sum` default value of `collapsibleTree`). How does `identity` (`[]T -> []T`) behave here given that it doesn't reduce?

I'm asking to

1.  get a better understanding of what's going on here

2.  make sure we're not relying on undefined behavior

3.  get a code comment added, since I think the behavior here is non-obvious

R/model-tree.R
```diff
@@ -681,6 +733,37 @@ color_tree_by <- function(tree_data, color_by = "run"){
...
+      tree_data$node_size[is.na(tree_data$node_size)] <- mean_val
+    }else{
+      col_class <- class(tree_data[[size_by]])
+      cli::cli_warn(
+        c(
+          "Only numeric columns are supported. Column {.val {size_by}} is {.val {col_class}}",
+          "i" = "Setting node size to constant"
+        )
+      )
+      tree_data$node_size <- 1
```
kyleam, line 758: Think it's fine to leave, but what's your thinking behind setting it to a constant size rather than warning and leaving the default "leafCount" behavior?

R/model-tree.R
```diff
@@ -681,6 +733,37 @@ color_tree_by <- function(tree_data, color_by = "run"){
...
+    }
+  }else{
+    tree_data$node_size <- 1
+  }
```
kyleam, line 762: Think it's fine to leave, but what's your thinking behind setting it to a constant size rather than warning and leaving the default "leafCount" behavior?

kyleam (commented):
I haven't finished reviewing these, but here's an initial wave

Testing this out, things are overall working as expected on my end. I noticed two behaviors that look off to me. Here's an example starting with `log_df` from the [`size_tree_by` tests](https://github.com/metrumresearchgroup/bbr/blob/48ca4e00e6bb2141f4bd84635f5c8c710c317faa/tests/testthat/test-model-tree.R#L387-L392).

```R
log_df$foo <- 1:5
log_df$bar <- 5:1
log_df$baz <- -2:2
#> Warning messages:
#> 1: In sqrt(x$SizeOfNode * scaleFactor) : NaNs produced
#> 2: In sqrt(x$SizeOfNode * scaleFactor) : NaNs produced
```

**First**: feeding negative values to `size_by` throws things off. There are warning messages and the relative sizing does not match the values.

```R
model_tree(log_df, add_summary = FALSE, size_by = "baz", static = TRUE)
```

baz

Perhaps `model_tree` should scale the input? (Only the relative values should matter for sizing.)

**Second**: I was surprised about how nodes were sized differently if values were reversed.

```R
model_tree(log_df, add_summary = FALSE, size_by = "foo", static = TRUE)
```

```R
model_tree(log_df, add_summary = FALSE, size_by = "bar", static = TRUE)
```

foo
bar

I expected the same set of sizes to be used for those nodes, just in reverse (e.g., run 1 in bar would be the same size as run 5 in foo). Is the above behavior intended?

---

> is only intended to support numeric columns for sizing, so I opted to map these values to: TRUE=5, FALSE=1, NA=3

Should this part of the PR description have been updated for f6e65542 (Dont support logical columns, improve error formatting)?

R/model-tree.R
kyleam: (Sorry, I put in the wrong comment; I've replaced the text.)

R/model-tree.R
barrettk: This reasoning makes sense. If anything this can be a feature request at a later date but will just remove for now

R/model-tree.R
barrettk: My understanding was that `if/else` is mainly necessary when the resulting objects can differ based on the condition. I would have thought `ifelse` was appropriate here. Can you explain a bit more?

kyleam: (Sorry for misplacing my comment; it should have been on line 182)

`if`/`else` is the more basic form.  For a scalar condition, I see no reason to use the higher-level `ifelse`.   

Reasons I see to avoid:
 * `ifelse` suggests "vectorized" (at least to this one reader, but I don't think it's just me), so it requires the reader to analyze more closely to realize it's not
 * using `ifelse` for scalars can lead to unintentionally and silently dropping parts of a non-scalar result (example: https://github.com/metrumresearchgroup/mrgvalidate/commit/c42777c67007d4dc5b03e431045e7266b4138f2a)

R/model-tree.R
```diff
@@ -653,7 +691,19 @@ color_tree_by <- function(tree_data, color_by = "run"){
     n_levels <- dplyr::n_distinct(vals)
 
     # To preview color palette: scales::show_col(pal_bbr)
-    if(inherits(vals, "logical")){
+    if(inherits(vals, "numeric")){
```
kyleam, line 694: Did you intend for integers to go down this path as well?  As is, integers go down the `else` arm and get treated as factors. 

Here's the [color_by regression case](https://github.com/metrumresearchgroup/bbr/blob/48ca4e00e6bb2141f4bd84635f5c8c710c317faa/tests/testthat/test-model-tree.R#L375-L380) data frame, changed to use integers:

```
> log_df
# A tibble: 5 × 11
  absolute_model_path             run   yaml_md5 model_type description bbi_args     based_on tags   notes  star  numeric_vals
  <chr>                           <chr> <chr>    <chr>      <chr>       <list>       <list>   <list> <list> <lgl>        <int>
1 /data/home/kylem/src/github/me… 1     6ccf206… nonmem     original a… <named list> <NULL>   <chr>  <NULL> FALSE         1534
2 /data/home/kylem/src/github/me… 2     b5f8010… nonmem     NA          <named list> <chr>    <chr>  <NULL> TRUE          3892
3 /data/home/kylem/src/github/me… 3     99d902b… nonmem     NA          <named list> <chr>    <chr>  <NULL> FALSE          731
4 /data/home/kylem/src/github/me… 4     5161777… nonmem     NA          <named list> <chr>    <NULL> <NULL> FALSE         2653
5 /data/home/kylem/src/github/me… 5     690952d… nonmem     NA          <named list> <chr>    <NULL> <NULL> FALSE         3574
```


ints

tests/testthat/test-model-tree.R
```diff
@@ -346,6 +370,84 @@ withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
...
+
+      # Regression test: numeric color_by is sorted appropriately
+      log_df <- run_log(MODEL_DIR)
+      log_df$numeric_vals <- c(1534, 3892, 731, 2653, 3574)
+      pl_tree <- model_tree(
+        log_df, add_summary = FALSE, color_by = "numeric_vals",
+        include_info = "numeric_vals"
+      )
+      node_colors <- get_node_attribute(pl_tree$x$data$children, attr = "fill")
+      expected_colors <- c("#F4DBD3", "#EB003D", "#FFFFFF", "#ED9D84", "#E35B44")
```
kyleam, line 382: I don't love embedding these values because, if I understand correctly, the values are calculated by another package (scales) and now our tests are tied to details of the scales calculation.  (I'm viewing it similar to tests matching against non-API messages from libraries with the expectation that they won't change, but perhaps this calculation is more set in stone.)

I'm fine with leaving this as is, and adjusting later if it turns out to be brittle.  (I see there are other tests in this file that already do this.)

R/model-tree.R
barrettk: Well im using `size_tree_by` for some error handling, so id have to return something to indicate a change to make it "leafCount" (probably not the `node_size` being 1 since that feels sloppy and prone to unintended consequences).  Im am open to making that the default if you think that's preferable

kyleam: I don't feel strongly about what the fallback is given it's for a mis-specified case.  It does seem odd to me to let how you structured an internal helper be the reason for why you should set it to a constant rather than leave it at a default, but there are more important things to focus on with this change.

barrettk: I addressed this and your next comment in [e2c0909](https://github.com/metrumresearchgroup/bbr/pull/720/commits/e2c0909c88736176ccee06a70505f66231115e22). Let me know what you think about that handling.

R/model-tree.R
barrettk: Addressed in [e2c0909](https://github.com/metrumresearchgroup/bbr/pull/720/commits/e2c0909c88736176ccee06a70505f66231115e22). Will mark this one as resolved for now, and any follow up regarding either this or the previous comment can be discussed in that thread since the referenced commit refactored both sections in the same way.

R/model-tree.R
kyleam: I think the behavior change is good.  My initial reaction is that I don't find sneaking it through an attribute easy to follow, but I think it's fine.  Let's leave this be.

R/model-tree.R
barrettk: Great catch. I could see reasons for both in different situations (factor if there's no conceptual order - like representing SEX or RACE), though I think integers should be treated as numeric in most cases and think that should be the default handling (or really only handling since you wouldnt be able to override that).

R/model-tree.R
barrettk: Conditional logic has been removed in favor of attribute setting for both `size_by` and `color_by` as of [71be083](https://github.com/metrumresearchgroup/bbr/pull/720/commits/71be083f056db14eb61f3919b58609ddfcc353f9) (`size_by` addressed in [e2c0909](https://github.com/metrumresearchgroup/bbr/pull/720/commits/e2c0909c88736176ccee06a70505f66231115e22))

R/model-tree.R
barrettk: Edit: I should add that the default of _this function_ is leafCount, though the default of `collapsibleTreeNetwork` is the handling I had before. From the docs:

> numeric column that will be used to determine relative node size. Default is to have a constant node size throughout. 'leafCount' can also be used here (cumulative count of a node's children), or 'count' (count of node's immediate children).

I still think the change I made makes sense given the docs for `model_tree`. For our use case I think `leafCount` makes sense as the default conceptually too.

kyleam: > I should add that the default of _this function_ is leafCount, though the default of `collapsibleTreeNetwork` is the handling I had before.

Hmm, I having a hard time figuring how that's relevant if it's never been exposed to model_tree users.  Perhaps I'm missing something.

To me, the point is just simply to align the fallback with the default `model_tree` behavior.  (Whether leafCount is a good default is of course another question, but that's already been decided before this PR.)

barrettk: >having a hard time figuring how that's relevant if it's never been exposed to model_tree users

Not relevant to users, just thought you may want to know that. I don't believe that's why I set `tree_data$node_size <- 1` previously (think I just wanted to make it obvious an incorrect `size_by` column was specified so they'd correct their code, as well as keeping the returned data frame consistent) but just noticed that now when looking through the docs.

R/model-tree.R
```diff
@@ -95,74 +108,96 @@ model_tree <- function(
...
   # Compile attributes into tooltip
-  tree_data <- make_tree_tooltip(tree_data, digits = digits)
+  tree_data <- make_tree_tooltip(tree_data, digits = digits, font_size = font_size)
 
   # Create model tree
-  tree_attr <- ifelse(is.null(color_by), "leafCount", color_by)
+  # - Notes about aggFun:
+  #  - identity is not the same as base::identity. collapsibleTree has specific
+  #  handling for `aggFun = identity`, and it's the only way for sizing to work
+  #  based on a column without aggregating values.
```
kyleam, line 189: Oh! Well that clears up my main confusion.  Definitely worth a code comment.  Thanks.

R/model-tree.R
barrettk: @kyleam I did a lot of investigating about `aggFun` and left details in the comment made in [0cf42e4](https://github.com/metrumresearchgroup/bbr/pull/720/commits/0cf42e40fa2ece78a977f4ab5db14f7942479c04).

passing `aggFun = identity` is not the same as `aggFun = base::identity` because `collapsibleTree` has specific handling for when that name is matched (see [here](https://github.com/AdeelK93/collapsibleTree/blob/master/R/collapsibleTreeNetwork.R#L204-L220)). There is additional `data.tree` nuance too, but I couldnt find a way to fix the issue of inconsistent sizing. I tried providing a custom `aggFun` and changing the data beforehand but nothing seemed to work. Im ok to merge with this known "quirk", but let me know if this should be mentioned in the documentation too.

R/model-tree.R
```diff
@@ -666,16 +724,79 @@ color_tree_by <- function(tree_data, color_by = "run"){
...
+      mean_val <- mean(tree_data$node_size, na.rm = TRUE)
+      sd_val <- sd(tree_data$node_size, na.rm = TRUE)
+
+      if(sd_val != 0){
+        tree_data$node_size <- (tree_data$node_size - mean_val) / sd_val
+      }else{
+        # Would only happen if all values are the same
+        # - avoids dividing by 0
+        tree_data$node_size <- rep(1, length(tree_data$node_size))
+      }
```
kyleam, line 774: Are you sure any of this is needed?

```R
xs <- rnorm(100, mean = -30, sd = 70)
all.equal(scales::rescale(xs, to = c(0, 3)), scales::rescale((xs - mean(xs)) / sd(xs), to = c(0, 3)))
#> [1] TRUE
xs <- sample(1:1000, 100)
all.equal(scales::rescale(xs, to = c(0, 3)), scales::rescale((xs - mean(xs)) / sd(xs), to = c(0, 3)))
#> [1] TRUE
```

R/model-tree.R
kyleam: Thanks for looking into it.  A comment there is definitely helpful.

> Im ok to merge with this known "quirk", but let me know if this should be mentioned in the documentation too.

I think it's fine to leave as is too, at least for now.

R/model-tree.R
barrettk: I asked myself the same question and had trouble coming up with good examples to test but your example proves it to me. I Noticed that passing `rescale_to = c(1, 100)` made the size differences incredibly large and was trying to normalize to the standard deviation to avoid that, but I seem to have had a brain fart.

kyleam (approved):

PR #724, opened 2024-12-06
Add submission mode for Slurm

kyleam:
This is the sister PR to metrumresearchgroup/bbi#338 and adds support for calling `bbi nonmem run slurm` by passing `.mode = "slurm"` to  `submit_model`.

Most of the commits in this series are minor cleanup of the tests that check dry-run calls to `submit_model`.  The seventh commit makes the main change, which boils down to registering the new mode and adjusting the `check_mode_argument` helper.

---

Reminder for testing: by default, bbr will fail if a developmental version of bbi is used.  You can prevent that by setting the `bbr.DEV_no_min_version` option to `TRUE`.

### DIFF
--- a/R/submit-model.R
+++ b/R/submit-model.R
@@ -30,9 +30,9 @@
 #'   4)`. Run [print_bbi_args()] to see valid arguments. Note that bbr does
 #'   not support changing the output directory (including through the model or
 #'   global YAML files).
-#' @param .mode Either `"sge"`, the default, to submit model(s) to the grid or
-#'   `"local"` for local execution. This can be passed directly to this argument
-#'   or set globally with `options("bbr.bbi_exe_mode")`.
+#' @param .mode Mode for model submission: "local", "sge", or "slurm". If
+#'   unspecified, the value is set to the value of the `bbr.bbi_exe_mode`
+#'   option. This option defaults to "sge" on Linux and "local" otherwise.
 #' @param ... args passed through to `bbi_exec()`
 #' @param .overwrite Logical to specify whether or not to overwrite existing
 #'   model output from a previous run. If `NULL`, the default, will defer to
@@ -354,6 +354,38 @@ check_mode_argument <- function(.mode) {
 
   if (!(.mode %in% BBI_VALID_MODES)) {
     stop(BBI_EXE_MODE_INVALID_ERR_MSG, call. = FALSE)
+  }
+
+  if (isTRUE(getOption("bbr.DEV_skip_system_mode_checks"))) {
+    # ^ Allow dry-run tests to skip the system mode checks.
+    return(invisible(TRUE))
+  }
+
+  # All of the remaining checks depend on system details.
+
+  if (identical(.mode, "slurm") && !test_bbi_version(.min_version = "3.4.0")) {
+    stop(
+      "Installed bbi version is ", bbi_version(),
+      ", but .mode='slurm' requires at least version 3.4.0"
+    )
+  }
+
+  if (identical(.mode, "sge")) {
+    qsub <- unname(Sys.which("qsub"))
+    if (identical(qsub, "")) {
+      stop(".mode='sge' but qsub is not available on system")
+    }
+
+    # This guard is Metworx (or really ParallelCluster) specific. Slurm ships
+    # with a qsub shim. That leads to a confusing situation where 'bbi nonmem
+    # run sge ...' will not abort upfront, but the execution doesn't entirely
+    # match the expected behavior (especially under --parallel).
+    if (identical(qsub, "/opt/slurm/bin/qsub")) {
+      stop(
+        ".mode is 'sge' but qsub points to Slurm shim.\n",
+        "Instead set .mode to 'slurm' (requires bbi 3.4.0 or later)\n"
+      )
+    }
   }
 
   return(invisible(TRUE))
--- a/tests/testthat/test-submit-model.R
+++ b/tests/testthat/test-submit-model.R
@@ -1,4 +1,4 @@
-context("submit_model(.dry_run=T)")
+context("submit_model(.dry_run=TRUE)")
 
 ###################################
 # testing single model submission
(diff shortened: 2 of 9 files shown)


tests/testthat/test-submit-model.R
```diff
@@ -1,117 +1,158 @@
...
 })
 
+test_that("submit_model(.mode) errors when invalid", {
+  expect_error(
+    submit_model(MOD1, .dry_run = TRUE, .mode = "naw"),
+    regexp = "Invalid value passed.+mode"
+  )
+})
+
+test_that("submit_model aborts if .mode='sge' is used with Slurm's qsub", {
```
kyleam, line 117: The tests from here down are new.

tests/testthat/test-submit-models.R
```diff
@@ -1,199 +1,200 @@
...
+context("submit_models(.dry_run=TRUE)")
 
 #####################################
 # testing multiple model submission
 #####################################
 
-withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
+withr::local_options(list(
+  bbr.bbi_exe_path = read_bbi_path(),
+  bbr.DEV_skip_system_mode_checks = TRUE
```
kyleam, line 9: This option is new, but there is no new test content.

R/submit-model.R
```diff
@@ -356,6 +356,38 @@ check_mode_argument <- function(.mode) {
...
+    qsub <- unname(Sys.which("qsub"))
+    if (identical(qsub, "")) {
+      stop(".mode='sge' but qsub is not available on system")
+    }
+
+    # This guard is Metworx (or really ParallelCluster) specific. Slurm ships
+    # with a qsub shim. That leads to a confusing situation where 'bbi nonmem
+    # run sge ...' will not abort upfront, but the execution doesn't entirely
+    # match the expected behavior (especially under --parallel).
+    if (identical(qsub, "/opt/slurm/bin/qsub")) {
```
seth127, line 383: I like this idea, but I want to think through situations where a user may have been previously using bbr/bbi with Slurm via the `qsub` shim... If they now upgrade to this (upcoming) bbr release, will it be as simple as this?

* They start seeing this error
* They need to upgrade `bbi` and switch their code to explicitly call `.mode = "slurm` (or set the option)
* It should work as it did before

It seems like there may be some nuance that I'm missing though. Maybe we don't need to try to go too far to cover these somewhat hacky cases, but I want to at least discuss it.

tests/testthat/test-submit-model.R
```diff
@@ -111,3 +114,46 @@ test_that("submit_model(.mode) errors when invalid", {
     regexp = "Invalid value passed.+mode"
   )
 })
+
+test_that("submit_model aborts if .mode='sge' is used with Slurm's qsub", {
+  skip_if_old_bbi("3.4.0")
+
+  sbatch <- unname(Sys.which("sbatch"))
+  if (identical(sbatch, "") || identical(Sys.getenv("METWORX_VERSION"), "")) {
+    skip("not on Metworx with SGE")
```
seth127, line 122: Is this supposed to be `"not on Metworx with Slurm"` (instead of `SGE`)?

It seems like you would hit this when `sbatch` doesn't show up, and the point of this test is for systems _with_ Slurm.

seth127 (commented):
I stopped by here first, on my way to metrumresearchgroup/bbi#338, and dropped a couple of comments.

All looking good overall, but I'm going to look at `bbi` a bit and do some manual testing before I come back around to approve this. 

As usual, thank for the detailed commit messages and the nicely organized clean up.

tests/testthat/test-submit-model.R
kyleam: Doh, thanks for catching that.

R/submit-model.R
kyleam: > [...] I want to think through situations where a user may have been previously using bbr/bbi with Slurm via the `qsub` shim...

I agree this is worth thinking through.  Here's the rationale from ba073b1b for this (admittedly aggressive) approach:

```
...

Slurm includes a qsub shim that has some behavioral differences when
used under 'bbi nonmem run sge ...' (e.g., output files going to the
parent directory).  That shim would very likely cause confusion when
people on Slurm machines forget to override the "sge" default.

Prevent that confusion on Metworx or other systems using
ParallelCluster by aborting if "sge" is specified for .mode and qsub
is Slurm's shim.  That's aggressive and perhaps will be triggered by
people that have already been making of the shim.  However, it's
better to be strict about it because it's not an option we intend to
support.
```

To rephrase and expand a bit:

 1) I'm concerned about users unintentionally selecting "sge" given the default

 2) I was seeing some pretty wonky results during my (unsystematic) testing  of `bbi nonmme run sge` with Slurm's qsub shim.  Two things that come to mind (both with `--parallel`, I think) is 1) some generated files being placed outside the output directory and 2) the `--threads` value not being respected.

> If they now upgrade to this (upcoming) bbr release, will it be as simple as this? [...]

I think that captures it, though "should work as it did before" will hopefully not include any weirdness like what I mentioned above.

> Maybe we don't need to try to go too far to cover these somewhat hacky cases, but I want to at least discuss it.

Definitely worth discussing, and I'm open to other ideas here.  As mentioned above, I would be concerned with removing the guard entirely.

If some users rely on this shim heavily, they could abuse the `bbr.DEV_skip_system_mode_checks` option to prevent the guard from firing, at least in the short term.  If turns out that the shim is relied on more than we think, we can elevate that to a non-dev option and document it.

tests/testthat/test-submit-model.R
kyleam: I've amended the commit to say Slurm instead of SGE ([diff](https://github.com/metrumresearchgroup/bbr/compare/25a7f1c5dbc9371457c1960939d743c4540ab997..01f39e6fd4a7bbc6dfc419336e83053a1ba7dfad)).

kyleam:
> I did some manual testing [...]

Thanks for giving it thorough testing.

Issue #725, opened 2024-12-10
model_tree fails if no parent model exists in model directory


barrettk:
Current error message:
```r
> model_tree(modDir)
Error in data.frame(from = start, to = full_log$run[full_log$based_on ==  :                                                                                   
  arguments imply differing number of rows: 1, 0
```

#### Failure
This happens when there is no "starting model"
```r
> run_log(modDir) %>% tidyr::unnest(based_on)
# A tibble: 3 × 10
  absolute_model_path                           run   yaml_md5 model_type description bbi_args based_on tags   notes  star 
  <chr>                                         <chr> <chr>    <chr>      <chr>       <list>   <chr>    <list> <list> <lgl>
1 /data/Projects/trainings/AZ_project_template… 102   feeedbe… nonmem     Two-compar… <NULL>   101      <NULL> <chr>  TRUE 
2 /data/Projects/trainings/AZ_project_template… 106-… a0be936… nmboot     NA          <NULL>   106      <chr>  <NULL> FALSE
3 /data/Projects/trainings/AZ_project_template… 106   8006f38… nonmem     Two-compar… <NULL>   102      <NULL> <chr>  TRUE 
```
This happens because run 102 is based on run 101, which was deleted from the directory. We currently have error handling for when any other model "disappears", but not when the starting/origin model does. There currently must be at least one starting/origin model to avoid this error. This is a very unlikely scenario since you'd usually want to keep the origin model around for traceability purposes, which is why this bug slipped through.

#### Working
If I remove the `based_on` attribute from run 102, it becomes the starting model, which allows the tree diagram to render:
```r
> run_log(modDir)
# A tibble: 3 × 10
  absolute_model_path                           run   yaml_md5 model_type description bbi_args based_on tags   notes  star 
  <chr>                                         <chr> <chr>    <chr>      <chr>       <list>   <list>   <list> <list> <lgl>
1 /data/Projects/trainings/AZ_project_template… 102   887d872… nonmem     Two-compar… <NULL>   <NULL>   <NULL> <chr>  TRUE 
2 /data/Projects/trainings/AZ_project_template… 106-… a0be936… nmboot     NA          <NULL>   <chr>    <chr>  <NULL> FALSE
3 /data/Projects/trainings/AZ_project_template… 106   8006f38… nonmem     Two-compar… <NULL>   <chr>    <NULL> <chr>  TRUE 

> run_log(modDir) %>% tidyr::unnest(based_on)
[...]

PR #726, opened 2024-12-11
bug fix: model_tree works if starting/orign model is missing

barrettk:
`model_tree` previously supported "broken links". i.e. when a model referenced as "based on" was no longer present in the model directory (e.g., deleted). However this handling was previously insufficient if no "origin" model was found.

If the origin model is missing, we still create the node since it was referenced in a model yaml file. 
 - I think this is important since a key use for this function is to visualize the modeling process. This, along with the warning message users would see, will help modelers ensure model yamls are up-to-date and the models they expect to exist still do.

### Example
Here is an example warning message a user would see if mods 1 and 1000 were missing (where mod 1 is the origin model):
```
>  model_tree(run_df, add_summary = FALSE)
Warning message:
The following models could not be linked properly: "1000" and "1"
ℹ Setting these as starting models in tree diagram.
ℹ Check the yaml files or run bbr::run_log() to make sure all `based_on` models still exist 
```
#### Output of this scenario:
Missing `mod 1` (origin model)            |  Missing `mod 1000`
:-------------------------:|:-------------------------:
image  | image

closes https://github.com/metrumresearchgroup/bbr/issues/725

### DIFF
--- a/R/model-tree.R
+++ b/R/model-tree.R
@@ -165,6 +165,7 @@ model_tree.bbi_log_df <- function(
     font_size = 10,
     ...
 ){
+  check_bbi_run_log_df_object(.log_df)
   # Make sure required dependencies are installed
   stop_if_tree_missing_deps(static = static)
 
@@ -343,7 +344,6 @@ make_tree_data <- function(
 #' @param .log_df a `bbr` run log
 #' @noRd
 unnest_based_on <- function(.log_df){
-  check_bbi_run_log_df_object(.log_df)
   .log_df %>% dplyr::mutate(
     # Replace NULL based_on elements with empty string to preserve rows when unnesting
     based_on = purrr::map(.data$based_on, function(.x){if(is.null(.x)) "" else .x}),
@@ -433,7 +433,6 @@ make_model_network <- function(full_log){
 #' @param .log_df a `bbr` run log
 #' @noRd
 add_log_recurse_dirs <- function(.log_df){
-  check_bbi_run_log_df_object(.log_df)
   checkmate::assert_true(all(c(ABS_MOD_PATH, "based_on") %in% names(.log_df)))
   checkmate::assert_character(.log_df$based_on)
 


R/model-tree.R
```diff
@@ -381,14 +381,25 @@ make_model_network <- function(full_log){
...
+  parent_mods <- full_log$run[full_log$based_on==""]
+  if(!rlang::is_empty(parent_mods)){
+    parent_mods_df <- data.frame(from = start, to = parent_mods)
+    network_df <- rbind(noref_df, parent_mods_df, child_mods_df)
+  }else{
+    # This happens when there is no "starting" or origin model
+    # - i.e. the first referenced based_on model no longer exists
+    network_df <- rbind(noref_df, child_mods_df)
+  }
+
```
kyleam, line 400: Stepping through this for both the origin and no origin cases, this logic made sense to me.

tests/testthat/test-model-tree.R
```diff
@@ -117,6 +117,43 @@ withr::with_options(list(bbr.bbi_exe_path = read_bbi_path()), {
...
+      #  - broken_link creates another path from mod1 and deletes the first one
+      create_tree_models(mod1 = mod1, broken_link = TRUE)
+      delete_models(mod1, .tags = NULL, .force = TRUE)
+
+      run_df <- run_log(MODEL_DIR_BBI)
+
+      expect_warning(
+        tree_data <- make_tree_data(run_df, add_summary = FALSE),
+        "The following models could not be linked properly"
+      )
```
kyleam, line 147: Why do this `make_tree_data` call separately rather than use the `expect_warning` for the main `model_tree` call on line 149?  Does the cover something that doesn't?

R/model-tree.R
```diff
@@ -381,14 +381,25 @@ make_model_network <- function(full_log){
...
-  # Check network links - set unlinked models as starting points
+  noref_df <- data.frame(from = NA, to = start)
+
+  child_mods_df <- data.frame(
+    from = full_log$based_on[full_log$based_on!=""],
+    to = full_log$run[full_log$based_on!=""]
+  )
+
+  parent_mods <- full_log$run[full_log$based_on==""]
+  if(!rlang::is_empty(parent_mods)){
```
kyleam, line 392: nit-pick: if you negate the condition of an if/else, it is almost always more readable (to my eyes) to drop the negation (i.e. remove `!` and flip the if/else arms

kyleam (approved):
The fix looks good to me.  Thanks for the details here, including the nice write-up in gh-725.

I left a couple of minor comments for your consideration but nothing blocking.

Let's wait to merge this until gh-720 lands.

tests/testthat/test-model-tree.R
barrettk: >Does the cover something that doesn't?

No, I guess I was just trying to "zoom in" as much as possible to show the highest level function this could be triggered on without having to manually create a custom dataset. I can make that change if you think that would be more clear.

kyleam: Up to you.  I'm good with merging this as is, but, yes, I was confused about why the inner helper was used there.  (I'm not saying it's not useful to do a lot of times, especially for exhaustively testing cases that are more  convenient to hit from the inner function.)

R/model-tree.R
barrettk: Fair. I like to have the more likely scenario appear first but I see the argument for that

PR #729, opened 2025-01-02
Fix nm_data(filter = TRUE)

barrettk:
- Previously returned no rows if column is read in as all NA and a string detection filter was applied (e.g., C='C').
 - Instead of adding `C!='C'`, we now append an `is.na()` expression, which would yield `(C!='C' | is.na(C)`.

closes https://github.com/metrumresearchgroup/bbr/issues/723, https://github.com/metrumresearchgroup/bbr/issues/727

PR #730, opened 2025-01-03
update IGNORE=C to work like IGNORE=@

kylebaron:
Proposing that when we encounter `IGNORE=C`, we _only_ look for that character in column 1.  

From NM-help, `$DATA`

```
 IGNORE=c1

      Specifies that any data record having the character c1 in  column
      1  should be ignored, i.e., these records are not included in the
      NONMEM data set.  This allows comment records to be  included  in
      the  NM-TRAN  data set.  In general, records having the character
      c1 in column 1 will be called "comment records".
      Also permitted: IGNORE='c' or IGNORE="c",  where  c  may  be  any
      character  except  space.   IGNORE=# is the default.  That is, in
      the absence of IGNORE option, any record whose first character is
      # is treated as a comment record.
```

### DIFF
--- a/R/filter-nm-data.R
+++ b/R/filter-nm-data.R
@@ -203,9 +203,9 @@ translate_nm_expr <- function(
       }else if(grepl('^[a-zA-Z]$', expr)){
         # This is for `IGNORE=C` columns. Meaning ignore rows if the _first_ column
         # contains 'C' (this form always points to the _first_ column)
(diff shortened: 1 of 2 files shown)


barrettk (approved):
This change looks good and makes sense based on our conversation [here](https://github.com/metrumresearchgroup/bbr/pull/729#discussion_r1902205706) and offline. I adjusted the test and added a new one for this case in the last commit

Issue #677, opened 2024-04-26
dOFV and AIC columns for summary_log()


seth127:
We would like to add several columns to the `summary_log()` output.

## AIC

This is a straightforward calculation. Consider if there are any edge cases to catch.

**Example code**

```r
add_aic <- function(.runlog) {
  ## check summary log appended
  if(any(str_detect(names(.runlog), "param_count")) & 
     any(str_detect(names(.runlog), "ofv"))) {
    return(mutate(.runlog, aic=2*param_count + ofv))
  }
  ## if not, return unchanged
  warning("AIC could not be computed")
  return(.runlog)
}
```

## dOFV

The change in OFV from the parent model. Several considerations:

* What to do if there are multiple models in `based_on`? Likely select the first?
* What to do if the based on model has `NA` for `ofv`? This could happen with Bayesian models, or the new bootstrap model type.

**Example code**

```r
## compare ofv to that of the model in based_on
## if multiple models are in based_on, use the first
add_dofv <- function(.runlog) {
  .runlog %>%
    mutate(based_on_join = purrr::map_chr(based_on, function(x) {ifelse(length(x) > 0, x[1], NA)})) %>% 
    left_join(
      select(.runlog, based_on_join=run, based_on_ofv=ofv), 
      by="based_on_join"
    ) %>% 
    mutate(dofv = as.numeric(format(ofv - based_on_ofv, nsmall = 2))) %>% 
    select(-based_on_ofv, -based_on_join)
}
```

## dAIC

If we're adding the above, should we also calculate the difference in AIC?

seth127:
Consider also adding a method for easily comparing two model objects (as opposed to a log tibble), similar to the discussion in https://github.com/metrumresearchgroup/bbr/issues/401

barrettk:
Once consideration worth talking about: summary logs dont have a `based_on` column, which is required for `add_dofv`. Do we:
 1) want to only implement this for `run_log() %>% add_summary()`, or
 2) add `based_on` to the `summary_log` (and maybe even the `config_log()`). 
    - I dont think it would make sense to remove the `based_on` column when `dofv` is present (i.e. dont just add it to compute `dofv`, and then remove), as this could cause confusion.

PR #731, opened 2025-01-07
Add aic dofv

barrettk:
Add AIC, BIC, and DOFV columns to the `summary_log`

closes https://github.com/metrumresearchgroup/bbr/issues/677

_Summarizing comment from @seth127 [here](https://github.com/metrumresearchgroup/bbr/pull/731#issuecomment-2794771300)._


barrettk:
Still need to write tests, but the prototype is complete. I think this is a good place to gather feedback before continuing.

barrettk:
@callistosp can you take a look at [how I calculated BIC](https://github.com/metrumresearchgroup/bbr/blob/03f425809c4add3735c27dea8ce3fa0fe0bf5e84/R/summary-log.R#L301-L306) when you get a chance?

I think we'll want to have some more design discussions (from both a user/scientist and technical perspective) too, but wanted to confirm that part early on.

callistosp:
LGTM

R/summary-log.R
```diff
@@ -72,9 +72,23 @@ add_summary <- function(
 #' @importFrom tibble tibble
 #' @importFrom tidyselect all_of
 #' @param .mods List of model objects that will be passed to [model_summaries()].
+#' @param calc_aic_bic Logical(T/F). If `TRUE`, calculate the Akaike Information
```
kylebaron, line 75: @barrettk Does this change the usage pattern? requesting something in the output through a function argument? Why not just put it in the output and let people either select or deselect it?

kylebaron: I was hoping that AIC and BIC are added by default and I think it is now. 

The dofv part does look like there is a ton of extra computation that _must_ be done to get a run log out (whether you want dofv or not). I think I'd rather just opt into that later on if it were needed; like move this `add_dofv()` function to be something you call on the summary log object.

kylebaron: Maybe the only practical way I can get a run log out without a warning is by asking for all the models. 

I guess there's some comittment to allowing the log to span multiple directories (with `.recurse = TRUE`) and that the linking could span the directories. IMO this is more of an edge case rather typical scenario; especially forking a model from one directory to another. 

Rather than having to map the _entire_ model tree / space, what if you created this data frame object and sent it back to the user. The user could then opt in to do the computation. 

You could potentially calculate dofv from information already in the log object (if the user requested the based_on run in the log). If not, you could just traverse the rows of that object, and quickly look at the based_on run and get the ofv. 

Maybe this pattern is too straightforward to be desirable. In that case, let the user opt in to the dofv calculation. 

One more note that we might want to consider: you might have at least equal chance of wanting to calculate dofv relative to a single model in the run log; I'd say I'd want this most of the time. In that case, you could open up this `add_dofv()` function to either do the work to get the reference ofv from each base model or just look up one model to set the base and then calculate dofv off that. Maybe you could allow both. But I think separating this calculation out would clarify things and avoid lots of additional / required computation.

kylebaron: To clarify my comments so far

1. I want to opt into the `dofv` calculation later, especially if it requires lots of extra calc
2. I want to choose how `dofv` is calculated, either via `based_on` or from a single reference or both

barrettk: @kylebaron 
>I see now that these are for internal control only?

Yes - I wanted to be able to set this to `FALSE` for bootstrap runs for example.

>I was hoping that AIC and BIC are added by default and I think it is now.

Yes for normal run or summary logs it will automatically happen

>The dofv part does look like there is a ton of extra computation that must be done to get a run log out (whether you want dofv or not). I think I'd rather just opt into that later on if it were needed; like move this add_dofv() function to be something you call on the summary log object.

Unfortunately yes, and as you noted, it's due to needing to support `.recurse`. Taking that into account significantly complicates things. I know it's not typical, though it would feel wrong to diverge here when other functions support it. If we kept this approach I would of course have to adjust the warning messages you see (so it doesnt mention `model_tree()`, but I was also thinking of revisiting how we map all the runs in general, and pulling that logic out of `make_tree_data` (dedicated PR). Felt we had to have some discussion before doing that though.

>Maybe the only practical way I can get a run log out without a warning is by asking for all the models.

Regarding the filtering issue you mention, thanks for noting this! I hadn't messed with the `.include` argument. Im thinking we may want to have this mapping of models happen during `find_models()` before the filtering takes place, and store this metadata for later use. 

>One more note that we might want to consider: you might have at least equal chance of wanting to calculate dofv relative to a single model in the run log; I'd say I'd want this most of the time. In that case, you could open up this add_dofv() function to either do the work to get the reference ofv from each base model or just look up one model to set the base and then calculate dofv off that. Maybe you could allow both. But I think separating this calculation out would clarify things and avoid lots of additional / required computation.

This is interesting. The idea of wanting to compare both cases definitely supports the idea of calculating `dofv` later, though I dont think we'd want the mapping to occur in `find_models()` in that case since we'd already have a run log, and we dont want to introduce new columns we have to keep around. Will think on this. 
 - cc'ing @callistosp for his opinion as well. 
 - Would it make sense to calculate `dofv` using some standard in the `summary_log`, but allow users to create a custom table later? Or does it just make sense to exclusively do this separately?

callistosp: if you allow multiple `based_on`s, you are going to turn `dofv` into a list column. I guess that isn't a huge issue, but I would prefer to just have it default to using the first `based_on` value to calculate `dofv`, and if the user wants additional ofvs they could manually calculate them. In my opinion multiple dofv values seems like an edge case that is not common enough to necessitate the amount of work to implement it so flexibly.

barrettk:
@seth127 FYI this PR is failing because of ubuntu-20.04 being removed. The PR to address that is [here](https://github.com/metrumresearchgroup/bbr/pull/741), though I think we need to adjust github actions too once that lands

barrettk:
FYI to whomever reviews this PR: it's functionally ready to go, but I did have some concerns about the tests, specifically for calculating `dofv`. Given that we only have one model in the repo, we would only ever get a difference of `0`, without adjusting the model and re-running it. 

Worth thinking through if we feel the need to do something like this, though perhaps a careful code review is sufficient.

seth127:
We discussed offline and @kyleam is going to give this a review. There was a fair amount of discussion (both here and offline), so I'll try to summarize where we landed on a few things:

1. AIC/BIC will be added to `summary_log()` output, by default.
2. dOFV (change in objective function value) will be implemented in a separate function, which can take a log table as input.
3. Some specs on the dOFV function:
  i. It should compare the input model(s) to a _single model_, which will produce a single output value (for each input model). 
  ii. By default, it should compare each model (e.g. `.mod`) to the _first model_ in that model's `based_on` field (e.g., `get_based_on(.mod)[1]`)
  iii. Optionally, the user can pass a different model for comparison. In this case all input model(s) are compared against the passed model. This allows creative users to code up various edge cases (e.g., comparing all models to a single "base model", comparing to multiple `based_on` models, etc.).
  iv. For the log table input, dOFV should be computed for all models in the table, as long as their `based_on` model can be found (i.e., even if a given `based_on` model is _not_ in the table itself). In cases where the `based_on` cannot be found, `NA` can be returned for that row, but the function should _not_ error.

**One final note:** The current implementation (as of ce164d261e5157372f27ed08614bbc753e91a690) likely calls `read_model()` and (more importantly) `model_summary()` more often than it needs to. We should consider whether this is worth optimizing at this point. This may be unnecessary (given the scale of, at most, several hundred iterations) and doing so may cause the code to become unnecessarily complicated. That said, if review leads to a more efficient implementation being proposed, that may be preferable. 

Related to this last point, if we end up reimplementing parts of the dOFV function, I am fine with pulling the AIC/BIC part into a separate PR as well, to more easily focus on the two different pieces being implemented here.

R/summary-log.R
```diff
@@ -72,9 +73,20 @@ add_summary <- function(
...
+#'
+#' @details
+#' `calc_aic_bic` should be set to `FALSE` for certain model types, such as
+#'  bootstrap runs.
+#'
 #' @keywords internal
-summary_log_impl <- function(.mods, ...) {
+summary_log_impl <- function(
+    .mods,
+    calc_aic_bic = TRUE,
```
kyleam, line 87: This argument gets exposed to the user via `...`, which is documented as getting passed to `model_summaries`.

In [this message](https://github.com/metrumresearchgroup/bbr/pull/731#discussion_r1954944405), you say it's for "internal control only". In that case, you could change the internal spots to use `summary_log_impl` (e.g., `summarize_bootstrap_run`) and avoid exposing it externally (e.g., by adjusting `summary_log` and `add_summary` to pass `calc_aic_bic=TRUE` explicitly).

If the intention is to expose it, then `calc_aic_bic` should be added as an argument to `summary_log` and documented.

R/summary-log.R
```diff
@@ -266,3 +281,153 @@ do_if_bbi_sum <- function(fn, mode) {
...
   }
 }
+
+#' Calculate Akaike Information Criterion (AIC) and Bayesian Information
+#' Criterion (BIC), and append to summary log
+#' @param .log_df The output of [summary_log()]) or `run_log() %>% add_summary()`
+#' @keywords internal
+add_aic_bic <- function(.log_df){
+  .log_df %>% dplyr::mutate(
+    !!rlang::sym(AIC_COL) := 2*.data[[PARAM_COUNT_COL]] + .data[[OFV_COL]],
```
kyleam, line 291: Why is the `rlang::sym` call needed here? Doesn't `!!AIC_COL := ...` give the same result?

(Same comment applies to lines 292, 393-395, and 425-427.)

R/summary-log.R
```diff
@@ -266,3 +281,153 @@ do_if_bbi_sum <- function(fn, mode) {
...
+    !!rlang::sym(BIC_COL) := .data[[PARAM_COUNT_COL]]*log(.data$number_of_obs) + .data[[OFV_COL]]
```
kyleam, line 292: Both these calculations rely on the NONMEM objective function value being `-2 log-likelihood`, correct? It'd probably be good to document that with a comment.

R/summary-log.R
```diff
@@ -266,3 +281,153 @@ do_if_bbi_sum <- function(fn, mode) {
...
+
+
+#' Calculate the difference in objective function value
+#'
+#' Calculate the difference in objective function value (`ofv`) between the
+#' `.bbi_object` and a reference model.
+#'
+#' @details
+#' If `.bbi_object` is a `tibble` of class `bbi_log_df`, `"absolute_model_path"`
+#' is the only required column. If the `tibble` is the output of [bbr::run_log()]
```
kyleam, line 304: Looking at the guard in `add_dofv.bbi_log_df` (line 354), "run" is also a required column. So there's a code/doc mismatch, though I don't actually spot the "run" column used downstream, so perhaps you want to adjust the code instead of the docs.

R/summary-log.R
```diff
@@ -266,3 +281,153 @@ do_if_bbi_sum <- function(fn, mode) {
...
+#' If `.bbi_object` is a `tibble`, `add_dofv()` appends a `dofv` column.
+#'
+#' If `.bbi_object` is a `bbi_nonmem_model` object, `add_dofv()` returns a numeric value.
+#' @export
+add_dofv <- function(.bbi_object, .mod2 = NULL){
+  UseMethod("add_dofv")
+}
+
+#' @export
+add_dofv.bbi_nonmem_model <- function(.bbi_object, .mod2 = NULL){
```
kyleam, line 343: `add_dofv` doesn't seem like a good name for a method that takes a model and returns an scalar (rather than adding/extending to the passed log).

My opinion is that, at least at this point, we shouldn't bother adding an S3 method and should instead convert `add_dofv` to a regular function that takes a run/summary log.

Another option would be to change the name to something that fits both scenarios better.

(I realize there was some previous discussion that landed on this. I discussed this offline with @seth127 and he was on board with switching to a regular function. He can of course chime in if his thoughts have since changed on that.)

R/summary-log.R
```diff
@@ -266,3 +281,153 @@ do_if_bbi_sum <- function(fn, mode) {
...
+  if(!check_nonmem_finished(.bbi_object)){
+    cli::cli_abort("Model {get_model_id(.bbi_object)} has not finished running")
+  }
+  dofv_df <- calc_dofv_impl(.bbi_object, .mod2 = .mod2)
+  return(dofv_df$dofv)
+}
+
+#' @export
+add_dofv.bbi_log_df <- function(.bbi_object, .mod2 = NULL){
+  # Check for required columns and starting format
```
kyleam, line 353: I don't know what "starting format" refers to.

R/summary-log.R
```diff
@@ -266,3 +281,153 @@ do_if_bbi_sum <- function(fn, mode) {
...
+  req_cols <- c(ABS_MOD_PATH, RUN_ID_COL)
+  if(!(all(req_cols %in% names(.bbi_object)))){
+    cols_missing <- setdiff(req_cols, names(.bbi_object))
+    cli::cli_abort(
+      "The following {.emph required} columns are missing: {.val {cols_missing}}"
+    )
+  }
+
+  mods <- purrr::map(.bbi_object[[ABS_MOD_PATH]], ~ read_model(.x))
```
kyleam, line 362: nit-pick: `~ read_model(.x)` simplifies to `read_model`

R/summary-log.R
```diff
@@ -266,3 +281,153 @@ do_if_bbi_sum <- function(fn, mode) {
...
+  dofv_df <- purrr::map_dfr(mods, calc_dofv_impl, .mod2 = .mod2)
```
kyleam, line 363: It'd be good to avoid `map_dfr` given it's marked as superseded.

tests/testthat/test-summary-log.R
```diff
@@ -217,3 +218,47 @@ test_that("summary_log() works with filtering parameter string [BBR-SMLG-011]",
...
+
+  # Same behavior with add_summary
+  log_df <- run_log(MODEL_DIR) %>% add_summary()
+  expect_true(all(c(AIC_COL, BIC_COL) %in% names(log_df)))
+
+  log_df <- run_log(MODEL_DIR) %>% add_summary(calc_aic_bic = FALSE)
+  expect_false(all(c(AIC_COL, BIC_COL) %in% names(log_df)))
+})
+
+test_that("add_dofv() works correctly", {
```
kyleam, line 244: @barrettk:

> I did have some concerns about the tests, specifically for calculating `dofv`. Given that we only have one model in the repo, we would only ever get a difference of `0`, without adjusting the model and re-running it.

Thanks for highlighting this. I agree with that concern.

Rather than rerunning, you can probably get by with some custom test setup that adjusts the `OBJV` line for the test model lst's so that they have different and known values, letting you check for an expected value.

In addition to tweaking the OBJV values and testing for expected values, it'd be good to assert that calls with a run log and summary log for the same set of models give the same dofv values.

kyleam (changes requested):

kyleam:
@seth127:

> The current implementation (as of [ce164d2](https://github.com/metrumresearchgroup/bbr/commit/ce164d261e5157372f27ed08614bbc753e91a690)) likely calls `read_model()` and (more importantly) `model_summary()` more often than it needs to. We should consider whether this is worth optimizing at this point. This may be unnecessary (given the scale of, at most, several hundred iterations) and doing so may cause the code to become unnecessarily complicated. That said, if review leads to a more efficient implementation being proposed, that may be preferable.

I agree it'd be nice to avoid spawning a `bbi nonmem summary` process for each model, though as you suggest the slowdown likely doesn't matter too much in practice (especially since `add_dofv` comes from an explicit call rather than being triggered by `summary_log`).

For comparison, here is an initial crack at a more efficient implementation. It determines upfront the set of models needed to get ofv and summarizes those in one go.

<details>
<summary>code</summary>

```R
add_dofv <- function(.log_df, .mod_ref = NULL) {
  res <- make_ofv_pairs(.log_df, .mod_ref)

  if (!OFV_COL %in% names(.log_df)) {
    .log_df[OFV_COL] <- res[["ofv_mod"]]
  }

  tibble::add_column(
    .log_df,
    !!DOFV_COL := res[["ofv_mod"]] - res[["ofv_ref"]],
    .after = OFV_COL
  )
}

#' Pair each model OFV with a reference OFV
#'
#' If mod_ref is passed, it is used as the reference model. Otherwise, the first
#' parent of each model is taken as the reference model.
#'
#' Return a data frame with the following columns:
#'
#'  * path_mod, path_ref: The absolute paths of the main model and the reference
#'    model. path_mod matches the absolute_model_path column from the input run
#'    log.
#'
#'  * ofv_mod, ofv_ref: The objective function values for the main model and the
#'    reference model.
#'
#' @noRd
make_ofv_pairs <- function(log_df, mod_ref) {
  if (OFV_COL %in% names(log_df)) {
    ofv <- log_df[c(ABS_MOD_PATH, OFV_COL)]
  } else {
    ofv <- tibble::tibble(!!ABS_MOD_PATH := character(), !!OFV_COL := double())
  }

  mpaths_log <- log_df[[ABS_MOD_PATH]]

  if (is.null(mod_ref)) {
    mods <- purrr::map(mpaths_log, read_model)
    mpaths_ref <- purrr::map_chr(mods, function(m) {
      get_based_on(m, .check_exists = FALSE)[1] %||% NA_character_
    })
  } else {
    mods <- NULL
    mpaths_ref <- mod_ref[[ABS_MOD_PATH]]
  }

  if (any(!is.na(mpaths_ref) & !mpaths_ref %in% ofv[[ABS_MOD_PATH]])) {
    mods <- mods %||% purrr::map(mpaths_log, read_model)
    mpaths_extra <- setdiff(mpaths_ref, mpaths_log)
    mods_extra <- purrr::map(mpaths_extra[!is.na(mpaths_extra)], read_model)
    mods_all <- c(mods, mods_extra)
    mpaths_all <- purrr::map_chr(mods_all, ABS_MOD_PATH)
    mods_need <- mods_all[!mpaths_all %in% ofv[[ABS_MOD_PATH]]]
    if (!length(mods_need)) {
      dev_error("mods_need should never come up empty")
    }
    # To extract OFV, call summary_log_impl rather than model_summaries directly
    # to reuse its logic for reshaping the result.
    slog <- summary_log_impl(mods_need, calc_aic_bic = FALSE)
    slog <- slog[!slog[[ABS_MOD_PATH]] %in% ofv[[ABS_MOD_PATH]], ]
    ofv <- dplyr::bind_rows(ofv, slog[c(ABS_MOD_PATH, OFV_COL)])
  }

  tibble::tibble(path_mod = mpaths_log, path_ref = mpaths_ref) %>%
    dplyr::left_join(
      dplyr::rename(ofv, "ofv_mod" = all_of(OFV_COL)),
      by = c("path_mod" = ABS_MOD_PATH)
    ) %>%
    dplyr::left_join(
      dplyr::rename(ofv, "ofv_ref" = all_of(OFV_COL)),
      by = c("path_ref" = ABS_MOD_PATH)
    )
}
```

</details>


Here are some timings (in milliseconds) on a log of four models.


<details>
<summary>benchmark script</summary>

```R
devtools::load_all()

system("git rev-parse HEAD:R")

l <- run_log("inst/model/nonmem/basic/")
sl <- add_summary(l)

mutate(l, based_on = purrr::map_chr(based_on, \(x) x[1] %||% NA_character_)) %>%
  select(run, based_on)

res <- microbenchmark::microbenchmark(
  rlog = add_dofv(l),
  slog = add_dofv(sl),
  times = 20
)

summary(res, include_cld = FALSE)
```

</details>


<details>
<summary>times on tip of this PR (d3a08e5a)</summary>

```
ℹ Loading bbr
d282051145aa463b0b79e3c5f8b37dc2742c51a8
# A tibble: 4 × 2
  run   based_on
  <chr> <chr>
1 1     NA
2 2     1
3 3     1
4 4     2
  expr      min       lq     mean   median       uq      max neval
1 rlog 323.9841 334.5248 348.4071 344.6260 356.6335 400.1861    20
2 slog 324.8949 343.7585 350.9603 353.8762 357.3788 375.5862    20
```

</details>

<details>
<summary>times with proposed changes</summary>

```
ℹ Loading bbr
2aa591cb9ec7079ad3d6b74c57f4ebdce3d844c2
# A tibble: 4 × 2
  run   based_on
  <chr> <chr>
1 1     NA
2 2     1
3 3     1
4 4     2
  expr       min        lq      mean    median        uq       max neval
1 rlog 100.46002 106.24854 117.31686 109.47293 115.60588 245.19558    20
2 slog  11.98419  12.88671  13.95734  13.69641  14.48372  19.76291    20
```

</details>

I'll let you, @barrettk, and others weigh in on whether it adds too much complexity (or if you see ways to simplify it). I don't have any objections to sticking with the current serial approach.

seth127:
Hi @leovuong thanks for commenting here. Glad to have you engaged with `bbr`! I'm not sure what you mean by this:

> I can't use add_dofv() recently and I was wondering what should I use instead?

`add_dofv` has _not_ been released in `bbr` yet, this PR is its first appearance. We hope to have this wrapped up soon, and available in the next release (likely `bbr 1.14.0`, hopefully later this month).

In terms of what you can use instead, I would advise just upgrading to `bbr 1.14.0` once it's released. However, you could also code up a simplified version of this yourself, using the `based_on` and `ofv` columns in the `summary_log()` output. Let us know if you have any other questions.

kyleam:
> Could you just make a commit(s) on top of this series, with your proposed approach [...]

Okay, will do.

> [...] would you be up for tweaking the tests in the way you describe [here](https://github.com/metrumresearchgroup/bbr/pull/731#discussion_r2046866211)?

Sure.

seth127:
@kyleam ok great, thanks. As a side note, your commits will probably make some of your previous review comments irrelevant. Feel free to "resolve" those. @barrettk will plan to make another pass after your commits, both to review them, but also to address any remaining comments from your earlier review.

Thanks again.

leovuong:
@seth127 I'm pretty sure I was able to use add_dofv in my previous project. I took the code from the MeRGE workshop at PAGE 2024, and it worked. That's why I'm quite surprised when you said that it's not yet available... Anyways I'm looking forward to bbr 1.14.0, and thanks for replying to my comment by the way.

seth127:
@leovuong ah yes, now I understand your confusion. That PAGE workshop had an early prototype of this feature, sourced from the `script/functions/bbr-helpers.R` file. In case you've lost that code, here it is:

<details>
<summary>add_dofv prototype code</summary>

```r
#' Compare ofv to that of the model in based_on
#' If multiple models are in based_on, use the first
add_dofv <- function(.runlog) {
  .runlog <- .runlog %>%
    mutate(based_on_join = purrr::map_chr(based_on, function(x) {ifelse(length(x) > 0, x[1], NA)})) %>% 
    left_join(
      select(.runlog, based_on_join=run, based_on_ofv=ofv), 
      by="based_on_join"
    )
  
  suppressSpecificWarning({
    .runlog <- mutate(.runlog, dofv = as.numeric(format(ofv - based_on_ofv, nsmall = 2)))  
  }, "NAs introduced by coercion")
    
  select(.runlog, -based_on_ofv, -based_on_join)
}
```

</details>

As you can see from this PR, that code is much simpler than the function that will end up in `bbr 1.14.0`. This is a actually a good example of the difference between writing ad-hoc helper functions vs. actual package code! Notably, that package code needs to be prepared to handle all manner of errors and edge cases.

So, feel free to use this code as needed, but I would highly recommend deleting it and using the actual `bbr::add_dofv()` function as soon as it is released.

tests/testthat/test-summary-log.R
kyleam: I said:

>  some custom test setup that adjusts the OBJV line
> [...]
> it'd be good to assert that calls with a run log and summary log for the same set of models give the same dofv values.

Both these are done with 2cd0480b.

kyleam:
@seth127:

> As a side note, your commits will probably make some of your previous review comments irrelevant. Feel free to "resolve" those.

I've pushed the changes (and marked all related comments as resolved).  It's ready for both you and @barrettk to take a look and address the remaining issues.

To highlight what I view as the remaining blockers:

 * I think `calc_aic_bic` should either be added as a `summary_log` parameter and documented, or `summary_log_impl` should be reworked to avoid exposing `calc_aic_bic` via `summary_log`'s `...`  (which is documented as arguments to `model_summaries`).

   https://github.com/metrumresearchgroup/bbr/pull/731#discussion_r2046854454

 * My understanding is that the AIC/BIC calculation relies on NONMEM's objective function being `-2 loglikelihood`.  At the very least, I think that should be documented as a code comment.  But, as you know, the comment in this PR has prompted some wider internal discussion.  So that comment at this point is a stand-in to make whatever adjustments that discussion lands on.

   https://github.com/metrumresearchgroup/bbr/pull/731#discussion_r2046858657

R/summary-log.R
```diff
@@ -266,3 +281,120 @@ do_if_bbi_sum <- function(fn, mode) {
...
+#' summary_log(MODEL_DIR) %>% add_dofv()
+#'
+#' # Automatically adds `ofv` column in absence of `add_summary()` call:
+#' run_log(MODEL_DIR) %>% add_dofv()
+#' }
+#'
+#'
+#' @seealso [add_summary()]
+#' @export
+add_dofv <- function(.log_df, .mod_ref = NULL) {
```
kyleam, line 326: I'll push an update to restore the check that "absolute_model_path" is present as a column.

kyleam:
This PR has been replaced by gh-743.

PR #741, opened 2025-03-26
update ubuntu-20.04 jobs to ubuntu-22.04

barrettk:
update ubuntu-20.04 jobs to ubuntu-22.04
- The ubuntu-20.04 image is scheduled to be retired on 2025-04-01

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -32,7 +32,7 @@ jobs:
             bbi_version: v3.0.2
             mpn: 2023-01-25
             cran_override: 'https://mpn.metworx.com/snapshots/stable/2023-01-25'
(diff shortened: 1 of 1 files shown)


kyleam (changes requested):
The `rspm` URL should be updated from `focal` (20.04) to `jammy` (22.04).

barrettk:
@kyleam Awesome thanks. The date is ok as-is? i.e. `https://packagemanager.posit.co/cran/__linux__/jammy/2023-01-25`

kyleam:
> The date is ok as-is? i.e. `https://packagemanager.posit.co/cran/__linux__/jammy/2023-01-25`

Yes, the date should stay the same.  It matches the MPN snapshot that's the oldest snapshot we want to check compatibility with.

kyleam (approved):

Issue #740, opened 2025-02-11
Design Discussion: dOFV and based_on


barrettk:
This discussion is related to the latest effort of adding `dOFV`, `AIC`, and `BIC` columns to the `summary_log` (and `add_summary()`), which is outlined in more detail in https://github.com/metrumresearchgroup/bbr/issues/677. 

For the purposes of calculating `dOFV` however, there has been some discussion on how we should treat `based_on` references. Specifically, what do we do if a model is based on several models? 
 - [ ] Can we take the first one? If so it is important to label which one we used?
    -  Note that we use similar logic for `model_tree()` (which cant support multiple "based on" models due to dependency limitations).
    - Whatever we decide, the logic may inform future decisions where comparing models in model ancestry (`bbr::get_model_ancestry()`), so we should make a decision that holds for other comparisons as well. Diverging from `model_tree` behavior is ok, but the pattern should be consistent whenever math is performed (e.g., comparing parameter estimates).
 - [ ] Should we add rows and calculate `dOFV` for each `based_on` model (perhaps collapsed/nested like `tags`)?
    -  Doing this might be preferable, though **note** that the `summary_log` **_does not report_** `based_on`. _You only get_ `based_on` when you call `run_log(MODEL_DIR) %>% add_summary()`. In other words, it might be difficult to interpret a nested list of `dOFV` _without_ a corresponding `based_on` column indicating the order of the values.

In draft PR https://github.com/metrumresearchgroup/bbr/pull/731, we currently do the first option. 

If there are no models found with additional `based_on` flags, we get the following:
<details closed>
<summary>Default Behavior</summary>

```r
> run_log(MODEL_DIR) %>% add_summary() %>% dplyr::select(any_of(cols_keep))
# A tibble: 3 × 6                                                                                                                                                             
  run    based_on    ofv  dofv   aic   bic
  <chr>  <list>    <dbl> <dbl> <dbl> <dbl>
1 1-boot <chr [1]>   NA     NA   NA    NA 
2 1      <NULL>    2583.    NA 2597. 2630.
3 2      <chr [1]> 2583.     0 2597. 2630.

> summary_log(MODEL_DIR) %>% dplyr::select(any_of(cols_keep))
# A tibble: 3 × 5                                                                                                                                                             
  run      ofv  dofv   aic   bic
  <chr>  <dbl> <dbl> <dbl> <dbl>
1 1-boot   NA     NA   NA    NA 
2 1      2583.    NA 2597. 2630.
[...]

Issue #401, opened 2021-09-24
compare_ofv()


seth127:
Easily compare objective function values between models

# Proposed code

```r
#' Compare objective function values between models
#'
#' Takes a bbi_summary_log_df and character vector of models 
#' and filters to only those models. 
#' Returns tibble with run, ofv, tags,
#' and any columns passed to ...
compare_ofv <- function(models, log_df, ...) {
  log_df %>%
    filter(run %in% models) %>%
    select(run, ofv, tags, ...) %>%
    collapse_to_string(tags, ...) %>%
    knitr::kable()
}
```

oskarrgh:
> good idea. We had talked about adding that to `summary_log()`, so this could just grab it

This would be very helpful during model dev already in model_summary() but naturally key metric in in summary_log() output.

PR #743, opened 2025-05-01
Extend summary log with AIC and BIC columns and introduce add_dofv

### LINKED ISSUE #677: dOFV and AIC columns for summary_log()

kyleam:
This PR replaces gh-731.  Interface-wise, it makes the same changes that PR 731 landed on: 1) adds AIC and BIC columns to the summary log and 2) exports a new function (`add_dofv`) for extending a run/summary log with a column for dOFV.

Main user-facing changes from gh-731:

 * `ofv_with_constant` instead of `ofv_no_constant` is used for the `-2 log-likelihood` component of the AIC and BIC calculations

   (related PR 731 comments: <https://github.com/metrumresearchgroup/bbr/pull/731#discussion_r2046858657>, <https://github.com/metrumresearchgroup/bbr/pull/731#issuecomment-2814114798>)

 * reporting AIC, BIC, and dOFV is disabled (set to `NA`) in various scenarios (e.g., dOFV is reported as `NA` if the two models have different methods)

 * a `summary_log`/`add_summary` caller can no longer specify `calc_aic_bic` as part of `...` (which are documented as getting sent to `model_summaries`).

   (related PR 731 comments: <https://github.com/metrumresearchgroup/bbr/pull/731#discussion_r2046854454>, <https://github.com/metrumresearchgroup/bbr/pull/731#issuecomment-2814114798>)

Thanks to @callistosp, @kylebaron, and @timwaterhouse for discussion and input on the first two items.

cc: @barrettk @seth127

Re: #401
Re: #740
Closes #677


seth127:
Thanks for putting this together @kyleam . I talked through a lot of this today with @barrettk and he is going to review this and pull in me and/or @kylebaron @timwaterhouse if needed.

tests/testthat/test-dofv.R
```diff
@@ -0,0 +1,232 @@
...
+        ofv_no_constant = ofvs[9]
+      ),
+      list(
+        # dofv is set to NA due to unsupported method, despite base's matching
+        # method.
+        subpath = "10",
+        based_on = "09",
+        data_set = "../data.csv",
+        final_est_method = "Importance Sampling of Variance-Covariance (SIR)",
+        ofv_no_constant = ofvs[10]
```
barrettk, line 92: Comprehensive test suite - checks all the cases I would think to
 - tests confirm we support `run_log` and `run_log() %>% add_summary()`. Given the handling *I don't* think we necessarily need to test `summary_log() %>% add_dofv()`, though just noting that we don't currently (I tested manually, though given that we use `summary_log_impl` under the hood it makes sense that we don't really need to test that call too)

R/summary-log.R
```diff
@@ -212,15 +271,16 @@ extract_heuristics <- function(.s) {
 #' @describeIn extract_from_summary Extract objective function value (without constant) for the final estimation method
 #' @importFrom purrr map map_dbl
 #' @keywords internal
-extract_ofv <- function(.s) {
+extract_ofv <- function(.s, with_constant = FALSE) {
+  name <- if (with_constant) "ofv_with_constant" else "ofv_no_constant"
```
barrettk, line 275: missing `@param` documentation for `with_constant`. May want to update the title too since it says "(without constant)"

barrettk (commented):
First round of comments: only one suggested change (docs)

R/summary-log.R
```diff
@@ -415,3 +416,197 @@ add_aic_bic <- function(log_df) {
...
+
+  if (is.null(mod_ref)) {
+    mods <- purrr::map(mpaths_log, read_model)
+    mpaths_ref <- purrr::map_chr(mods, function(m) {
+      get_based_on(m, .check_exists = FALSE)[1] %||% NA_character_
+    })
+  } else {
+    mods <- NULL
+    mpaths_ref <- mod_ref[[ABS_MOD_PATH]]
+  }
```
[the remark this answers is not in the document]
kyleam: > but just wondering if we wanted to check the model path (in the linked section) too.

Sorry, I'm not following.  

Are you proposing that when a model is passed as a reference, it should return NA when compared with itself?

R/summary-log.R
kyleam: Thanks for spotting, updated.

R/summary-log.R
```diff
@@ -266,3 +329,287 @@ do_if_bbi_sum <- function(fn, mode) {
...
+#'     is not meaningful for model comparison (e.g., SAEM or Bayesian methods).
+#'
+#'   * the input data sets of the two models differs, as indicated by a
+#'     different file path or different number of observations recorded in the
+#'     `.lst` files.
+#'
+#'     The above criteria serve to flag the common cases where the models use
+#'     different input data. Note, however, that models may have used different
+#'     data despite having a matching data path and number of observations.
+#'
```
barrettk, line 456: Could we add some examples?

The first few examples below may be overkill, though I could see value in demonstrating the flexibility. 
 - Not that it's our responsibility, but I also think it can be helpful to remind users of other helper functions via examples and `@seealso` references (especially for a package like `bbr`). Your call but just my 2 cents.
```r
#' @examples
#' \dontrun{
#' 
#' run_log(MODEL_DIR) %>% add_dofv()
#' run_log(MODEL_DIR) %>% add_summary() %>% add_dofv()
#' summary_log(MODEL_DIR) %>% add_dofv()
#' 
#' # Provide a reference model
#' run_log(MODEL_DIR, .mod_ref = MOD1) %>% add_dofv()
#' }
```

barrettk (commented):
One more comment. Everything else looks good.

I'd like to have some confirmation on [this comment](https://github.com/metrumresearchgroup/bbr/pull/743#discussion_r2076038676) from a user perspective, but afterwards I will approve.

R/summary-log.R
kyleam: > remind users of other helper functions via examples and `@seealso` reference

Is there a seealso you'd like to see beyond that add_summary one that's already there?

As for the examples, they're usually good to have *when they run*.  If they don't, I think they have to be demonstrating something particularly useful or not apparent from the documented parameters to be worth adding.  (Part of the problem with examples that don't run is that they are likely to grow stale, or even be wrong in the first place like your last call above that passes `.mod_ref` to `run_log`).

barrettk: >Is there a seealso you'd like to see beyond that add_summary one that's already there?

No - I think `add_summary()` makes sense, though im not sure `run_log()` or `summary_log()` do. I might add `add_dofv` _to_ `summary_log()`, but not too concerned about that (I think `add_summary()` is the more commonly used function from scripts ive seen).

I see your point about stale code, typos/incorrect syntax, etc. I do think we already have a number of cases where we include examples anyways (e.g., `setup_bootstrap_run`...though there I think they help illustrate the process in a way that _isn't_ needed for `add_dofv`), and **your parameter description is already pretty clear**. 

I often immediately scroll down to the examples to see how functions are used, and only read the details to fill in missing gaps, though I may be injecting my bias too strongly here. Am ok with not including them.

kyleam: Yeah, it's of course very subjective.  I'll leave things as they are for now.  Thanks for your input.

R/summary-log.R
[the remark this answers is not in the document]
seth127: I don't have a strong opinion on this. Either 0 or NA seem sensible to me.

barrettk (approved):
Looks good; thanks for your replies

Issue #736, opened 2024-04-05
bbr::use_bbi() fails


bmreilly:
Recently I tried to use `bbr::use_bbi()` and I am getting a JSON parsing error. I'm wondering if the location of files to download has changed and needs to be updated?
![image](https://github.com/metrumresearchgroup/bbr/assets/7952375/c4335f89-b625-4a8d-adf2-80330791a5ea)

It looks like it is coming from `bbr:::current_release_url(owner = 'metrumresearchgroup', repo = 'bbi')`
![image](https://github.com/metrumresearchgroup/bbr/assets/7952375/3fe60de6-3cc3-4987-948f-642097c90d42)

seth127:
@kyleam I think the approach you mentioned in the previous comment makes sense. I support you going ahead and making a PR for that whenever you get the time.

PR #748, opened 2025-06-05
use_bbi: avoid errors due to rate limiting 

kyleam:
This PR

 * improves the error handling to avoid the confusing JSON parsing error if the `download.file` fails for any reason

 * switches from using api.github.com to using https://metrumresearchgroup.github.io/bbi/latest.json (metrumresearchgroup/bbi#343) to get the latest bbi version in order to avoid errors due to rate-limiting like the one reported in gh-736

I've manually exhausted the requests limit and confirmed that 1) the old `use_bbi` variant fails and 2) this PR's variant succeeds.

cc: @bmreilly  @andersone1 @seth127

---

- [ ] wait for approval of metrumresearchgroup/bbi#343


andersone1:
@kyleam 
Code/functionality looks good. Only comments are for `current_release_url`:
 - Can the arguments `owner` and `repo` go away given they are not used in the function anymore?
 - Can all of these go away since their imports are not required (jsonlite is used but it is explicitly name spaced):
 ```
#' @importFrom jsonlite fromJSON
#' @importFrom glue glue
#' @importFrom utils download.file
```

kyleam:
> Can the arguments `owner` and `repo` go away given they are not used in the function anymore?

Very good point, thanks for noticing.

>  Can all of these go away since their imports are not required (jsonlite is used but it is explicitly name spaced):

I'll clean those up.

kyleam:
@andersone updated series pushed (e68ab862)

 * removes unused parameters from `current_release_url`

 * renames `current_release_url` to make it clear it's bbi-specific and updates docstring

 * cleans up imports

andersone1 (approved):

PR #750, opened 2025-06-20
add_dofv: don't error on non-existent models

kyleam:
This PR updates `add_dofv` it warns rather than errors when a model referenced by a `based_on` value (or, less likely, by a row of the run log) doesn't exist.


barrettk (approved):
Changes look good

Issue #758, opened 2026-02-12
update_model_id as argument to copy_model_from


callistosp:
Currently you need to pipe the output from `copy_model_from()` into `update_model_id()` to update the run numbers in the new control stream. I would argue that the default workflow should be that `copy_model_from()` does this automatically by default, and can be turned off if desired. 

Proposed syntax:
`copy_model_from(..., update_model_id=TRUE)`

PR #761, opened 2026-03-04
copy_model_from: update model ID by default

kyleam:
In gh-758 @callistosp suggested that by default `copy_model_from` should call `update_model_id` underneath, allowing the behavior to be disabled via an argument.

This series makes that update, with the primary change in 1757de9f (copy_model_from.bbi_nonmem_model: add .update_id argument).

----

bbr.bayes needs to be adjusted for compatibility.  I'll leave this PR as a draft until the bbr.bayes PR is up too.

### DIFF
--- a/R/copy-model-helpers.R
+++ b/R/copy-model-helpers.R
@@ -90,7 +90,7 @@ update_model_id <- function(
   )
 
   txt <- gsub(
-    paste0("\\Q",based_on_id,"\\E", .suffixes),
+    paste0("\\b\\Q",based_on_id,"\\E", .suffixes),
     paste0(mod_id, "\\1"),
     txt,
     ignore.case = TRUE
--- a/R/copy-model-helpers.R
+++ b/R/copy-model-helpers.R
@@ -70,7 +70,7 @@ update_model_id <- function(
   if (is.null(based_on)) {
     stop(glue("Cannot call update_model_id() because .mod$based_on is empty for model {mod_id}"))
   }else{
-    based_on_id <- get_model_id(based_on)
+    based_on_id <- get_model_id(based_on)[1]
   }
   message(glue("replacing {based_on_id} with {mod_id} in {modelfile}"))
 
--- a/R/copy-model-helpers.R
+++ b/R/copy-model-helpers.R
@@ -67,11 +67,11 @@ update_model_id <- function(
   modelfile <- get_model_path(.mod)
   based_on <- get_based_on(.mod)
 
-  if (is.null(based_on)) {
+  if (!length(based_on)) {
     stop(glue("Cannot call update_model_id() because .mod$based_on is empty for model {mod_id}"))
-  }else{
-    based_on_id <- get_model_id(based_on)[1]
   }
+  based_on_id <- get_model_id(based_on)[1]
+
   message(glue("replacing {based_on_id} with {mod_id} in {modelfile}"))
 
   ## construct suffixes regex string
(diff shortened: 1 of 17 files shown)


R/copy-model-helpers.R
```diff
@@ -67,11 +72,11 @@ update_model_id <- function(
...
 
-  if (is.null(based_on)) {
+  if (!length(based_on)) {
     stop(glue("Cannot call update_model_id() because .mod$based_on is empty for model {mod_id}"))
-  }else{
-    based_on_id <- get_model_id(based_on)
   }
+  based_on_id <- get_model_id(based_on)[1]
+
   message(glue("replacing {based_on_id} with {mod_id} in {modelfile}"))
```
kyleam, line 80: `copy_model_from` will now trigger this message.  That's probably a good thing in terms of users being aware of the change in behavior, but I wonder if some users will be annoyed by `copy_model_from` now being noisy.  Any thoughts on whether we should adjust `update_model_id` to use `verbose_msg` (i.e. honor the `bbr.verbose` option)?

kyleam:
> I'll leave this PR as a draft until the bbr.bayes PR is up too.

That's here: https://github.com/metrumresearchgroup/bbr.bayes/pull/172

R/copy-model-helpers.R
barrettk: I imagine you'll get a lot of opinions on this. My vote would be to leave it as-is for now so, to your point, users will notice the change. We could then change to a verbose_msg in a later release if users report that as being annoying, but at least the current message will spread more awareness _of the change_ for the time being. Not sure my opinion is as relevant as actual users, but just my two cents.

vignettes/model-tree.Rmd
```diff
@@ -71,22 +71,21 @@ create_tree_models <- function(MODEL_DIR){
...
-    update_model_id()
-  mod4 <- copy_model_from(mod2, "4", .overwrite = TRUE) %>% update_model_id()
-  mod5 <- copy_model_from(mod4, "5", .overwrite = TRUE) %>% update_model_id()
-  
+  mod2 <- copy_model_from(mod1, "2", .overwrite = TRUE, .add_tags = tags, .update_id = TRUE) %>%
+    add_star()
+  mod3 <- copy_model_from(mod2, "3", .overwrite = TRUE, .add_tags = "DV: nmol", .update_id = TRUE)
+  mod4 <- copy_model_from(mod2, "4", .overwrite = TRUE, .update_id = TRUE)
+  mod5 <- copy_model_from(mod4, "5", .overwrite = TRUE, .update_id = TRUE)
+
```
barrettk, line 79: Did you add `.update_id = TRUE` just for clarity (here and other cases)? It wouldnt be needed since it's the default right? Ok regardless, just want to clarify

kyleam: > Did you add `.update_id = TRUE` just for clarity (here and other cases)?

Correct.  (Mentioned in fe2ffd527 for changes to test calls, but I didn't repeat that rationale in c3d9267a).

R/copy-model-helpers.R
```diff
@@ -90,7 +90,7 @@ update_model_id <- function(
   )
 
   txt <- gsub(
-    paste0("\\Q",based_on_id,"\\E", .suffixes),
+    paste0("\\b\\Q",based_on_id,"\\E", .suffixes),
```
barrettk, line 93: Nice catch and accompanying test

barrettk (approved):
Changes make sense and tests passed

kyleam:
@barrettk Thanks for the review.  I'll leave this open until metrumresearchgroup/bbr.bayes#171 is merged so that bbr.bayes CI doesn't fail on merge.

So the merge order will be:
 * merge metrumresearchgroup/bbr.bayes#171
 * merge this PR
 * drop tip commit from metrumresearchgroup/bbr.bayes#172 and merge

PR #765, opened 2026-05-12
options: don't set mrg.id_col and mrg.num_col

kyleam:
bbr sets the default values of many options via .onLoad.  The recent commits that wired up the mrg.num_col and mrg.id_col options followed the same pattern (632e1248, 25ac6c1c).  However, these two options are shared across several packages, so it's confusing for bbr to set them.

Instead inline the fallback value for each getOption call.


barrettk (approved):
Looks good. Rationale makes sense to me

PR #767, opened 2026-05-22
submit model: `.mode` guard updates

kyleam:
* update test of `.mode="sge"` abort on Slurm workflows for Metworx change

 * when `.model="slurm"` is specified but `sbatch` isn't found, abort upfront rather than leaving it to bbi

 * if either `.mode="sge"` or `.mode="slurm"` is specified but the corresponding executable isn't available, suggest `.mode="local"` in error message


barrettk (approved):
Looks good

